PR #74, opened 2023-07-07
chore(pydantic) bump to pydantic 2

twolfvb:
In our project we can't use pydantic 2 (reportedly [5-50 x faster than pydantic 1](https://news.ycombinator.com/item?id=35490449), as it was rewritten using rust), given that we use Langchain and it uses pydantic 1.

on Langchain, removing then adding pydantic using poetry:
```
Because langchainplus-sdk (0.0.20) depends on pydantic (>=1,<2)
 and no versions of langchainplus-sdk match >0.0.20,<0.0.21, langchainplus-sdk (>=0.0.20,<0.0.21) requires pydantic (>=1,<2).
So, because langchain depends on both langchainplus-sdk (^0.0.20) and pydantic (^2.0.2), version solving failed.
```

changes code using [bump-pydantic](https://docs.pydantic.dev/latest/migration/#code-transformation-tool), as recommended by the documentation

manually created default factories in the client as env= is no longer used

Please let me know if this seems ok.


hinthornw:
Thank you for the PR! We are looking into the migration. Since `langchain` currently depends on this, we will likely first have to support both 1 and 2 to remain backwards compatible with peoples workloads

mjspeck:
Langchain also depends on `openapi-schema-pydantic` which depends on a pydantic 1.9, which has a vulnerability. However, that project is no longer maintained. I've put some suggestions for how to address this in [this issue](https://github.com/langchain-ai/langchain/issues/4974#issuecomment-1654769187).

twolfvb:
I will be closing this for now to keep the repo clean.

hinthornw:
We're planning to rm the `openapi-schema-pydantic` dep in langchain and do some more cleanup. Thanks everyone for your patience.

PR #85, opened 2023-07-12
Use a single session per client instance for all http requests

nfcampos:


python/langsmith/client.py
```diff
@@ -144,18 +166,49 @@ def _headers(self) -> Dict[str, str]:
...
+        url: str,
+        request_kwargs: Mapping,
+    ) -> Response:
+        try:
+            response = self.session.request(
+                request_method, url, stream=False, **request_kwargs
+            )
+            raise_for_status_with_text(response)
+            return response
+        except HTTPError as e:
```
nfcampos, line 195: @hinthornw I left this abundance of error handling here, but I suspect now that we're using urllib3 for retries a lot of this is pointless?

hinthornw: Yeah probably

python/langsmith/client.py
```diff
@@ -122,6 +128,36 @@ def validate_api_key_if_hosted(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
+    def __init__(
+        self,
+        api_key: Optional[str] = None,
+        api_url: Optional[str] = None,
+        retry_config: Optional[Retry] = None,
+        timeout_ms: Optional[int] = None,
+        **kwargs: Any,
+    ) -> None:
+        kwargs = kwargs or {}
+        if api_key is not None:
```
nfcampos, line 140: pydantic is sad

hinthornw: :( It also messes up the docstrings / type hinting when we add an `__init__` method, unfortunately. Should I just stop using pydantic for the client? I think that would be good..

hinthornw (approved):

PR #94, opened 2023-07-15
Update default url

hinthornw:
Currently presumes localhost if not specified.
Suggested change would be to assume prod `https://api.smith.langchain.com` if an api key is specified, otherwise fall back to localhost.

Would eliminate 1 strictly required env var


nfcampos (approved):

agola11:
Won't this actually make it hard to test dev?

agola11:
nvm I see you can still explicitly set it. I would just sanity check that!

PR #96, opened 2023-07-17
Add pagination

hinthornw:


js/src/run_trees.ts
```diff
@@ -62,7 +62,7 @@ export class RunTree implements BaseRun {
       api_key: getEnvironmentVariable("LANGCHAIN_API_KEY"),
       caller_options: {},
       start_time: Date.now(),
-      end_time: Date.now(),
+      end_time: null,
```
dqbd, line 65: _nit_: shouldn't that be `undefined` instead? Or maybe we should change `end_time` data type?

dqbd (commented):
Smaller nitpicks :)

dqbd (approved):
Smaller nitpicks :)

js/src/tests/client.int.test.ts
```diff
@@ -264,8 +279,11 @@ test("test create dataset", async () => {
     apiUrl: "http://localhost:1984",
   });
   const datasetName = "__test_create_dataset";
-  const datasets = await langchainClient.listDatasets({});
-  if (datasets.map((dataset) => dataset.name).includes(datasetName)) {
+  const datasets = [];
+  for await (const dataset of langchainClient.listDatasets({})) {
+    datasets.push(dataset);
+  }
+  if (datasets.some((dataset: any) => dataset.name === datasetName)) {
```
dqbd, line 286: _nit_: Using `toArray` we would be able to get rid of `any`

js/src/client.ts
```diff
@@ -108,6 +108,14 @@ const raiseForStatus = async (response: Response, operation: string) => {
   }
 };
 
+export async function toArray<T>(iterable: AsyncIterable<T>): Promise<T[]> {
```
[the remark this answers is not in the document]
hinthornw: Hm probably not to start ya

js/src/run_trees.ts
hinthornw: Yes!

PR #111, opened 2023-07-21
remove nginx config mount in sdk [LS-446]

efriis:
Resolve docker mount errors like
```
Error response from daemon: Mounts denied: 
The path /opt/homebrew/lib/python3.11/site-packages/langsmith/cli/conf/nginx.conf is not shared from the host and is not known to Docker.
You can configure shared paths from Docker -> Preferences... -> Resources -> File Sharing.
See https://docs.docker.com/desktop/mac for more info.
```


python/langsmith/cli/docker-compose.yaml
```diff
@@ -11,8 +11,6 @@ services:
     depends_on:
       - langchain-backend
       - langchain-playground
-    volumes:
-      - ./conf/nginx.conf:/etc/nginx/default.conf:ro
```
efriis, line 15: @hinthornw any reason we needed this? Works without it for me

hinthornw: I don't think so

hinthornw (approved):

PR #121, opened 2023-07-30
Add two more run types for Runnable

nfcampos:
I thought we had decided to remove all validation from SDK and rely on server instead?


hinthornw:
Ya we dual type as string so it doesn't raise an error but I'll remove the enum entirely.

hinthornw (approved):

PR #127, opened 2023-08-02
Rm Pydantic Dep

hinthornw:
So dependents don't have to choose between v1 and v2 and to reduce the number of reqs.

Instead just use regular classes. In this case, I've inherited from `dict` to make them easily serializable, though we could remove that

Tested with langchain integration tests on pr https://github.com/langchain-ai/langchain/pull/8650
Tested on the integration tests that do exist.

Bumping the minor version to avoid breaking existing langchain builds, since the pydantic syntax in the tracer currently would make certain keys not work. It _may_ be possible with dataclasses there but dataclasses make inheritance frustrating also (e.g., can't have required keys in child classes)

Still want to add a lot more tests from the docs examples since that's not yet hooked up explicitly.


python/langsmith/utils.py
```diff
@@ -23,6 +28,63 @@ class LangSmithConnectionError(Exception):
     """Couldn't connect to the LangSmith API."""
 
 
+class DictMixin(dict):
```
nfcampos, line 31: If we're making such an involved mixing might as well use dataclass ?

hinthornw: Dataclasses have annoying inheritance patterns but I could try not using `RunBase` or anything like that for these

PR #142, opened 2023-08-11
Code to be both pydantic v1 and v2 compatible

eyurtsev:
Code is both v1 and v2 compatible


hinthornw:
Gonna land and cut a release

hinthornw (approved):
Tested against langchain and seems like it works still

PR #157, opened 2023-08-19
Add openapi

hinthornw:


hinthornw:
Won't do this . Will do in backend instead just scoping things out

PR #161, opened 2023-08-22
Add hub to sdk

efriis:
Adds the hub backend to the dockerfile

Also changes the `--dev` flag to point to `dev-` images instead of `rc-` images

And adding a `--beta` flag to point to `rc-` images moving forward

Currently, hub is only built with the dev images.


hinthornw:
I'll take over

hinthornw (approved):

PR #220, opened 2023-09-13
Support urllib3<1.26

joshuasundance-swca:
The `allowed_methods` keyword argument was added to `urllib3.util.Retry` in version 1.26.

This PR alters `langsmith.client._default_retry_config` to:

1. Check the version of `urllib3`
2. If version is >= 1.26, use the `allowed_methods` keyword of the `Retry` object
3. Otherwise, do not use the `allowed_methods` keyword.

### DIFF
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -8,7 +8,6 @@ import io
 import json
 import logging
 import os
-import pkg_resources
 import socket
 import uuid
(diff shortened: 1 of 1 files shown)


hinthornw:
Looks good - thanks for the contribution!

hinthornw (approved):

PR #229, opened 2023-09-18
Sort by dotted order in py sdk

nfcampos:
Same needs to be done in JS SDK

### DIFF
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -719,7 +719,7 @@ class Client:
             list
         )
         runs: Dict[uuid.UUID, ls_schemas.Run] = {}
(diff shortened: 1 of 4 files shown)


python/langsmith/client.py
```diff
@@ -719,7 +719,7 @@ def _load_child_runs(self, run: ls_schemas.Run) -> ls_schemas.Run:
             list
         )
         runs: Dict[uuid.UUID, ls_schemas.Run] = {}
-        for child_run in sorted(child_runs, key=lambda r: r.execution_order):
+        for child_run in sorted(child_runs, key=lambda r: r.dotted_order):
```
hinthornw, line 722: This fn is also something we'd change once trace ID changes are made

PR #237, opened 2023-09-22
RFC: httpx

hinthornw:
Would love feedback on this. Especially on the following:

1. Finalizing for the async client (Right now the event loop often fails)
2.  Thoughts on how to reduce code copying.


python/langsmith/client.py
```diff
@@ -332,93 +347,47 @@ def request_with_retries(
...
-            Additional request parameters.
-
-        Returns
-        -------
-        Response
-            The response object.
+        filter_null_params: bool = True,
+    ) -> httpx.Response:
+        """Send a request with retries."""
+        with _request_error_handler(request_method, url) as handler_state:
```
hinthornw, line 353: Thoughts on this error handling context manager to reduce duplicate handling code?

python/langsmith/client.py
```diff
@@ -645,16 +592,128 @@ def create_run(
...
+                All root run traces have execution_order 1.
+        **kwargs : Any
+            Additional keyword arguments.
+
+        Raises
+        ------
+        LangSmithUserError
+            If the API key is not provided when using the hosted service.
+        """
+        request_kwargs = self._prepare_run_create(
```
hinthornw, line 676: Lots of instances like this where prep code is shared and then the request method differs. Could maybe reduce LoC by doing a decorator or something but it's a bit ugly

python/langsmith/client.py
```diff
@@ -332,93 +347,47 @@ def request_with_retries(
...
+        """Send a request with retries."""
+        with _request_error_handler(request_method, url) as handler_state:
+            if filter_null_params and "params" in request_kwargs:
+                params = request_kwargs["params"]
+                if params is not None:
+                    request_kwargs["params"] = {
+                        k: v for k, v in params.items() if v is not None
+                    }
+            response = self._client.request(request_method, url, **request_kwargs)
+            handler_state["response"] = response
```
hinthornw, line 361: A bit weird way to add checking of response in the error handling...

python/langsmith/client.py
```diff
@@ -145,16 +116,62 @@ def _serialize_json(obj: Any) -> str:
...
+    if hasattr(asyncio, "Runner"):
+        # Python 3.11+
+        # Run the coroutines in a new event loop, taking care to
+        # - install signal handlers
+        # - run pending tasks scheduled by `coros`
+        # - close asyncgens and executors
+        # - close the loop
+        with asyncio.Runner() as runner:
+            # Run the coroutine, get the result
+            runner.run(coro)
```
hinthornw, line 166: This line still giving a `RuntimeError: Event loop is closed`

it was locking previously with simple stuff.

python/langsmith/client.py
```diff
@@ -145,16 +116,64 @@ def _serialize_json(obj: Any) -> str:
...
+        if loop_running:
+            # If we try to submit this coroutine to the running loop
+            # we end up in a deadlock, as we'd have gotten here from a
+            # running coroutine, which we cannot interrupt to run this one.
+            # The solution is to create a new loop in a new thread.
+            with concurrent.futures.ThreadPoolExecutor(1) as executor:
+                executor.submit(_run_coro, coro).result()
+        else:
+            _run_coro(coro)
+    except RuntimeError as e:
```
hinthornw, line 154: I haven't yet figured out how to avoid this in all circumstances

python/langsmith/client.py
```diff
@@ -145,16 +123,44 @@ def _serialize_json(obj: Any) -> str:
...
+    client.close()
+
+
+def close_async_client(client: httpx.AsyncClient) -> None:
+    logger.debug("Closing Client._aclient")
+    if client.is_closed:
+        return
+    try:
+        with concurrent.futures.ThreadPoolExecutor(1) as executor:
+            executor.submit(lambda: asyncio.run(client.aclose())).result(timeout=10)
```
hinthornw, line 156: This may sometime lock if called explicitly in our tests if garbage collection isn't triggered explicitly beforehand (not always but sometimes). I'm not sure why at the moment.

PR #278, opened 2023-11-01
Infra/fix sdk compose

langchain-infra:
Also use the frontend-dynamic/single-origin url so that our application can spin up by only exposing 1 port.


hinthornw (approved):

js/src/cli/docker-compose.yaml
```diff
@@ -2,20 +2,26 @@ version: "3"
 services:
   langchain-playground:
     image: langchain/${_LANGSMITH_IMAGE_PREFIX-}langchainplus-playground:latest
+    ports:
+      - 3001:3001
   langchain-frontend:
-    image: langchain/${_LANGSMITH_IMAGE_PREFIX-}langchainplus-frontend:latest
+    image: langchain/${_LANGSMITH_IMAGE_PREFIX-}langchainplus-frontend-dynamic:latest
```
agola11, line 8: is frontend-dynamic the preferred frontend image now? People will likely not want to change the backend url when running with compose

agola11 (approved):

PR #282, opened 2023-11-03
Validate uuids

hinthornw:
In path


python/langsmith/schemas.py
```diff
@@ -399,3 +399,6 @@ class DatasetShareSchema(TypedDict, total=False):
     dataset_id: UUID
     share_token: UUID
     url: str
+
+
+Example.update_forward_refs()
```
efriis, line 404: was this intentional?

hinthornw: Ya I can separate to diff pr

PR #289, opened 2023-11-08
Add method to consistently get project

jakerachleff:
We plan to introduce an environment variable in hosted langserve that will overwrite the project used for traces from hosted deployments. We don't centralize how we refer to projects in any of our repos, so in order to ensure that our tracing works properly and our project selection is consistent, we need to introduce a shared method for getting the project that we can use throughout our different repos

### DIFF
--- a/python/langsmith/utils.py
+++ b/python/langsmith/utils.py
@@ -4,7 +4,7 @@ import functools
 import logging
 import os
 import subprocess
-from typing import Any, Callable, Dict, List, Mapping, Tuple, Union
+from typing import Any, Callable, Dict, List, Mapping, Optional, Tuple, Union
 
 import requests
 
@@ -249,6 +249,25 @@ def is_base_message_like(obj: object) -> bool:
             isinstance(getattr(obj, "additional_kwargs", None), dict),
             hasattr(obj, "type") and isinstance(getattr(obj, "type"), str),
         ]
+    )
+
+def get_tracer_project(return_default_value=True) -> Optional[str]:
+    """Get the project name for a LangSmith tracer."""
+    return os.environ.get(
+        # Hosted LangServe projects get precedence over all other defaults.
+        # This is to make sure that we always use the associated project
+        # for a hosted langserve deployment even if the customer sets some
+        # other project name in their environment.
+        "HOSTED_LANGSERVE_PROJECT_NAME",
+        os.environ.get(
+            "LANGCHAIN_PROJECT",
+            os.environ.get(
+                # This is the legacy name for a LANGCHAIN_PROJECT, so we it
+                # has lower precedence than LANGCHAIN_PROJECT
+                "LANGCHAIN_SESSION",
+                "default" if return_default_value else None,
+            ),
+        ),
     )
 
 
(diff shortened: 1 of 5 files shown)


python/langsmith/utils.py
```diff
@@ -251,6 +251,25 @@ def is_base_message_like(obj: object) -> bool:
         ]
     )
 
+def get_tracer_project(return_default_value=True) -> Optional[str]:
```
jakerachleff, line 254: Do we want to return `None` if `tracing_is_enabled() is False`?

python/pyproject.toml
```diff
@@ -28,6 +28,7 @@ langsmith = "langsmith.cli.main:main"
 python = ">=3.8.1,<4.0"
 pydantic = ">=1,<3"
 requests = "^2"
+pytest-subtests = "^0.11.0"
```
jakerachleff, line 31: This lets us run subtests

python/langsmith/utils.py
```diff
@@ -252,6 +252,26 @@ def is_base_message_like(obj: object) -> bool:
...
+    return os.environ.get(
+        # Hosted LangServe projects get precedence over all other defaults.
+        # This is to make sure that we always use the associated project
+        # for a hosted langserve deployment even if the customer sets some
+        # other project name in their environment.
+        "HOSTED_LANGSERVE_PROJECT_NAME",
+        os.environ.get(
+            "LANGCHAIN_PROJECT",
+            os.environ.get(
+                # This is the legacy name for a LANGCHAIN_PROJECT, so we it
```
jakerachleff, line 266:
```suggestion
                # This is the legacy name for a LANGCHAIN_PROJECT, so it
```

python/langsmith/utils.py
hinthornw: I don't think so, reason being if we are planning to use this within LangChain and a user manually creates a tracer or uses the context manager to start tracing, that currently doesn't commit to any environment variables.

python/langsmith/utils.py
```diff
@@ -252,6 +252,26 @@ def is_base_message_like(obj: object) -> bool:
...
+        # This is to make sure that we always use the associated project
+        # for a hosted langserve deployment even if the customer sets some
+        # other project name in their environment.
+        "HOSTED_LANGSERVE_PROJECT_NAME",
+        os.environ.get(
+            "LANGCHAIN_PROJECT",
+            os.environ.get(
+                # This is the legacy name for a LANGCHAIN_PROJECT, so it
+                # has lower precedence than LANGCHAIN_PROJECT
+                "LANGCHAIN_SESSION",
```
hinthornw, line 268: We should probably start actually warning users for this, it's a really old (1 yr) variables so i don't think it's used in practies

hinthornw (approved):
Seems reasonable

python/langsmith/utils.py
jakerachleff: Do you want that as part of this PR, or something different?

python/pyproject.toml
ericbuehl: could this be separated to a something like [tool.poetry.group.test.dependencies] so it doesn't install for prod builds?

jakerachleff: Yeah that's my bad, let me put up a fix

jakerachleff: https://github.com/langchain-ai/langsmith-sdk/pull/293

PR #296, opened 2023-11-16
Add inline_s3_urls query param to make multimodal evals work

samnoyes:


python/langsmith/client.py
```diff
@@ -1741,6 +1741,7 @@ def list_examples(
             pass
         if example_ids is not None:
             params["id"] = example_ids
+        params["inline_s3_urls"] = True
```
hinthornw, line 1744: Could we pass in as a keyword arg and have True as the default?

hinthornw (approved):

PR #301, opened 2023-11-20
Add get_test_results

hinthornw:


python/langsmith/client.py
```diff
@@ -1165,6 +1165,50 @@ def read_project(
...
+
+        Returns
+        -------
+        pd.DataFrame
+            A dataframe containing the test results.
+        """
+        import pandas as pd  # type: ignore
+
+        runs = self.list_runs(
+            project_id=project_id, project_name=project_name, execution_order=1
```
eyurtsev, line 1187: is this paginated?

python/langsmith/client.py
```diff
@@ -1165,6 +1165,50 @@ def read_project(
             **response.json(), _host_url=self._host_url
         )
 
+    def get_test_results(
+        self,
+        project_id: Optional[ID_TYPE] = None,
```
eyurtsev, line 1170:
```suggestion
        *,
        project_id: Optional[ID_TYPE] = None,
```

python/langsmith/client.py
```diff
@@ -1165,6 +1165,50 @@ def read_project(
...
+        *,
```
eyurtsev, line 1171:
```suggestion
```

python/langsmith/client.py
eyurtsev: not sure if you like, but i like forcing named invocation for whatever could be a literal since it makes user code more readable (but longer to type), but I like typing more code

eyurtsev (approved):
Thank you thank you!!

python/langsmith/client.py
hinthornw: ya

python/langsmith/client.py
hinthornw: It's more consistent with the rest of the API so ya good point. I've grown tired of typing "project_id" so often

python/langsmith/client.py
```diff
@@ -1165,6 +1165,50 @@ def read_project(
...
+                example_outputs = self.read_example(r.reference_example_id).outputs
+                for k, v in (example_outputs or {}).items():
+                    row[f"reference.{k}"] = v
+            if r.feedback_stats:
+                for k, v in r.feedback_stats.items():
+                    row[f"feedback.{k}"] = v.get("avg")
+            execution_time = (
+                (r.end_time - r.start_time).total_seconds() if r.end_time else None
+            )
+            row["execution_time"] = execution_time
```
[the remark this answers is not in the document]
hinthornw: Works for me

python/langsmith/client.py
eyurtsev: but it's an opportunity to not think about anything since and let muscle memory engage for a few seconds

PR #342, opened 2024-01-08
Batch Tracer Endpoint & Tracer Sampling

hinthornw:


python/langsmith/client.py
```diff
@@ -690,13 +715,64 @@ def upload_csv(
             **result, _host_url=self._host_url, _tenant_id=self._get_tenant_id()
         )
 
+    def _run_transform(
```
jacoblee93, line 718: How about calling this `_prepare_run_create_params()`?

jacoblee93: Also, does it need to be an instance method?

python/langsmith/client.py
```diff
@@ -690,13 +715,64 @@ def upload_csv(
...
+            run_create = cast(dict, run)
+        if isinstance(run["id"], str):
+            run["id"] = uuid.UUID(run["id"])
+        if "inputs" in run_create:
+            run_create["inputs"] = _hide_inputs(run_create["inputs"])
+        if "outputs" in run_create:
+            run_create["outputs"] = _hide_outputs(run_create["outputs"])
+        return run_create
+
+    def _insert_runtime_env(self, runs: Sequence[dict]) -> None:
```
jacoblee93, line 733: Same as above, does this need to be an instance method? And naming could be clearer

python/langsmith/client.py
```diff
@@ -690,13 +715,64 @@ def upload_csv(
...
 
+    def _run_transform(
+        self, run: Union[ls_schemas.Run, dict, ls_schemas.RunLikeDict]
+    ) -> dict:
+        if hasattr(run, "dict") and callable(getattr(run, "dict")):
+            run_create = run.dict()  # type: ignore
+        else:
+            run_create = cast(dict, run)
+        if isinstance(run["id"], str):
+            run["id"] = uuid.UUID(run["id"])
```
jacoblee93, line 726: Shouldn't this be `run_create`?

python/langsmith/client.py
jacoblee93: `_merge_runtime_env_into_create_run_param_list`?

python/langsmith/client.py
hinthornw: No it doesn't

python/langsmith/client.py
hinthornw: Or update

PR #350, opened 2024-01-10
fix url hosting url parse

snlamm:
The current logic to fetch the langsmith host url unintentionally turns the default langsmith url of `https://api.smith.langchain.com` into an invalid url of `https:/.smith.langchain.com` because it matches on the /api portion.

This PR changes the logic to exempt those cases where the subdomain begins with `api` so that the correct host url is returned.


snlamm:
cc @hinthornw

hinthornw (approved):
Thank you! I'll add some unit tests

PR #365, opened 2024-01-18
Change LANGCHAIN_REVISION_ID to revision_id; use `git describe` as backup; don't overwrite metadata from env variables

samnoyes:


python/langsmith/env/_runtime_env.py
```diff
@@ -164,11 +164,39 @@ def get_langchain_env_var_metadata() -> dict:
...
+
+    try:
+        return (
+            subprocess.check_output(
+                ["git", "describe", "--tags", "--dirty"], stderr=subprocess.DEVNULL
+            )
+            .strip()
+            .decode()
+        )
+    except Exception:
```
hinthornw, line 198: I think we should catch BaseException

python/langsmith/env/_runtime_env.py
```diff
@@ -164,11 +164,39 @@ def get_langchain_env_var_metadata() -> dict:
...
+
+
+@functools.lru_cache(maxsize=1)
+def _get_default_revision_id() -> Optional[str]:
+    """Get the default revision ID based on `git describe`."""
+    import subprocess
+
+    try:
+        return (
+            subprocess.check_output(
```
hinthornw, line 192: Any reason not to use `_exec_git`?

hinthornw (approved):
This looks alright. 1 request and 1 question

python/langsmith/env/__init__.py
```diff
@@ -27,4 +27,5 @@
     "get_runtime_environment",
     "get_system_metrics",
     "get_git_info",
+    "exec_git",
```
hinthornw, line 30: Let's not export this on the `env` level (don't want to update our public API). We can still use with other internal files though

python/langsmith/env/_git.py
```diff
@@ -12,16 +12,12 @@
...
     try:
         return subprocess.check_output(
             ["git"] + command, encoding="utf-8", stderr=subprocess.DEVNULL
         ).strip()
-    except FileNotFoundError:
-        logger.warning("git is not installed, or cannot be found in PATH")
-        return None
-    except subprocess.CalledProcessError as e:
-        logger.debug(f"Error running git command: {e}")
+    except (FileNotFoundError, subprocess.CalledProcessError, BaseException):
```
hinthornw, line 20: Could just do BaseException at this point! Captures all cases

python/langsmith/env/_runtime_env.py
```diff
@@ -164,11 +165,33 @@ def get_langchain_env_var_metadata() -> dict:
...
+    elif default_revision_id := _get_default_revision_id():
+        langchain_metadata["revision_id"] = default_revision_id
+
+    return langchain_metadata
+
+
+@functools.lru_cache(maxsize=1)
+def _get_default_revision_id() -> Optional[str]:
+    """Get the default revision ID based on `git describe`."""
+    import subprocess
```
hinthornw, line 189: non longer needed

hinthornw (approved):

PR #370, opened 2024-01-19
Add optional tracing sampling rate to langsmith sdk

nfcampos:
- controlled by LANGCHAIN_TRACING_SAMPLING_RATE
- if a POST was allowed then the matching PATCH is allowed too
- applied on both single and batch tracing endpoints

### DIFF
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -21,6 +21,7 @@ from typing import (
     Callable,
     DefaultDict,
     Dict,
+    Iterable,
     Iterator,
     List,
     Mapping,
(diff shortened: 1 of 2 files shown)


python/langsmith/client.py
```diff
@@ -202,6 +203,25 @@ def _validate_api_key_if_hosted(api_url: str, api_key: Optional[str]) -> None:
...
+    float
+        The tracing sampling rate.
+    """
+    sampling_rate_str = os.getenv("LANGCHAIN_TRACING_SAMPLING_RATE")
+    if sampling_rate_str is None:
+        return None
+    sampling_rate = float(sampling_rate_str)
+    if sampling_rate < 0 or sampling_rate > 1:
+        raise ls_utils.LangSmithUserError(
+            "LANGCHAIN_TRACING_SAMPLING_RATE must be between 0 and 1 if set"
```
hinthornw, line 220:
```suggestion
            "LANGCHAIN_TRACING_SAMPLING_RATE must be between 0 and 1 if set."
            f" Got: {sampling_rate}"
```

python/tests/integration_tests/test_runs.py
```diff
@@ -72,13 +72,15 @@ def my_chain_run(text: str):
...
-    for _ in range(5):
+    for _ in range(15):
         try:
             runs = list(langchain_client.list_runs(project_name=project_name))
             assert len(runs) == 3
             break
         except (ls_utils.LangSmithError, AssertionError):
             time.sleep(1)
+    else:
+        raise AssertionError("Failed to get runs after 5 attempts.")
```
hinthornw, line 83:
```suggestion
        raise AssertionError("Failed to get runs after 15 attempts.")
```

I need to fix this logic up heh thanks for the patience

hinthornw (approved):

PR #372, opened 2024-01-20
Add auto_batch_tracing modality for Client

nfcampos:
- if on, starts a background thread that batches inserts/updates
- only applies to insert/updates w/ trace_id and dotted_order
- after release, bump sdk version here https://github.com/langchain-ai/langchain/pull/16305

### DIFF
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -11,7 +11,6 @@ import json
 import logging
 import os
 import random
-import signal
 import socket
 import sys
 import threading
@@ -185,10 +184,6 @@ def close_session(session: requests.Session) -> None:
     session.close()
 
 
-def signal_exit(event: threading.Event, signum: int, frame: Any) -> None:
-    event.set()
-
-
 def _validate_api_key_if_hosted(api_url: str, api_key: Optional[str]) -> None:
     """Verify API key is provided if url not localhost.
 
@@ -344,11 +339,12 @@ class Client:
         # Initialize auto batching
         if auto_batch_tracing:
             self.tracing_queue: Optional[Queue] = Queue()
-            exit_event = threading.Event()
 
-            signal.signal(signal.SIGINT, functools.partial(signal_exit, exit_event))
             threading.Thread(
-                target=_tracing_thread_func, args=(weakref.ref(self), exit_event)
+                target=_tracing_thread_func,
+                # arg must be a weakref to self to avoid the Thread object
+                # preventing garbage collection of the Client object
+                args=(weakref.ref(self),),
             ).start()
         else:
             self.tracing_queue = None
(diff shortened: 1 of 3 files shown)


python/langsmith/client.py
```diff
@@ -837,6 +896,7 @@ def batch_ingest_runs(
         update: Optional[
             Sequence[Union[ls_schemas.Run, ls_schemas.RunLikeDict, Dict]]
         ] = None,
+        pre_sampled: bool = False,
```
hinthornw, line 868: We should document this - not super obvious what the effect is just by looking at the signature.

And should this be in the public API anyways? We could put the shared logic in a private method and then call that from our queue

python/langsmith/client.py
```diff
@@ -810,6 +861,14 @@ def create_run(
             "run_type": run_type,
             "execution_order": execution_order if execution_order is not None else 1,
         }
+        if not self._filter_for_sampling([run_create]):
+            return
+        if (
+            self.tracing_queue is not None
+            and run_create.get("trace_id") is not None
```
hinthornw, line 837: Can we document why we have these latter conditions?

hinthornw (commented):
We'd have to add the trace_id and dotted order to trigger it right?

python/langsmith/client.py
nfcampos: because the backend requires calls to batch endpoint to have these 2 attrs

hinthornw: ya - it's not suuper obvious without having the BE context though

python/langsmith/client.py
nfcampos: I've documented, left as public as could be useful to someone else conceivably

python/langsmith/client.py
nfcampos: Added comments and exceptions thrown if missing

hinthornw (approved):

python/langsmith/client.py
```diff
@@ -3036,3 +3097,52 @@ def _evaluate_strings(self, prediction, reference=None, input=None, **kwargs) ->
...
+    client = client_ref()
+    if client is None:
+        return
+    tracing_queue = client.tracing_queue
+    assert tracing_queue is not None
+
+    # loop until
+    while (
+        # the main thread dies
+        threading.main_thread().is_alive()
```
hinthornw, line 3138: If we create the client in a separate thread, I guess this could keep running even after the calling thread dies?
We shouldn't be doing this I'm just making sure I understand

nfcampos: No that's what the ref count is for. This main thread check is to stand in for a signal trap, if the main thread has died then we should die to let the process die

hinthornw (approved):

python/langsmith/client.py
```diff
@@ -266,6 +270,13 @@ def _as_uuid(value: ID_TYPE, var: str) -> uuid.UUID:
         ) from e
 
 
+@dataclass(order=True)
```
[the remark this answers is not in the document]
nfcampos: Ah straight from python docs lol

PR #376, opened 2024-01-23
respect revision_id if it is passed into kwargs of create_run

samnoyes:


python/langsmith/client.py
```diff
@@ -747,6 +747,8 @@ def create_run(
         metadata.update(
             {k: v for k, v in langchain_metadata.items() if k not in metadata}
         )
+        if "revision_id" in kwargs:
```
hinthornw, line 750: Should probably remove it before making the "create" dict

hinthornw (approved):

hinthornw (approved):

PR #382, opened 2024-01-25
Autoscale background threads for tracer auto batching

nfcampos:


python/langsmith/client.py
```diff
@@ -3138,11 +3138,11 @@ def _evaluate_strings(self, prediction, reference=None, input=None, **kwargs) ->
 
 
 def _tracing_thread_drain_queue(
-    tracing_queue: Queue, limit: Optional[int] = None
+    tracing_queue: Queue, limit: int = 100, block: bool = True
 ) -> List[TracingQueueItem]:
     next_batch: List[TracingQueueItem] = []
     try:
-        while item := tracing_queue.get(block=True, timeout=0.25):
+        while item := tracing_queue.get(block=block, timeout=0.25):
```
jakerachleff, line 3145: This logic is pretty confusing, summarizing for my own understanding:

1. We check the queue to see if anything is there. **if not**: wait 250 ms more and if nothing is there, return all the accumulated items. **if so**: move to step to
2. We pop an item off the queue and add it to the batch
3. If we hit the limit on batch size, return

python/langsmith/client.py
```diff
@@ -3163,24 +3163,63 @@ def _tracing_thread_handle_batch(
...
     while (
         # the main thread dies
         threading.main_thread().is_alive()
         # or we're the only remaining reference to the client
-        and sys.getrefcount(client) > 3
+        and sys.getrefcount(client) > 3 + len(sub_threads)
         # 1 for this func, 1 for getrefcount, 1 for _get_data_type_cached
     ):
-        if next_batch := _tracing_thread_drain_queue(tracing_queue, 100):
+        print("im looping", tracing_queue.qsize())
```
jakerachleff: :)

python/langsmith/client.py
```diff
@@ -3163,24 +3163,63 @@ def _tracing_thread_handle_batch(
...
+        for thread in sub_threads:
+            if not thread.is_alive():
+                sub_threads.remove(thread)
+        if tracing_queue.qsize() > 1000:
```
jakerachleff, line 3186: nit: let's put this in a variable...

python/langsmith/client.py
jakerachleff: Is there a degenerate case where you only get a run added every 200ms that could lead to something like 200ms * 100 = 20s run delays?

How can we make this debuggable?

python/langsmith/client.py
```diff
@@ -3163,24 +3163,63 @@ def _tracing_thread_handle_batch(
...
 
-def _tracing_thread_func(client_ref: weakref.ref[Client]) -> None:
+def _tracing_control_thread_func(client_ref: weakref.ref[Client]) -> None:
     client = client_ref()
     if client is None:
         return
     tracing_queue = client.tracing_queue
     assert tracing_queue is not None
 
+    sub_threads: List[threading.Thread] = []
```
jakerachleff, line 3173: Can we have a max size here?

python/langsmith/client.py
hinthornw: will want to rm

python/langsmith/client.py
```diff
@@ -3163,24 +3163,63 @@ def _tracing_thread_handle_batch(
...
+        for thread in sub_threads:
+            if not thread.is_alive():
+                sub_threads.remove(thread)
+        if tracing_queue.qsize() > 1000:
+            new_thread = threading.Thread(
+                target=_tracing_sub_thread_func, args=(weakref.ref(client),)
+            )
+            sub_threads.append(new_thread)
+            new_thread.start()
+        if next_batch := _tracing_thread_drain_queue(tracing_queue):
```
jakerachleff, line 3192: How does this get assigned to the right thread?

nfcampos: not sure I understand the comment, this function runs all in the same thread, so any thing in this function is happening in this one thread

python/langsmith/client.py
nfcampos: I can make it so only the first call blocks

hinthornw, line 3145: Seems we are always setting block=False: is there a reason for it being configurable then? And then it looks like `timeout` is ignored in this situation? (I'm not 100% sure here) https://docs.python.org/3/library/queue.html#queue.Queue.get

nfcampos: There is a reason why I made it like this, if we special case the first item as you suggest, then it ends up creating a number of very small batches whenever requests into the queue are ramping up

hinthornw (commented):
I think this makes sense, but I'm not 100% confident that the queue timeout will behave as desired here

python/langsmith/client.py
nfcampos: block defaults to True both in python lib and our code

nfcampos: Pushed a commit which is a compromise

python/langsmith/client.py
nfcampos: I'd rather not put in a var something I really don't want to ever use again

jakerachleff: sorry - in a constant that has a fixed name like `_QUEUE_SIZE_TO_SPAWN_NEW_THREAD` or something silly like that

hinthornw (approved):

python/langsmith/client.py
jakerachleff: ah! So the main thread also drains the queue and handles requests alongside the subthreads, which it spawns to do the same thing? 

It's weird we have a main thread and a subthread doing the same action, but in different places, but I guess that works

python/langsmith/client.py
hinthornw: Ah i missed the other calls - yupp looks good

python/langsmith/client.py
nfcampos: yea because I think most uses of this don't need the other threads, can change if we think its better

jakerachleff: so theoretically could you round robin dequeue records from all the different threads as they're accumulated? Nothing wrong with that, just thinking through how this works

python/langsmith/client.py
nfcampos: done

python/langsmith/client.py
nfcampos: done

python/langsmith/client.py
nfcampos: there is a single queue shared by all consumer threads. increasing nr of threads just increases the rate at which we drawdown the single queue

python/langsmith/client.py
nfcampos: FYI after some testing this compromise loses a lot of the effectivness of the batching behavior, ie ends up creating too many small batches, I think a delay is needed

jakerachleff: makes sense to me

jakerachleff (approved):

PR #385, opened 2024-01-26
Wrap OpenAI

hinthornw:


python/langsmith/run_helpers.py
```diff
@@ -483,9 +493,12 @@ def generator_wrapper(
...
             selected_wrapper: Callable = async_generator_wrapper
-        elif inspect.iscoroutinefunction(func):
-            selected_wrapper = async_wrapper
-        elif inspect.isgeneratorfunction(func):
+        elif is_async(func):
+            if reduce_fn:
+                selected_wrapper = async_generator_wrapper
+            else:
+                selected_wrapper = async_wrapper
+        elif reduce_fn or inspect.isgeneratorfunction(func):
```
hinthornw, line 501: Use the generator wrapper if user explicitly provides a reduce function

python/langsmith/wrappers/_openai.py
```diff
@@ -0,0 +1,90 @@
...
+    @functools.wraps(original_create)
+    def create(*args, stream: bool = False, **kwargs):
+        decorator = run_helpers.traceable(
+            name=name, run_type="llm", reduce_fn=reduce_fn if stream else None
+        )
+
+        return decorator(original_create)(*args, stream=stream, **kwargs)
+
+    @functools.wraps(original_create)
+    async def acreate(*args, stream: bool = False, **kwargs):
```
hinthornw, line 60: Could probably simplify this. Main reason for it is that the response type may or may not be a generator, and we don't handle that well by default

PR #389, opened 2024-01-29
Make start_time type mandatory in create run params

jacoblee93:
CC @nfcampos @hinthornw

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -93,7 +93,7 @@ interface CreateRunParams {
   run_type: string;
   execution_order?: number;
   id?: string;
-  start_time?: number;
+  start_time: number;
   end_time?: number;
   extra?: KVMap;
   error?: string;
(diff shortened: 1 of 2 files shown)


js/src/client.ts
```diff
@@ -93,7 +93,7 @@ interface CreateRunParams {
   run_type: string;
   execution_order?: number;
   id?: string;
-  start_time?: number;
+  start_time: number;
```
hinthornw, line 96: Why not auto-create in the `createRun` call? seems like an arg users shouldn't have to worry about 99% of the time

nfcampos: That works too, one issue we see is that the current behavior (ie. created in the server) ends up with runs getting inconsistent start times, because it's not guaranteed the server will handle requests in the order they are sent. Creating it here in the sdk code feels better

hinthornw: Yeah I agree. Probably good to add a default in the SDK before putting in the queue

hinthornw:
Gonna close as superseded by other PRs

PR #392, opened 2024-01-31
Add trace ID + dotted order

hinthornw:
Also
- Delete execution_order
- Add default start_time in js


js/src/run_trees.ts
```diff
@@ -51,10 +47,29 @@ export class RunTree implements BaseRun {
...
+    if (!this.trace_id) {
+      if (this.parent_run) {
+        this.trace_id = this.parent_run.trace_id;
+      } else {
+        this.trace_id = this.id;
+      }
+    }
+    if (!this.dotted_order) {
+      const currentDottedOrder =
+        new Date(this.start_time).toISOString() + this.id;
```
nfcampos, line 65: that's not the exact format of dotted order, see tracer

python/langsmith/schemas.py
```diff
@@ -271,14 +271,16 @@ class Run(RunBase):
...
-    """The execution order of the run within a run trace."""
     _host_url: Optional[str] = PrivateAttr(default=None)
 
     def __init__(self, _host_url: Optional[str] = None, **kwargs: Any) -> None:
         """Initialize a Run object."""
+        if not kwargs.get("trace_id"):
+            kwargs = {"trace_id": kwargs.get("id"), **kwargs}
         super().__init__(**kwargs)
         self._host_url = _host_url
+        if not self.dotted_order.strip():
```
nfcampos, line 282: can this be wrong if run has parent id without dotted order?

hinthornw: hm yeah probably. I'll update

PR #394, opened 2024-02-01
Update default python serialization

hinthornw:
Fall back to repr when you reach a certain depth of regular class traversal

### DIFF
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -151,11 +151,19 @@ def _default_retry_config() -> Retry:
     return Retry(**retry_params)  # type: ignore
 
 
-def _serialize_json(obj: Any) -> Any:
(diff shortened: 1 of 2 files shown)


python/langsmith/client.py
```diff
@@ -151,11 +151,19 @@ def _default_retry_config() -> Retry:
     return Retry(**retry_params)  # type: ignore
 
 
-def _serialize_json(obj: Any) -> Any:
+_PRIMITIVE_TYPES = (str, int, float, bool)
+_MAX_DEPTH = 3
+
+
+def _serialize_json(obj: Any, depth: int = 0) -> Any:
```
nfcampos, line 158: do we need to handle lists or dicts or tuples here?

hinthornw: Updated

PR #404, opened 2024-02-05
Wfh/use system batch config

hinthornw:


python/langsmith/client.py
```diff
@@ -3250,6 +3252,23 @@ def _tracing_thread_handle_batch(
...
 
+def _get_batch_ingest_config(client: Client) -> Dict[str, Any]:
+    try:
+        info = client.info
+        if not info:
+            return {}
+        if not info.batch_ingest_config:
+            return {}
+        return info.batch_ingest_config
+    except BaseException:
```
langchain-infra, line 3263: should we turn off if we 404 on info? to avoid breaking old langsmith clients?

hinthornw:
Rather than continuing to attempt?

python/langsmith/client.py
```diff
@@ -3250,6 +3252,23 @@ def _tracing_thread_handle_batch(
 _AUTO_SCALE_DOWN_NEMPTY_TRIGGER = 4
 
 
+def _get_batch_ingest_config(client: Client) -> Dict[str, Any]:
+    try:
+        info = client.info
```
langchain-infra, line 3257: so this will still fail on old on-prem isntances right? where we 404-> empty info dict -> uses default config? can we just fallback to not batching in this case?

nfcampos:
Yea when 404 assume can't batch

langchain-infra (approved):
lgtm

PR #411, opened 2024-02-08
Adds batch tracing support check for JS

jacoblee93:
@hinthornw @nfcampos I think this is better than doing a check on startup?


hinthornw (dismissed):

jacoblee93:
@hinthornw spoke with @nfcampos today and this is not going to work - we should use the `/info` endpoint.

jacoblee93:
Ok, this is now ready for a look @hinthornw and @nfcampos!

hinthornw (approved):

PR #441, opened 2024-02-14
Immediately trigger a batch send on root run end in JS

jacoblee93:
@nfcampos @hinthornw


jacoblee93:
I've also bumped version to 0.1.0 - @hinthornw anything special to be done to release this?

js/src/client.ts
```diff
@@ -705,7 +736,14 @@ export class Client {
...
       data.dotted_order !== undefined
     ) {
-      this.appendRunCreateToAutoBatchQueue({ action: "update", item: data });
+      if (run.end_time !== undefined && data.parent_run_id === undefined) {
+        // Trigger a batch as soon as a root trace ends and block to ensure trace finishes
+        // in serverless environments.
+        await this.processRunOperation({ action: "update", item: data }, true);
+        return;
+      } else {
+        void this.processRunOperation({ action: "update", item: data });
```
hinthornw, line 745: Cool so this one is non blocking then?

hinthornw (approved):

hinthornw (approved):

js/src/client.ts
jacoblee93: Yes, we only want to block on root run ends

PR #445, opened 2024-02-15
Release

hinthornw:


python/langsmith/run_helpers.py
```diff
@@ -239,6 +241,16 @@ def _setup_run(
 
 R = TypeVar("R", covariant=True)
 
+_VALID_RUN_TYPES = {
```
langchain-infra, line 244: should we have a default run type? or maybe an "unknown"?

langchain-infra (approved):

nfcampos (approved):

python/langsmith/run_helpers.py
hinthornw: Background of this is I originally had it as required then realized everything should just be chain by default, but it was already a positional arg and ididn't want to break things

PR #451, opened 2024-02-16
Add OpenAI traceable wrapper for JS

dqbd:


hinthornw (approved):
We should probably add tests for this to check when streaming and not, etc., right?

PR #461, opened 2024-02-21
Update Handoff

hinthornw:
Depends on https://github.com/langchain-ai/langchainjs/pull/4477 

Would update the tests to reflect this.

### DIFF
--- a/js/src/run_trees.ts
+++ b/js/src/run_trees.ts
@@ -100,8 +100,9 @@ export class RunTree implements BaseRun {
   dotted_order: string;
 
   constructor(config: RunTreeConfig) {
-    const defaultConfig = RunTree.getDefaultConfig(config.client);
-    Object.assign(this, { ...defaultConfig, ...config });
+    const defaultConfig = RunTree.getDefaultConfig();
+    const client = config.client ?? new Client();
+    Object.assign(this, { ...defaultConfig, ...config, client });
     if (!this.trace_id) {
       if (this.parent_run) {
         this.trace_id = this.parent_run.trace_id ?? this.id;
@@ -164,7 +165,7 @@ export class RunTree implements BaseRun {
     return rt;
   }
 
-  private static getDefaultConfig(client?: Client): object {
+  private static getDefaultConfig(): object {
     return {
       id: uuid.v4(),
       run_type: "chain",
@@ -181,7 +182,6 @@ export class RunTree implements BaseRun {
       serialized: {},
       inputs: {},
       extra: {},
-      client: client ?? new Client({}),
     };
   }
 
--- a/js/src/run_trees.ts
+++ b/js/src/run_trees.ts
@@ -65,8 +65,8 @@ export interface RunnableConfigLike {
 }
 
 interface CallbackManagerLike {
-  _parentRunId?: string;
   handlers: TracerLike[];
+  getParentRunId(): string | undefined;
 }
 
 interface TracerLike {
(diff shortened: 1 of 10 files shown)


js/src/run_trees.ts
```diff
@@ -41,16 +43,52 @@ export interface RunTreeConfig {
...
+   * Callbacks for this call and any sub-calls (eg. a Chain calling an LLM).
+   * Tags are passed to all callbacks, metadata is passed to handle*Start callbacks.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  callbacks?: any;
+}
+
+interface CallbackManagerLike {
+  handlers: TracerLike[];
+  getParentRunId(): string | undefined;
```
jacoblee93, line 69: Would make this an optional param:

`getParentRunId?: () => string | undefined`

To avoid compatibility issues

js/src/run_trees.ts
```diff
@@ -41,16 +43,52 @@ export interface RunTreeConfig {
...
+}
+
+interface TracerLike {
+  name: string;
+}
+interface LangChainTracerLike extends TracerLike {
+  name: "langchain_tracer";
+  projectName: string;
+  getRun(id: string): RunTree | undefined;
```
jacoblee93, line 78: See above

jacoblee93:
I think seems reasonable but if you're using LangChain, couldn't you just wrap in a `RunnableLambda`?

js/src/tests/run_trees.int.test.ts
```diff
@@ -5,23 +5,27 @@ import {
   convertToDottedOrderFormat,
 } from "../run_trees.js";
 
-async function toArray<T>(iterable: AsyncIterable<T>): Promise<T[]> {
+export async function toArray<T>(iterable: AsyncIterable<T>): Promise<T[]> {
```
jacoblee93, line 8: We shouldn't export from these tests

js/src/traceable.ts
```diff
@@ -108,6 +115,9 @@ export function traceable<Func extends (...args: any[]) => any>(
     if (isRunTree(args[0])) {
       currentRunTree = args[0];
       rawInputs = args.slice(1) as Inputs;
+    } else if (isRunnableConfigLike(args[0])) {
```
jacoblee93, line 118: I think this was my idea but I'm not sure we need it

js/src/tests/run_trees.int.test.ts
jacoblee93: Better to make a separate utils file

js/src/run_trees.ts
```diff
@@ -213,3 +297,32 @@ export function isRunTree(x?: unknown): x is RunTree {
     typeof (x as RunTree).postRun === "function"
   );
 }
+
+function containsLangChainTracerLike(x: unknown): x is LangChainTracerLike[] {
```
jacoblee93, line 301: `x` can be undefined

js/src/run_trees.ts
```diff
@@ -84,6 +122,50 @@ export class RunTree implements BaseRun {
...
+      projectName = langChainTracer?.projectName;
+    }
+    const deduppedTags = [
+      ...new Set((parentRun?.tags ?? []).concat(config?.tags ?? [])),
+    ];
+    const dedupedMetadata = {
+      ...parentRun?.extra?.metadata,
+      ...config?.metadata,
+    };
+    console;
```
jacoblee93, line 155: Remove

js/src/run_trees.ts
```diff
@@ -84,6 +122,50 @@ export class RunTree implements BaseRun {
...
+    const rt = new RunTree({
+      name: props?.name ?? "<lambda>",
+      parent_run: parentRun,
+      tags: deduppedTags,
+      extra: {
+        metadata: dedupedMetadata,
+      },
+      project_name: projectName,
+    });
+    console.log("fromRunnableConfig", rt);
```
jacoblee93, line 165: Remove

jacoblee93 (commented):
Don't love adding an implicit reverse dep like this but if it'll make people's lives easier I'm ok with it

js/src/run_trees.ts
hinthornw: lol ty!

PR #463, opened 2024-02-22
Allow manually specified client in run tree with no env vars

jacoblee93:
We were initializing a client by default every time, which means if you don't have an env var set, you can't use the run tree.

Should we remove the thrown error altogether on client initialization? Why was that added in the first place?

@hinthornw @dqbd


hinthornw:
We could remove the thrown error ya thanks for flagging this

PR #465, opened 2024-02-22
Add 429 retries

hinthornw:


js/src/client.ts
```diff
@@ -695,9 +695,23 @@ export class Client {
       Accept: "application/json",
     };
 
+    const handle429 = async (response?: Response) => {
+      if (response?.status === 429) {
+        const retryAfter =
+          Number(response.headers.get("retry-after") ?? 30) * 1000;
```
jacoblee93, line 701: Think you might want `parseInt(retryAfter, 10)` here instead

js/src/client.ts
```diff
@@ -695,9 +695,23 @@ export class Client {
...
+          return true;
+        }
+      }
+      // Fall back to existing status checks
+      return false;
+    };
     try {
       const response = await this.caller.call(
         fetch,
+        handle429,
```
jacoblee93, line 714: This is a little weird since all the other args passed here are bound to the first function (`fetch` in this case) - why not just make a second caller instance?

PR #469, opened 2024-02-23
Add split batches

hinthornw:


python/langsmith/client.py
```diff
@@ -215,6 +216,24 @@ def _serialize_json(obj: Any, depth: int = 0) -> Any:
...
+        The object to serialize.
+    default : Callable[[Any], Any] or None, default=None
+        The default function to use for serialization.
+
+    Returns:
+    -------
+    str
+        The JSON formatted string.
+    """
+    return orjson.dumps(obj, default=_serialize_json)
```
nfcampos, line 234: lets add some of the flags like serialize numpy etc etc

hinthornw: is that applied before or after our default serialize json?

nfcampos: before I think

hinthornw: ah ya otherwise the passthrough options wouldn't make sense. thanks

PR #475, opened 2024-02-25
Respect batch byte size limits in JS client

jacoblee93:
CC @hinthornw @nfcampos @dqbd

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -266,6 +266,9 @@ export class Queue<T> {
   }
 }
 
+// 20 MB
+export const DEFAULT_SIZE_LIMIT_BYTES = 20_971_520;
+
 export class Client {
   private apiKey?: string;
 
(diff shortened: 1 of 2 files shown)


js/src/client.ts
```diff
@@ -711,6 +725,40 @@ export class Client {
...
+    const sizeLimitBytes =
+      this.serverInfo?.batch_ingest_config?.size_limit_bytes ??
+      DEFAULT_BATCH_SIZE_LIMIT_BYTES;
+    const batchChunks = {
+      post: [] as (typeof rawBatch)["post"],
+      patch: [] as (typeof rawBatch)["patch"],
+    };
+    let currentBatchSizeBytes = 0;
+    for (const k of ["post", "patch"]) {
+      const key = k as keyof typeof rawBatch;
```
hinthornw, line 737: oh this syntax is weird to me - cool

hinthornw (approved):

js/src/client.ts
jacoblee93: It's just shorthand so we don't have to cast everywhere

PR #482, opened 2024-03-01
Add presigned feedback method

hinthornw:


python/langsmith/client.py
```diff
@@ -2921,6 +2921,7 @@ def create_feedback(
         ] = ls_schemas.FeedbackSourceType.API,
         source_run_id: Optional[ID_TYPE] = None,
         feedback_id: Optional[ID_TYPE] = None,
+        feedback_config: Optional[ls_schemas.FeedbackConfig] = None,
         eager: bool = False,
```
nfcampos, line 2925: this doesn't do anything since a few months ago

hinthornw: ohhh

PR #485, opened 2024-03-03
Add generic JS wrapClient method

jacoblee93:
CC @dqbd @hinthornw


js/src/client.ts
```diff
@@ -853,6 +853,7 @@ export class Client {
         signal: AbortSignal.timeout(this.timeout_ms),
       }
     );
+    console.log(response);
```
dqbd, line 856: remove

jacoblee93:
Example trace: https://smith.langchain.com/public/323d9897-2014-44eb-99d1-aa22691fa34a/r

js/src/client.ts
jacoblee93: Thanks!

jacoblee93:
We might have to add an Anthropic key to CI if we want to do this @hinthornw

dqbd (approved):

hinthornw (approved):
Do we need some custom handling to show how to handle streaming or other outputs (make them look like OAI format)?

PR #486, opened 2024-03-03
add metadata to example, allow for creating feedback on the session level

hwchase17:


python/langsmith/client.py
```diff
@@ -2924,6 +2932,7 @@ def create_feedback(
         feedback_id: Optional[ID_TYPE] = None,
         feedback_config: Optional[ls_schemas.FeedbackConfig] = None,
         stop_after_attempt: int = 10,
+        session_id: Optional[ID_TYPE] = None,
```
hwchase17, line 2935: check if both none, xor

hinthornw: How about a separate method? `create_aggregate_feedback()`?

PR #517, opened 2024-03-12
fix: add adapters if not set

langchain-infra:


hinthornw (approved):

hinthornw:
Will have to fixup some tests

PR #519, opened 2024-03-13
feat: allow multiple write URLS for runs in SDK (python)

agola11:
* Allow user to set `LANGSMITH_ENDPOINT` and `LANGSMITH_API_KEY` (in addition to `LANGCHAIN_ENDPOINT` and `LANGCHAIN_API_KEY`). This priority will be former > latter > passed-in parameter.
* Allow user to set `LANGSMITH_RUNS_ENDPOINTS` environment variable or pass in `api_urls` which are a map of api urls to api keys. The first element will be used for all requests, whereas all elements will be used for writing (Runs only)

### DIFF
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -288,19 +288,21 @@ def _get_tracing_sampling_rate() -> float | None:
 
 
 def _get_api_key(api_key: Optional[str]) -> Optional[str]:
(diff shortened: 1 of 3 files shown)


hinthornw (commented):
Seems reasonable. I'll have to double check the batch ingestion logic but can land today

PR #542, opened 2024-03-23
[Python] `evaluate()` 

hinthornw:
- [X] base  `evaluate` api
- [X] support wrapper to support OTS evaluators from langchain
- [X] add examples

Feedback I'd love (all welcome)
- Desired behavior if the project you specify already exists
- 2-step API (2 functions? `predict, evaluate` vs. 1...)

To add in a second PR:
- async support
- hill-climbing as a first-class citizen


python/langsmith/evaluation/_runner.py
```diff
@@ -0,0 +1,534 @@
...
+    batch_evaluators: Optional[Sequence[BATCH_EVALUATOR_T]] = None,
+    metadata: Optional[dict] = None,
+    experiment: str | None = None,
+    max_concurrency: int | None = None,
+    client: langsmith.Client | None = None,
+    stream: bool = False,
+) -> ExperimentResults:
+    manager = _ExperimentManager(
+        data, client=client, metadata=metadata, experiment=experiment
+    ).start()
```
hinthornw, line 62: For `start()` if the project already exists, what is the desired behavior?

Current: `error`

Options:
- error
- predict again within the same project  (add multiple "trials" within the same test) - could warn.
- fetch runs from existing test and just apply evaluators to them
- create new project by adding a uuid afterwards
- something else

python/langsmith/evaluation/_runner.py
```diff
@@ -0,0 +1,534 @@
...
+]
+
+
+def evaluate(
+    target: TARGET_T,
+    /,
+    data: DATA_T,
+    evaluators: Optional[Sequence[EVALUATOR_T]] = None,
+    batch_evaluators: Optional[Sequence[BATCH_EVALUATOR_T]] = None,
+    metadata: Optional[dict] = None,
```
hinthornw, line 54: revision_id currently not special, assumed w/in metadata. Feels better than adding another argument here.

python/langsmith/evaluation/_runner.py
```diff
@@ -0,0 +1,534 @@
...
+    experiment: str | None = None,
+    max_concurrency: int | None = None,
+    client: langsmith.Client | None = None,
+    stream: bool = False,
```
hinthornw, line 58: @nfcampos I'm not aware of a good way to prevent the program from exiting if we always eagerly return with the streaming in a background thread.

python/langsmith/evaluation/_runner.py
hinthornw: Decided to always append a `:uid` bit

python/langsmith/run_helpers.py
```diff
@@ -178,13 +219,22 @@ def _setup_run(
...
+        project_cv  # From parent trace
         or langsmith_extra.get("project_name")  # at invocation time
         or container_input["project_name"]  # at decorator time
         or utils.get_tracer_project()  # default
     )
-    if not parent_run_ and not utils.tracing_is_enabled():
+    reference_example_id = langsmith_extra.get("reference_example_id")
+    id_ = langsmith_extra.get("run_id")
+    if (
+        not project_cv
```
hinthornw, line 232: Still trace if you're manually providing run ID, example, project, etc. via the context var

python/langsmith/evaluation/_runner.py
```diff
@@ -0,0 +1,895 @@
...
+    all_runs = {}
+    for run in runs:
+        if run.parent_run_id is not None:
+            treemap[run.parent_run_id].append(run)
+        else:
+            results.append(run)
+        all_runs[run.id] = run
+    for run_id, child_runs in treemap.items():
+        all_runs[run_id].child_runs = sorted(child_runs, key=lambda r: r.dotted_order)
+    return results
```
agola11, line 423: why don't we just map trace_id to list of runs, then sort each entry based on dotted order?

python/langsmith/evaluation/_runner.py
```diff
@@ -0,0 +1,895 @@
...
+    EvaluationResults,
+    RunEvaluator,
+    run_evaluator,
+)
+from langsmith.evaluation.integrations import LangChainStringEvaluator
+
+logger = logging.getLogger(__name__)
+
+PIPELINE_T = Callable[[dict], dict]
+TARGET_T = PIPELINE_T
```
agola11, line 47: why do we need both

python/langsmith/evaluation/_runner.py
```diff
@@ -0,0 +1,895 @@
...
+# dataset-name, dataset_id, or examples
+DATA_T = Union[str, uuid.UUID, Iterable[schemas.Example]]
+SUMMARY_EVALUATOR_T = Callable[
```
agola11, line 50: What is summary evaluator?

python/langsmith/evaluation/_runner.py
```diff
@@ -0,0 +1,895 @@
...
+            (result["example"] for result in r1),
+            experiment=self._experiment,
+            metadata=self._metadata,
+            client=self.client,
+            runs=(result["run"] for result in r2),
+            evaluation_results=(result["evaluation_results"] for result in r3),
+            aggregate_results=self._aggregate_results,
+        )
+
+    def with_batch_scores(
```
agola11, line 615: We should not use batch -- confusing terminology. Know this is internal api but still. Let's consolidate between batch / summary

python/langsmith/evaluation/_runner.py
```diff
@@ -0,0 +1,895 @@
...
+
+def _load_tqdm() -> Callable[[Iterable], Iterable]:
+    try:
+        from tqdm.auto import tqdm
+    except ImportError:
+        return lambda x: x
+    return tqdm
+
+
+class _ExperimentManager:
```
agola11, line 434: Need more comments everywhere within this class, logic is hard to follow

python/langsmith/evaluation/_runner.py
hinthornw: Still need to reconstruct the tree right?

I could also just default to only the roots, which is all people need in 90% of usage rn

python/langsmith/client.py
```diff
@@ -49,11 +49,12 @@
 from langsmith import env as ls_env
 from langsmith import schemas as ls_schemas
 from langsmith import utils as ls_utils
-from langsmith.evaluation import evaluator as ls_evaluator
 
 if TYPE_CHECKING:
     import pandas as pd  # type: ignore
 
+    from langsmith.evaluation import evaluator as ls_evaluator
```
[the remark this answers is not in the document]
hinthornw: Cyclic dep now

python/langsmith/evaluation/_runner.py
hinthornw: aggregate level over whole experiment

agola11 (approved):

PR #555, opened 2024-03-27
Add RAGAS

hinthornw:
See docstring for example usage


python/langsmith/evaluation/integrations/_base.py
```diff
@@ -0,0 +1,11 @@
+from typing import Protocol, runtime_checkable
+
+from langsmith.evaluation.evaluator import RunEvaluator
+
+
+@runtime_checkable
+class EvaluatorWrapper(Protocol):
+    """A protocol that represents an evaluator wrapper."""
+
+    def as_run_evaluator(self) -> RunEvaluator:
```
hinthornw, line 10: I'm somewhat ambivalent on whether we want this as a constructor or just to directly subclass. Maybe simpler to do the latter

PR #561, opened 2024-03-27
Ensure examples are properly sorted

hinthornw:
Also:
-  fix the ref dataset ID type of tracer schema
- fix a mistyping issue in the evaluate_run method

### DIFF
--- a/python/langsmith/evaluation/_runner.py
+++ b/python/langsmith/evaluation/_runner.py
@@ -138,6 +138,9 @@ def evaluate(
         ...     data=dataset_name,
         ...     evaluators=[accuracy],
         ...     summary_evaluators=[precision],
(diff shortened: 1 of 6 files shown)


python/langsmith/evaluation/_runner.py
```diff
@@ -926,7 +943,14 @@ def _get_run(r: run_trees.RunTree) -> None:
                 reference_example_id=example.id,
                 on_end=_get_run,
                 project_name=experiment_name,
-                metadata=metadata,
+                metadata={
+                    **metadata,
+                    "example_version": (
```
hinthornw, line 948: cc @nfcampos

langchain-infra (approved):

PR #563, opened 2024-03-27
Update serialization

hinthornw:
Better handle cases where a pydantic object raises an error when trying to serialize itself


python/langsmith/client.py
```diff
@@ -180,27 +180,31 @@ def _serialize_json(obj: Any, depth: int = 0) -> Any:
...
                     json_str = (
                         method(exclude_none=exclude_none) if exclude_none else method()
                     )
-                    return json.loads(json_str)
+                    if isinstance(json_str, str):
+                        return json.loads(json_str)
+                    return orjson.loads(_dumps_json(json_str, depth=depth + 1))
                 except Exception as e:
                     logger.debug(f"Failed to serialize {type(obj)} to JSON: {e}")
-                    return repr(obj)
```
hinthornw, line 195: This was the line that needed to be less aggressive

python/langsmith/client.py
```diff
@@ -180,27 +180,31 @@ def _serialize_json(obj: Any, depth: int = 0) -> Any:
...
+                    pass
+        all_attrs = {}
         if hasattr(obj, "__slots__"):
-            all_attrs = {slot: getattr(obj, slot, None) for slot in obj.__slots__}
```
hinthornw, line 197: These are just other changes made from observations

python/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "langsmith"
-version = "0.1.34"
+version = "0.1.36"
```
baskaryan, line 3: where did 0.1.35 go

hinthornw: Was an ill-timed manual release

baskaryan (approved):

PR #564, opened 2024-03-28
fix example metadata support

agola11:


js/src/client.ts
```diff
@@ -1867,7 +1869,7 @@ export class Client {
   public async createExample(
     inputs: KVMap,
     outputs: KVMap,
-    { datasetId, datasetName, createdAt, exampleId }: CreateExampleOptions
```
hinthornw, line 1870: Should also add  to `createExamples` below

hinthornw (approved):

PR #570, opened 2024-04-02
getattr for reduced cold start times

efriis:


python/langsmith/__init__.py
```diff
@@ -1,17 +1,43 @@
...
-
-from langsmith.client import Client
-from langsmith.evaluation.evaluator import EvaluationResult, RunEvaluator
-from langsmith.run_helpers import trace, traceable
-from langsmith.run_trees import RunTree
+from typing import Any
+
+
+def __getattr__(name: str) -> Any:
+    if name == "__version__":
```
hinthornw, line 7: Let's add the imports in a `if TYPE_CHECKING` block as well for a better IDE exp

hinthornw (approved):

python/langsmith/__init__.py
efriis: Was there a reason you didn't include the evaluation ones? Can add in a followup (also want to remove dir from below one)

hinthornw: Mainly bcs I think it's fine to encourage importing from `langsmith.evaluation` for evals related stuff, but i should probably have the type checking just have it reflect what's available in general

PR #571, opened 2024-04-03
Add select keys

hinthornw:


js/src/client.ts
```diff
@@ -1132,6 +1143,44 @@ export class Client {
...
+      "completion_tokens",
+      "dotted_order",
+      "end_time",
+      "error",
+      "events",
+      "extra",
+      "feedback_stats",
+      "first_token_time",
+      "id",
+      "in_dataset",
```
nfcampos, line 1159: you actually have this in pydantic model?

js/src/client.ts
```diff
@@ -1132,6 +1143,44 @@ export class Client {
...
+      "inputs",
+      "inputs_s3_urls",
+      "last_queued_at",
```
nfcampos, line 1162: this is in pydantic model?

nfcampos (approved):

PR #580, opened 2024-04-08
Add a testing decorator

hinthornw:
Basically, each unit test <-> example, and a single test run creates or updates a dataset and generates runs + metrics

Improvements to be made:
- [x] Fix pytest warning that it's returning something other than None
- [x] Support for parametrization
- [x] Update example if inputs/outputs differ 

Later:
- Printout the test suite results
- Print out the examples updated/created/deleted/unused
- caching
- etc.

### DIFF
--- a/python/langsmith/_testing.py
+++ b/python/langsmith/_testing.py
@@ -10,7 +10,6 @@ import os
 import threading
 import uuid
 import warnings
-from enum import Enum
 from typing import Any, Callable, Optional, Sequence, Tuple, TypeVar, overload
 
 from typing_extensions import TypedDict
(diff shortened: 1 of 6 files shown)


python/langsmith/_testing.py
```diff
@@ -0,0 +1,384 @@
...
+            client.update_example(
+                example_id=example.id,
+                inputs=inputs,
+                outputs=outputs,
+            )
+    except ls_utils.LangSmithNotFoundError:
+        example = client.create_example(
+            example_id=example_id,
+            inputs=inputs,
+            # TODO: Handle output_keys
```
nfcampos, line 177: aren't you handling this above already?

hinthornw: ah ya i can rm the todo!

nfcampos (approved):

PR #589, opened 2024-04-10
Add test caching

hinthornw:


python/langsmith/_testing.py
```diff
@@ -188,7 +211,7 @@ def wrapper(*test_args, **test_kwargs):
 
 def _get_experiment_name() -> str:
     # TODO Make more easily configurable
-    prefix = ls_utils.get_tracer_project(False) or "TestSuite"
+    prefix = ls_utils.get_tracer_project(False) or "TestSuiteResult"
```
hinthornw, line 214: Think naming makes more sense as a "result" for the experiment and "Test Suite" for the dataset it creates

nfcampos (approved):

benissimo:
Does this belong as part of LangSmith? While caching is certainly helpful, it's difficult to solve this for all users. Nothing prevents users from using VCR directly, which would give them more flexibility.

I notice that the cassettes don't use `decode_compressed_response: true`, so gzipped response bodies are saved as binary data in the cassettes. That results in "security through obscurity" -- if the response contained any sensitive data, it wouldn't be immediately obvious, but anyone can decompress it and see the original values. I find it safer to decompress the data in cassettes as it makes it easier for a developer to review the contents of a cassette before pushing a commit.

PR #596, opened 2024-04-10
Expectations

hinthornw:
- [x] change default dataset creation (should be unique per file)
- [x] change env vars to use `LANGSMITH_` prefix
Want feedback on:

- [x] ergonomics: is this a general UX we want to support
- [x] imports: should i re-implement string, embedding distance, etc.?
- [x] Do we want default implementations? 
- [x] Any other we ought to include at the outset?
- [x] I could also do a general `expect(value).is(...)` or something: want anything super generic like that?

Example:

```python
@unit(inputs=x, outputs=y)
def test_output_semantically_close():
    response = oai_client.chat.completions.create(
        model="gpt-3.5-turbo",
        messages=[
            {"role": "system", "content": "You are a helpful assistant."},
            {"role": "user", "content": "Say hello!"},
        ],
    )
    # The embedding_distance call logs the embedding distance to LangSmith
    expect.embedding_distance(
        prediction=response.choices[0].message.content,
        reference="Hello!",
        # The following optional assertion logs a
        # pass/fail score to LangSmith
        # and raises an AssertionError if the assertion fails.
    ).to_be_less_than(0.5)
    # Compute damerau_levenshtein distance
    expect.string_distance(
        prediction=response.choices[0].message.content,
        reference="Hello!",
        # And then log a pass/fail score to LangSmith
    ).to_be_less_than(0.5)
```

The idea is it's still an easy onramp for developers to quickly write some scoring functions and get it running regularly in CI


python/langsmith/_testing.py
```diff
@@ -114,13 +114,13 @@ def unit(*args: Any, **kwargs: Any) -> Callable:
         Caching is faster if you install libyaml. See
         https://vcrpy.readthedocs.io/en/latest/installation.html#speed for more details.
 
-        >>> os.environ["LANGCHAIN_TEST_CACHE"] = "tests/cassettes"
+        >>> # os.environ["LANGCHAIN_TEST_CACHE"] = "tests/cassettes"
```
hinthornw, line 117: It's playing funky with doctest - looking into it more tonight

python/langsmith/_expect.py
```diff
@@ -0,0 +1,359 @@
...
+            AssertionError: If the expectation value does not contain the given value.
+        """
+        self._assert(
+            value in self.value,
+            f"Expected {self.key} to contain {value}, but it does not",
+            "to_contain",
+        )
+
+    # Custom assertions
+    def against(self, func: Callable, /) -> None:
```
hinthornw, line 193: @nfcampos I'm OK dropping or renaming  this one if it's premature. Basically wanted an easy way to log feedback in a unit test context

rlancemartin (approved):

PR #599, opened 2024-04-12
Support Context Propagation

hinthornw:
Client Side:

```
async def the_parent_function():
    async with AsyncClient(app=fake_app, base_url="http://localhost:8000") as client:
        headers = {}
        if span := get_current_span():
            headers.update(span.to_headers())
        return await client.post("/fake-route", headers=headers)

```

Server Side:

```
@fake_app.post("/fake-route")
async def fake_route(request: Request):
    with tracing_context(headers=request.headers):
        fake_function()
    return {"message": "Fake route response"}

```

If people like, we could add some fun middleware, but probably not necessary


agola11 (approved):
Looks good but would remove all of the code that switches from Run to Span, that's a bigger change that we're going to need to coordinate with the UI and docs

python/langsmith/run_helpers.py
```diff
@@ -39,19 +39,24 @@
...
 
-def get_current_run_tree() -> Optional[run_trees.RunTree]:
+def get_current_span() -> Optional[run_trees.Span]:
     """Get the current run tree."""
     return _PARENT_RUN_TREE.get()
 
 
+def get_parent_span() -> Optional[run_trees.Span]:
+    """Get the parent run tree."""
+    return _PARENT_RUN_TREE.get()
```
agola11, line 57: why do these two methods do the same thing?

hinthornw: copy pasta

PR #605, opened 2024-04-15
Async Eval Functions

hinthornw:


langchain-infra (approved):
lgtm

python/langsmith/evaluation/_arunner.py
```diff
@@ -192,6 +190,24 @@ async def aevaluate(
...
+        >>> results = asyncio.run(
+        ...     aevaluate(
+        ...         apredict,
+        ...         data=dataset_name,
+        ...         evaluators=[helpfulness],
+        ...         summary_evaluators=[precision],
+        ...         experiment_prefix="My Helpful Experiment",
+        ...     )
+        ... )  # doctest: +ELLIPSIS
+        View the evaluation results for experiment:...
```
agola11, line 210: nit: use await instead of `asyncio.run`

python/langsmith/evaluation/_arunner.py
```diff
@@ -482,12 +498,7 @@ async def awith_predictions(
 
     async def awith_evaluators(
         self,
-        evaluators: Sequence[
-            Union[
-                EVALUATOR_T,
-                RunEvaluator,
-            ]
-        ],
+        evaluators: Sequence[Union[EVALUATOR_T, AEVALUATOR_T]],
```
agola11, line 501: why not `RunEvaluator`?

python/langsmith/evaluation/evaluator.py
```diff
@@ -101,6 +102,9 @@ async def aevaluate_run(
         )
 
 
+_RUNNABLE_OUTPUT = Union[EvaluationResult, EvaluationResults, dict]
+
+
 class DynamicRunEvaluator(RunEvaluator):
```
agola11, line 108: not related to PR but docstring suggests we need to use `@run_evaluator` decorator but don't think that's true anymore?

python/langsmith/evaluation/_arunner.py
hinthornw: That breaks doctests since this is in the python repl

python/langsmith/evaluation/_arunner.py
hinthornw: That's contained within`EVALUATOR_T`

python/langsmith/evaluation/evaluator.py
hinthornw: You could initialize it directly or via the run_evaluator docstring.

We wrap user functions using this within the `evaluate()` function to promote them to `RunEvaluator`'s

PR #612, opened 2024-04-18
Fix Example Serialization

hinthornw:
(slightly breaking for comparison flows for unit testing)
1.  Update the ID representation to mostly match the pytest format of `path/to/test_file.py::test_function[arg#-arg#-arg#]` instead of 
2. Fallback to `repr` of object fo rexample inputs/outputs

Why?

- In unit testing, the example inputs/outputs are less important. Want to permit non-serializable objects.
- Don't want to keep creating populating new examples with new IDs just because the non-serializable object includes its memory ID in its repr

### DIFF
--- a/python/langsmith/_testing.py
+++ b/python/langsmith/_testing.py
@@ -5,14 +5,15 @@ import concurrent.futures
 import datetime
 import functools
 import inspect
-import json
 import logging
 import threading
 import uuid
 import warnings
+from collections import defaultdict
 from pathlib import Path
 from typing import Any, Callable, Optional, Sequence, Tuple, TypeVar, overload
 
+import orjson
 from typing_extensions import TypedDict
 
 from langsmith import client as ls_client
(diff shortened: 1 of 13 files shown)


python/langsmith/_testing.py
```diff
@@ -316,15 +317,28 @@ def _start_experiment(
         return client.read_project(project_name=experiment_name)
 
 
-def _get_id(func: Callable, inputs: dict, suite_id: uuid.UUID) -> uuid.UUID:
+# Track the number of times a parameter has been used in a test
+# This is to ensure that we can uniquely identify each test case
+# defined using pytest.mark.parametrize
+_param_dict: dict = defaultdict(lambda: defaultdict(int))
```
hinthornw, line 323: Global state not great, and being order based means if you insert an example inline the other examples will shift down to replace the others' ids, but:

1. Naming is cleaner and doens't require the parameters to be serializable.
2. If the example is updated, we get different versions so presumably this handles the "apples to apples" comparison issues

nfcampos (approved):

PR #614, opened 2024-04-18
Increase Read Timeout

hinthornw:


langchain-infra (commented):
Should we be exposing this as a separate param instead of Tuple?

hinthornw:
> Should we be exposing this as a separate param instead of Tuple?

Don't feel strongly here. Mainly copying the `requests` syntax

PR #621, opened 2024-04-23
JS[major]: Evaluate

bracesproul:
TODO:
- [ ] figure out how to apply metadata at top level instead of with context like py does. inside `_runEvaluators` and `_applySummaryEvaluators`


js/src/utils/_git.ts
bracesproul, line 1: I added sync and async versions.

Only using async atm but sync was added in case we need to add to an existing sync func without updating it to async

js/src/evaluation/evaluator.ts
```diff
@@ -14,35 +20,6 @@ export type Category = {
   label: string;
 };
 
-/**
- * Configuration for feedback.
- */
-export type FeedbackConfig = {
```
bracesproul, line 20: Duplicated type. Not exported from this file so okay to delete
cc @hinthornw

js/src/utils/_git.ts
hinthornw: Would it be possible to call the async one from a sync function or is that bad

js/src/client.ts
```diff
@@ -2430,4 +2443,49 @@ export class Client {
...
+  ): Promise<EvaluationResult[]> {
+    const results: Array<EvaluationResult> =
+      this._selectEvalResults(evaluatorResponse);
+    for (const res of results) {
+      let sourceInfo_ = sourceInfo || {};
+      if (res.evaluatorInfo) {
+        sourceInfo_ = { ...res.evaluatorInfo, ...sourceInfo_ };
+      }
+      let runId_: string | null = null;
+      if (res.targetRunId) {
```
hinthornw, line 2472: Do we need the ?.

js/src/evaluation/_runner.ts
```diff
@@ -0,0 +1,932 @@
...
+  /**
+   * The LangSmith client to use.
+   * @default undefined
+   */
+  client?: Client;
+  /**
+   * Whether to block until the evaluation is complete.
+   * @default true
+   */
+  blocking?: boolean;
```
hinthornw, line 99: This isn't used anywhere, right?

js/src/evaluation/_runner.ts
```diff
@@ -0,0 +1,932 @@
...
+  return _evaluate(target, options);
+}
+
+interface ExperimentResultRow {
+  run: Run;
+  example: Example;
+  evaluationResults: EvaluationResults;
+}
+
+class _ExperimentManagerMixin {
```
hinthornw, line 118: The mixin isn't really needed for JS since you don't have to split the sync and async experiment managers

js/src/evaluation/evaluator.ts
```diff
@@ -91,6 +71,146 @@ export type EvaluationResult = {
...
 export interface RunEvaluator {
-  evaluateRun(run: Run, example?: Example): Promise<EvaluationResult>;
+  evaluateRun(
+    run: Run,
+    example?: Example,
+    options?: Partial<RunTreeConfig>
+  ): Promise<EvaluationResult>;
+}
+
+export type DynamicRunEvaluatorParams = {
```
hinthornw, line 93: Unused i think

js/src/utils/_git.ts
```diff
@@ -0,0 +1,109 @@
+import { exec, execSync } from "child_process";
```
hinthornw, line 1: Do we need to check/handle anything to gracefully handle edge/browser/etc. environments?

js/src/evaluation/_runner.ts
```diff
@@ -0,0 +1,932 @@
...
+    this._evaluationResults = args.evaluationResults;
+    this._summaryResults = args.summaryResults;
+  }
+
+  get examples(): AsyncGenerator<Example> {
+    if (this._examples === undefined) {
+      if (!this._data) {
+        throw new Error("Data not provided in this experiment.");
+      }
+      return _resolveData(this._data, { client: this.client });
```
hinthornw, line 274: So we are fetching it but never adding to the examples Array? Why not consume and add to example array here rather than mixing and matching generators and the array?

js/src/evaluation/_runner.ts
```diff
@@ -0,0 +1,932 @@
...
+    evaluators: Array<EvaluatorT | RunEvaluator>,
+    options?: {
+      maxConcurrency?: number;
+    }
+  ): Promise<_ExperimentManager> {
+    const resolvedEvaluators = _resolveEvaluators(evaluators);
+    const experimentResults = this._score(resolvedEvaluators, options);
+
+    const results: AsyncGenerator<ExperimentResultRow>[] = [];
+    for await (const item of atee<ExperimentResultRow>(experimentResults)) {
```
hinthornw, line 367: Doesn't this consume the generator anyway? Don't think we need the `tee` operation then?

js/src/evaluation/_runner.ts
```diff
@@ -0,0 +1,932 @@
...
+        "Runs not provided in this experiment. Please predict first."
+      );
+    } else {
+      return this._runs;
+    }
+  }
+
+  async start(): Promise<_ExperimentManager> {
+    if (!this._examplesArray) {
+      this._examplesArray = [];
```
hinthornw, line 319: Maybe makes sense to just add this within the `examples` getter  as the cache so you don't have 2 ways of accessing it

js/src/evaluation/_runner.ts
```diff
@@ -0,0 +1,932 @@
...
+    metadata: {
+      ...metadata,
+      example_version: example.modified_at
+        ? new Date(example.modified_at).toISOString()
+        : new Date(example.created_at).toISOString(),
+    },
+    client,
+  };
+
+  const wrappedFn = wrapFunctionAndEnsureTraceable(fn, options) as ReturnType<
```
hinthornw, line 791: This will fail if the person passes in a traceable function right... since the `reference_example_id` won't be bound to the function? We need to guarantee that the run will be correctly traced, so I'd probably always wrap the function since we currently have no way of providing these values at invocation time

hinthornw (changes requested):

js/src/utils/_git.ts
bracesproul: We shouldn't --the `execGit` and `execSyncGit` functions both will catch errors and return null

js/src/utils/_uuid.ts
```diff
@@ -0,0 +1,5 @@
+export function isUUIDv4(uuid: string): boolean {
```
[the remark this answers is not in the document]
bracesproul: Nope! Was unaware this existed. Thanks

js/src/evaluation/_runner.ts
bracesproul: We return two generators here, which both use this original generator. We could get rid of it in favor of consuming `experimentResults` inside `withEvaluators` and returning the data instead of new generators.

js/src/evaluation/_runner.ts
bracesproul: `examples` getter has a return type of `AsyncGenerator<Example>` so if we want to add it there we either need to convert it to a generator, or add a union return type and add checks for the type everywhere it's used

js/src/evaluation/_runner.ts
bracesproul: Could do this, only reason I didn't is because we'd have to switch all the `this.examples` calls to `await this.getExamples()` since getters can't be async. I guess in the long run it's a little cleaner though, will fix.

js/src/evaluation/_runner.ts
bracesproul: Removed async generator examples getter, replaced with an async `getExamples` method

js/src/evaluation/_runner.ts
bracesproul: Fixed to always wrap

js/src/utils/_git.ts
hinthornw: We should probably just delete teh sync one if it isn't used

js/src/evaluation/_runner.ts
hinthornw: I'm less famliar with JS but looks like `atee` itself is sync right? Why do we need to do `for await (const...)`, etc. rather than just `const [r1, r2] = atee<ExperimentResultRow>(experimentResults)`

js/src/evaluation/_runner.ts
hinthornw: This still isn't used anywhere, right?

js/src/evaluation/_runner.ts
hinthornw: This still isn't needed, right? I'd rather not copy abstractions from python that were only necessary because of their asyncio nonsense

js/src/traceable.ts
```diff
@@ -165,6 +165,15 @@ export function traceable<Func extends (...args: any[]) => any>(
     return new Promise((resolve, reject) => {
       void asyncLocalStorage.run(currentRunTree, async () => {
         try {
+          const onEnd = config?.on_end;
+          if (onEnd) {
+            if (!currentRunTree) {
+              throw new Error(
+                "Can not call 'on_end' if currentRunTree is undefined"
```
hinthornw, line 172: hm - we shouldn't ever break someone's execution flow from tracing

js/src/utils/_git.ts
bracesproul: We can ignore anyone running in browser, since you'll need secrets etc which you can't expose in the browser anyway. I pushed a commit to change to a dynamic import though so it can still be used elsewhere and won't crash

js/src/evaluation/_runner.ts
bracesproul: Ah this is correct. The prev implementation I added was async which is why it was like that. Pushed up a fix

js/src/traceable.ts
bracesproul: Switched to a console.error

hinthornw:
Doesn't seem like we populate `revision-id` either in the experiment metadata by faliling back to the git information, if that's rpovided - would like to do that

hinthornw (approved):

PR #628, opened 2024-04-25
Allow for passing metadata into the wrapOpenAI method

jacoblee93:
CC @agola11 @hinthornw @nfcampos


jacoblee93:
I am pretty sure I did the peer deps correctly in this case since we don't have export tests to worry about, but would love a double check @nfcampos

Also, do we want to split into a `langsmith/wrappers/openai` entrypoint?

jacoblee93:
Split out into a separate entrypoint

hinthornw (approved):

PR #629, opened 2024-04-26
Allow aggregator for streamed chunks to be passed to traceable in TS

jacoblee93:
We'll automatically create and pass one of these for `wrapOpenAI`. Will do that in a third PR as I didn't want to create conflicts with #628.

@agola11 @hinthornw

### DIFF
--- a/js/src/traceable.ts
+++ b/js/src/traceable.ts
@@ -95,10 +95,14 @@ const isAsyncIterable = (x: unknown): x is AsyncIterable<unknown> =>
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export function traceable<Func extends (...args: any[]) => any>(
   wrappedFunc: Func,
-  config?: Partial<RunTreeConfig>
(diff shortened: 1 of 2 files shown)


hinthornw (approved):
Seems reasonable...

js/src/traceable.ts
```diff
@@ -157,7 +161,9 @@ export function traceable<Func extends (...args: any[]) => any>(
                 chunks.push(chunk);
                 yield chunk;
               }
-              await currentRunTree.end({ outputs: chunks });
+              const finalOutputs =
+                aggregator === undefined ? chunks : aggregator(chunks);
```
hinthornw, line 165: Should we try/catch the aggregator? Woudl probably not want to break the function execution just because someone mis-implemented the function

jacoblee93: Sure, down for that.

jacoblee93:
Updated

PR #631, opened 2024-04-26
Add expect.score

hinthornw:


python/langsmith/run_helpers.py
```diff
@@ -712,7 +712,7 @@ def trace(
     else:
         new_run = run_trees.RunTree(
             name=name,
-            run_id=run_id,
+            id=run_id,
```
hinthornw, line 715: bugfix :(

python/langsmith/_expect.py
```diff
@@ -336,6 +341,36 @@ def value(self, value: Any) -> _Matcher:
         """
         return _Matcher(self.client, "value", value, _executor=self.executor)
 
+    def score(
```
hinthornw, line 344: Mostly just an easier way to log stuff without having to fetch current run tree

python/langsmith/utils.py
```diff
@@ -488,3 +490,9 @@ def with_optional_cache(
             yield
     else:
         yield
+
+
+def _format_exc() -> str:
```
hinthornw, line 495: Clean up the stacktrace for errors in the UI

python/langsmith/_expect.py
```diff
@@ -72,7 +77,7 @@ def __init__(
             max_workers=3
         )
         rt = rh.get_current_run_tree()
-        self._run_id = rt.id if rt else run_id
+        self._run_id = rt.trace_id if rt else run_id
```
hinthornw, line 80: If you're running this within a nested function, we'd prefer it by default to be assigned to the root trace (since it's called beneath `@unit`

langchain-infra (approved):

PR #633, opened 2024-04-26
Adds streaming aggregation for OpenAI, fix metadata passing

jacoblee93:


jacoblee93:
@hinthornw

jacoblee93:
@agola11 need to test with tool calling

js/src/run_trees.ts
```diff
@@ -156,7 +163,7 @@ export class RunTree implements BaseRun {
     const rt = new RunTree({
       name: props?.name ?? "<lambda>",
       parent_run: parentRun,
-      tags: deduppedTags,
+      tags: dedupedTags,
```
hinthornw, line 166: nice :)

js/src/wrappers/openai.ts
```diff
@@ -60,6 +56,108 @@ type PatchedOpenAIClient<T extends OpenAIType> = {
...
+    }
+  }
+  const toolCalls: {
+    [
+      key: number
+    ]: Partial<OpenAI.Chat.Completions.ChatCompletionChunk.Choice.Delta.ToolCall>[];
+  } = {};
+  for (const c of choices) {
+    if (c.delta.content) {
+      message.content += c.delta.content;
```
hinthornw, line 80: Discussed offline, but may want to check this works even if content were a list (not unlikely some time in the future)

jacoblee93:
I think ready to go, final review?

js/src/wrappers/openai.ts
```diff
@@ -60,6 +56,108 @@ type PatchedOpenAIClient<T extends OpenAIType> = {
...
+      for (const chunk of toolCallChunks) {
+        if (chunk.function) {
+          if (!message.tool_calls[idx].function) {
+            message.tool_calls[idx].function = {
+              name: "",
+              arguments: "",
+            };
+          }
+          if (chunk.function.name) {
+            message.tool_calls[idx].function.name += chunk.function.name;
```
hinthornw, line 120: Shoudl this be overwrite instead of append?

hinthornw (approved):

PR #640, opened 2024-04-30
Fix traceable and wrapOpenAI

nfcampos:


hinthornw (approved):

js/src/run_trees.ts
```diff
@@ -192,7 +192,7 @@ export class RunTree implements BaseRun {
     };
   }
 
-  public async createChild(config: RunTreeConfig): Promise<RunTree> {
```
bracesproul, line 195: This is breaking. Need to revert this

PR #643, opened 2024-04-30
traceable: Add option to configure which args are sent as inputs to langsmith

nfcampos:


js/src/tests/traceable.int.test.ts
```diff
@@ -135,7 +135,7 @@ test.concurrent("Test traceable wrapper with async error thrown", async () => {
 
   expect(collectedRun).not.toBeNull();
   expect(collectedRun!.error).toEqual("Error: I am bad");
-  expect(collectedRun!.inputs).toEqual({ args: ["testing", 9] });
+  expect(collectedRun!.inputs).toEqual({ 0: "testing", 1: 9 });
```
dqbd, line 138: _nit_: this is likely a breaking change for `runOnDataset` in LangChain

hinthornw:
Handledwith processInputs

PR #656, opened 2024-05-03
fix(js): serialization of run tree if found in evaluate as input

dqbd:


js/src/run_trees.ts
```diff
@@ -253,27 +257,28 @@ export class RunTree implements BaseRun {
     this.end_time = this.end_time ?? endTime;
   }
 
-  private async _convertToCreate(
+  private _convertToCreate(
```
hinthornw, line 260: nice

hinthornw (approved):

PR #658, opened 2024-05-03
feat(js): add hasDataset function

dqbd:


hinthornw (approved):
Haha we should have a `peek` endpoint for this

PR #659, opened 2024-05-06
Limit batch size

hinthornw:


python/langsmith/client.py
```diff
@@ -2159,24 +2159,47 @@ def get_test_results(
...
+            examples = self.list_examples(example_ids=batch)
+            return [
                 {
+                    "example_id": example.id,
+                    **{f"reference.{k}": v for k, v in (example.outputs or {}).items()},
+                }
+                for example in examples
+            ]
+
+        batch_size = 50
```
hinthornw, line 2194: Main thing is this - bump down to 50 from 100 (was causing errors)

python/pyproject.toml
```diff
@@ -53,6 +53,7 @@ vcrpy = "^6.0.1"
 fastapi = "^0.110.1"
 uvicorn = "^0.29.0"
 pytest-rerunfailures = "^14.0"
+rapidfuzz = "^3.9.0"
```
nfcampos, line 56: this here on purpose?

hinthornw: was for linting warnings locally

hinthornw: just a dev dep but i can remove

langchain-infra (approved):

PR #662, opened 2024-05-06
feat(js): trace deferred values in input

dqbd:
- **Send inputs even when patching run**
- **Collect deferred values during invocation via proxy**
- **Handle ReadableStream in input**

### DIFF
--- a/js/src/traceable.ts
+++ b/js/src/traceable.ts
@@ -116,6 +116,12 @@ const isGenerator = (x: unknown): x is Generator =>
   // eslint-disable-next-line no-instanceof/no-instanceof
   x != null && typeof x === "function" && x instanceof GeneratorFunction;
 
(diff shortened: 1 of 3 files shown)


js/src/traceable.ts
```diff
@@ -167,6 +180,184 @@ const getTracingRunTree = (
...
+        return (
+          resolve: (value: unknown) => unknown,
+          reject?: (error: unknown) => unknown
+        ) => {
+          return boundThen(
+            (value) => {
+              proxyState.current = ["resolve", value];
+              return resolve(value);
+            },
+            (error) => {
```
nfcampos, line 203: Screenshot 2024-05-06 at 3 02 08 PM

I think this results in swallowing promise rejections when users call `.then` without 2nd arg

PR #663, opened 2024-05-07
fix(js): throw when wrapping same client twice, allow passing name in argsConfigExtra

dqbd:
- **Make sure we're only allowing wrapping once**
- **Allow passing name to runnable config like**

### DIFF
--- a/js/src/run_trees.ts
+++ b/js/src/run_trees.ts
@@ -62,11 +62,6 @@ export interface RunTreeConfig {
 
 export interface RunnableConfigLike {
   /**
-   * Name of the traced run, overriding the name provided by the traceable
-   * function.
-   */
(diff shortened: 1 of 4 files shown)


js/src/run_trees.ts
```diff
@@ -61,6 +61,11 @@ export interface RunTreeConfig {
 }
 
 export interface RunnableConfigLike {
+  /**
+   * Name of the traced run, overriding the name provided by the traceable
+   * function.
+   */
+  name?: string;
```
hinthornw, line 68: Haha the original intent of this was for the langchain -> langsmith handoff. I'm guessing the scope has crept a bit?

hinthornw (approved):

PR #668, opened 2024-05-07
js[patch]: Fix listRuns limit arg

bracesproul:
Added faker package for tests


hinthornw (approved):

js/src/client.ts
```diff
@@ -1194,11 +1194,25 @@ export class Client {
...
       "/runs/query",
       body
     )) {
-      yield* runs;
+      if (limit) {
+        if (runsYielded >= limit) {
+          break;
+        }
+        if (runs.length + runsYielded > limit) {
+          const newRuns = runs.splice(limit - runsYielded);
```
dqbd, line 1207: Shouldn't it be 

```suggestion
          const newRuns = runs.slice(0, limit - runsYielded)
```

dqbd: Or alternatively

```suggestion
          runs.splice(limit - runsYielded);
          yield* runs
```

bracesproul:
@hinthornw ci throwing 401 unauthorized errors. Did something happen to the LS secrets in gh actions?

hinthornw:
Nothing was changed in 3 months... Wonder if GH is acting up. I went ahead and updated the key just in case

PR #672, opened 2024-05-08
feat(pairwise evals): add option to randomize order of experiments; print URL for pairwise experiment

samnoyes:


python/langsmith/evaluation/_runner.py
```diff
@@ -720,6 +726,25 @@ def __getitem__(self, key):
...
+    experiments: Tuple[schemas.TracerSession, schemas.TracerSession],
+    comparative_experiment: schemas.ComparativeExperiment,
+) -> None:
+    url = experiments[0].url or experiments[1].url
+    if url:
+        project_url = url.split("?")[0]
+        dataset_id = comparative_experiment.reference_dataset_id
+        base_url = project_url.split("/projects/p/")[0]
+        comparison_url = (
+            f"{base_url}/datasets/{dataset_id}/compare?"
```
hinthornw, line 739: Nice - at some point may be nice to factor this uot

hinthornw: Does this work if there's > 2 experiments? Is it more future proof if we just added a list of all the IDs?

hinthornw (approved):

python/langsmith/evaluation/_runner.py
samnoyes: good point, will do that

PR #678, opened 2024-05-09
fix(js): LangChain / traceable handoff

### LINKED ISSUE #598: JS SDK traceable is not compatible with langchain JS chain.invoke tracing
### Feature request

Any chains invoked inside a function wrapped with `traceable` have separate trace ids (or vice versa, with chains calling `traceable` functions internally).

Ideally the trace id gets propagated automatically.

Related to https://github.com/langchain-ai/langchainjs/issues/5083

### Motivation

Increase compatibility with Langchain JS native tracing.

### Example

image

```ts
import { traceable } from 'langsmith/traceable';
import { RunnableLambda } from '@langchain/core/runnables';

const exampleRunnable = RunnableLambda.from(async (foo: string) =>
  foo.toUpperCase(),
);

exampleRunnable.name = 'exampleRunnable';

const exampleTraceable = traceable(
  async (foo: string) => {
    const result = await exampleRunnable.invoke(foo);

    console.log(result);

    return result;
  },
  {
    name: 'exampleTraceable',
  },
);

exampleTraceable('bar');
```

dqbd:
- Traceable to LangChain: add the entire tree of RunTrees into the BaseTracer `runMap`, assign the `_parentRunId` to the callback manager
- LangChain to Traceable: the `RunnableTraceable` invokes the function without calling `*WithConfig`, avoiding unwanted callback invocation

**Verifying package manager behaviour**

Exposed the LangChain utilities + added `@langchain/core` as an optional dependency to allow supporting older versions of LangChain. 

- [x] Yarn 1.x: does not install the `@langchain/core`, not included in package-lock.json
- [X] PNPM: does not install `@langchain/core`, included in package-lock.json
- [X] NPM: does not install `@langchain/core`, included in package-lock.json
- [X] Bun: does not install `@langchain/core`, included in package-lock.json
- [X] Deno: does not install `@langchain/core`

Upstream PR which implements similar functionality can be seen in https://github.com/langchain-ai/langchainjs/pull/5339

Closes #598


js/scripts/create-entrypoints.js
```diff
@@ -12,8 +12,10 @@ const entrypoints = {
   traceable: "traceable",
   evaluation: "evaluation/index",
   schemas: "schemas",
+  langchain: "langchain",
```
jacoblee93, line 15: We have `langchain/smith`, funny that we'll now have `langsmith/langchain` 😂

js/src/langchain.ts
```diff
@@ -0,0 +1,122 @@
...
+    rootVisited.add(rootRun.id);
+    rootRun = rootRun.parent_run;
+  }
+
+  const queue = [rootRun];
+  const visited = new Set<string>();
+
+  while (queue.length > 0) {
+    const current = queue.shift();
+    if (!current || visited.has(current.id)) continue;
```
jacoblee93, line 60: When would we need `visited` if we're only pushing child runs?

js/src/langchain.ts
```diff
@@ -0,0 +1,122 @@
...
+    visited.add(current.id);
+
+    runMap.set(current.id, current);
+    if (current.child_runs) {
+      queue.push(...current.child_runs);
+    }
+  }
+
+  if (callbacks != null) {
+    Object.assign(callbacks, { _parentRunId: runTree.id });
```
jacoblee93, line 70: This is to get around the fact that it's a private property?

Can we add a flag to add a setter method?

js/src/langchain.ts
```diff
@@ -0,0 +1,122 @@
...
+      throw new Error(
+        "RunnableTraceable requires a function that is wrapped in traceable higher-order function"
+      );
+    }
+
+    this.func = fields.func;
+  }
+
+  async invoke(input: RunInput, options?: Partial<RunnableConfig>) {
+    const [config] = this._getOptionsList(options ?? {}, 1);
```
jacoblee93, line 115: Not `mergeConfig`?

js/src/langchain.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type AnyTraceableFunction = TraceableFunction<(...any: any[]) => any>;
+
+/**
+ * RunnableTraceable is a Runnable that wraps a traceable function.
+ * This allows adding Langsmith traced functions into LangChain sequences.
+ */
+export class RunnableTraceable<RunInput, RunOutput> extends Runnable<
```
jacoblee93, line 92: What about streaming? Should this extend `RunnableBinding` instead?

js/src/run_trees.ts
```diff
@@ -157,44 +158,55 @@ export class RunTree implements BaseRun {
...
-      name: props?.name ?? "<lambda>",
-      parent_run: parentRun,
-      tags: dedupedTags,
+
+    const parentRunTree = new RunTree({
+      name: parentRun?.name ?? "<parent>",
+      id: parentRun?.id,
+      client,
+      // TODO: handle tracing enabled
+      tracingEnabled: true,
```
jacoblee93, line 192: Pull from env var?

js/src/singletons/traceable.ts
```diff
@@ -0,0 +1,128 @@
+import { RunTree, RunnableConfigLike } from "../run_trees.js";
+
+interface AsyncStorageLike {
+  getStore: () => RunTree | undefined;
+
+  run: (context: RunTree | undefined, fn: () => void) => void;
+}
+
+export const TraceableLocalStorageContext = (() => {
```
jacoblee93, line 9: Would prefer this to match the interface in `@langchain/core/singletons`

js/src/singletons/traceable.ts
```diff
@@ -0,0 +1,128 @@
...
+    );
+  }
+
+  return runTree;
+};
+
+export const ROOT = Symbol.for("langsmith:traceable:root");
+
+type SmartPromise<T> = T extends AsyncGenerator
+  ? T
```
jacoblee93, line 51: nit: Hard to read, can you add parenthesis?

js/src/singletons/traceable.ts
```diff
@@ -0,0 +1,128 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type UnionToIntersection<U> = (U extends any ? (x: U) => void : never) extends (
+  x: infer I
+) => void
+  ? I
+  : never;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type TraceableFunction<Func extends (...args: any[]) => any> =
```
jacoblee93, line 84: Move under a dedicated `types.ts`?

js/src/langchain.ts
```diff
@@ -0,0 +1,122 @@
+import { CallbackManager } from "@langchain/core/callbacks/manager";
+import { LangChainTracer } from "@langchain/core/tracers/tracer_langchain";
+import { Runnable, RunnableConfig } from "@langchain/core/runnables";
```
jacoblee93, line 3: Do we not need to import/override the storage in `@langchain/core/singletons`? Or does it not matter due to all the run tree overriding below?

jacoblee93 (commented):
I think looks solid! A few questions/comments

js/src/langchain.ts
dqbd: I don't think so, but RunTree's can be constructed by users themselves, so I'd rather be careful instead.

js/src/langchain.ts
dqbd: In the https://github.com/langchain-ai/langchainjs/pull/5339 PR we already made this writable (while still keeping it private), this is needed for backwards compatibility

js/src/langchain.ts
dqbd: I don't think it matters, given that:
1. The handoff happens at the boundary of `@traceable` and LangChain object
2. Provided config always has precedence over ALS-one, thus the entrypoint has the mutated callback manager
3. Invocation of `AsyncLocalStorage.run()` happens when invoking RunnableLambda, which by then already has the correct callback manager to create a child run manager

js/src/langchain.ts
dqbd: By default streaming and batching should work fine, as the base implementation of `stream`, `batch` etc. will call `invoke`. But you're right that we might need to override `transform` as well to avoid creating an extra trace in LangSmith

js/src/singletons/traceable.ts
dqbd: Tried to do so, but seems like Prettier doesn't allow it

js/src/langchain.ts
jacoblee93: Oh, right. I think we may want to support streaming via a returned generator, but we can do that in a followup if you add a note?

js/package.json
```diff
@@ -109,11 +117,15 @@
     "typescript": "^5.4.5"
   },
   "peerDependencies": {
-    "openai": "*"
+    "openai": "*",
+    "@langchain/core": "*"
```
jacoblee93, line 121: Lower bound this to `0.1` maybe?

js/src/env.ts
```diff
@@ -0,0 +1,16 @@
...
+  if (tracingEnabled !== undefined) {
+    return tracingEnabled;
+  }
+  const envVars = [
+    "LANGSMITH_TRACING_V2",
+    "LANGCHAIN_TRACING_V2",
+    "LANGSMITH_TRACING",
+    "LANGCHAIN_TRACING",
+  ];
+  return Boolean(
```
jacoblee93, line 13: nit: I'm not sure what this function does, I usually prefer `!!`

js/src/tests/traceable_langchain.test.ts
```diff
@@ -0,0 +1,275 @@
...
+  const parser = new StringOutputParser();
+  const chain = prompt.pipe(llm).pipe(parser);
+
+  test("invoke", async () => {
+    const { client, callSpy } = mockClient();
+
+    const main = traceable(
+      async (input: { text: string }) => {
+        return chain.invoke(input, {
+          callbacks: await getLangchainCallbacks(),
```
jacoblee93, line 26: Apologies if I'm still not getting it - there's no way to do this automatically?

I thought part of the idea was to pull in the local storage instance from `@langchain/core/singletons` so that this would happen seamlessly?

js/src/langchain.ts
dqbd: Implemented here: https://github.com/langchain-ai/langsmith-sdk/pull/708

js/src/env.ts
dqbd: Should be the same, replaced

js/src/tests/traceable_langchain.test.ts
dqbd: Resolved in Slack

jacoblee93 (approved):

PR #681, opened 2024-05-10
feat(pairwise): allow passing evaluator directly to wait for evaluation to complete

dqbd:
Allows the following syntax
```typescript
const pairwise = await evaluateComparative(
  [
    evaluate((input) => ({ foo: `first:${input.input}` }), {
      data: TESTING_DATASET_NAME,
    }),
    evaluate((input) => ({ foo: `second:${input.input}` }), {
      data: TESTING_DATASET_NAME,
    }),
  ],
  {
    evaluators: [
      (runs) => ({
        key: "latter_precedence",
        scores: Object.fromEntries(runs.map((run, i) => [run.id, i % 2])),
      }),
    ],
  }
);
```


js/src/evaluation/evaluate_comparative.ts
```diff
@@ -125,10 +141,34 @@ export async function evaluateComparative(
...
+    }
+
+    // if we know the number of runs beforehand, check if the
+    // number of runs in the project matches the expected number of runs
+    return Promise.all(
+      resolvedExperiments.map((experiment) =>
+        pRetry(
+          async () => {
+            const project = await loadExperiment(client, experiment);
+            if (project.run_count !== experiment?.results.length) {
```
hinthornw, line 162: What if the experiment was deliberately only run over a subset of the runs?

dqbd: That should be fine, as we are only comparing the local experiment state with the server state

hinthornw (approved):

PR #682, opened 2024-05-10
feat(pairwise): trace evaluators in evaluateComparitive

dqbd:


hinthornw (approved):

js/src/evaluation/evaluate_comparative.ts
```diff
@@ -332,12 +333,37 @@ export async function evaluateComparative(
...
+        evaluatorRun.inputs = { runs: runs, example: example };
+        evaluatorRun.outputs = result;
+
+        return {
+          ...result,
+          source_run_id: result.source_run_id ?? evaluatorRun.id,
+        };
+      },
+      {
+        project_name: "evaluators",
```
hinthornw, line 355: Do we need to do anything if they have their tracing vars turned off

dqbd: that should be already dealt by `tracingIsEnabled` inside `traceable`

PR #702, opened 2024-05-15
Add splits as param to list_examples, create_example/s, update_example (and typescript equivs)

samnoyes:


python/langsmith/client.py
```diff
@@ -3112,6 +3113,8 @@ def list_examples(
                 timestamp to retrieve the examples as of.
                 Response examples will only be those that were present at the time
                 of the tagged (or timestamped) version.
+            splits (List[str], optional): A list of dataset splits to filter by.
+                Returns examples only from the specified splits.
```
hinthornw, line 3117: Could you say what a split is here

hinthornw (approved):

PR #703, opened 2024-05-15
chore[client]: add .info() deprecation notice

dqbd:


hinthornw:
oh didn't realize this

PR #708, opened 2024-05-17
fix(handoff): support streaming returned generators

dqbd:


agola11 (approved):

js/src/langchain.ts
```diff
@@ -113,7 +119,40 @@ export class RunnableTraceable<RunInput, RunOutput> extends Runnable<
...
+      const iterator = result[Symbol.asyncIterator]();
+      while (true) {
+        const { done, value } = await iterator.next();
+        if (done) break;
+        yield value as RunOutput;
+      }
+      return;
+    }
+
+    if (isIteratorLike(result)) {
```
jacoblee93, line 146: What's the use case for this check? Seems unnecessary?

js/src/langchain.ts
```diff
@@ -113,7 +119,40 @@ export class RunnableTraceable<RunInput, RunOutput> extends Runnable<
...
+
+  async *_streamIterator(
+    input: RunInput,
+    options?: Partial<RunnableConfig>
+  ): AsyncGenerator<RunOutput> {
+    const result = await this.invoke(input, options);
+
+    if (isAsyncIterable(result)) {
+      const iterator = result[Symbol.asyncIterator]();
+      while (true) {
```
jacoblee93, line 138: nit: `for await... const` syntax?

js/src/langchain.ts
dqbd: If the traceable returns a generator, the stream wouldn't caught that

PR #716, opened 2024-05-20
chore(js): reexport TraceableFunction type from singleton

dqbd:
As `langsmith/singletons/traceable` is an entrypoint from langchain as well, we need to reexport the type


jacoblee93 (approved):
Seems fine to me!

PR #719, opened 2024-05-21
feat(js): move `getLangchainCallbacks` to TypeScript SDK

dqbd:
Moves the inline code snippet from langsmith new experiment pane to library instead. New entrypoint to make sure users don't need to depend on full-blown `langchain` when they can only depend on `@langchain/core`


agola11 (approved):

jacoblee93 (approved):
A little bit odd but long term we'll move all the evaluator code into langsmith right?

PR #724, opened 2024-05-22
implement repetitions in js and python

samnoyes:


python/langsmith/evaluation/_runner.py
```diff
@@ -82,6 +82,7 @@ def evaluate(
     experiment_prefix: Optional[str] = None,
     description: Optional[str] = None,
     max_concurrency: Optional[int] = None,
+    num_repetitions: Optional[int] = 1,
```
hinthornw, line 85: What's the difference between none and 1?

hinthornw: Seems like it shouldn't be nullable?

samnoyes: ya fair, good point

samnoyes: done

python/langsmith/evaluation/_arunner.py
```diff
@@ -81,6 +83,8 @@ async def aevaluate(
         description (Optional[str]): A description of the experiment.
         max_concurrency (Optional[int]): The maximum number of concurrent
             evaluations to run. Defaults to None.
+        num_repetitions (int): The number of times to repeat the evaluation.
```
hinthornw, line 86: ha sorry - maybe "the number of times to run the evaluation" since "to repeat" makes 1 sound like we would run twice

dqbd (approved):

js/src/evaluation/_runner.ts
```diff
@@ -183,7 +192,15 @@ class _ExperimentManager {
       for await (const example of unresolvedData) {
         exs.push(example);
       }
-      this.setExamples(exs);
+      if (this._numRepetitions) {
```
dqbd, line 195: _nit_: should we check for non-positive numbers?

PR #725, opened 2024-05-22
Add method to create multiple presigned feedback urls in one call

nfcampos:


python/langsmith/client.py
```diff
@@ -3808,6 +3808,77 @@ def create_presigned_feedback_token(
...
+        body: List[Dict[str, Any]] = [
+            {
+                "run_id": run_id,
+                "feedback_key": feedback_key,
+                "feedback_config": feedback_config,
+            }
+            for feedback_key, feedback_config in zip(feedback_keys, feedback_configs)
+        ]
+        for part in body:
+            if expiration is None:
```
hinthornw, line 3857: ooc, why not make this once and re-use in the loop?

hinthornw (approved):

python/langsmith/client.py
nfcampos: because I wasn't thinking

nfcampos: Updated

PR #746, opened 2024-05-29
Read parent run info from RunnableConfig if passed to function decorated with traceable

nfcampos:
- pass config into wrapped func only if signature declares it
- modify signature of wrapper func to declare config as a kw arg, otherwise runnables don't pass it in
- extract client and project_name from tracer, in addition to parent run info
- update as_runnable to delegate run tree creation to new method


agola11 (approved):

python/tests/unit_tests/test_run_helpers.py
```diff
@@ -322,6 +322,85 @@ async def my_function(a, b, d):
         assert result == [6, 7]
 
 
+def test_traceable_parent_from_runnable_config() -> None:
+    try:
+        from langchain.callbacks.tracers import LangChainTracer
+        from langchain.schema.runnable import RunnableLambda
+    except ImportError:
+        pytest.skip("Skipping test that requires langchain")
+    with tracing_context(enabled=True):
```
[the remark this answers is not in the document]
nfcampos: Ah not needed, was for the initial version of the test, but later changed

python/tests/unit_tests/test_run_helpers.py
```diff
@@ -322,6 +322,85 @@ async def my_function(a, b, d):
...
+
+        my_function_runnable = RunnableLambda(my_function)
+
+        assert (
+            my_function_runnable.invoke(
+                1, {"callbacks": [LangChainTracer(client=mock_client_)]}
+            )
+            == 2
+        )
+        time.sleep(1)
```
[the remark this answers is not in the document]
nfcampos: Yes

python/langsmith/run_helpers.py
```diff
@@ -645,6 +638,17 @@ def generator_wrapper(
         else:
             selected_wrapper = wrapper
         setattr(selected_wrapper, "__langsmith_traceable__", True)
+        sig = inspect.signature(selected_wrapper)
+        if not sig.parameters.get("config"):
+            sig = sig.replace(
+                parameters=[
+                    *sig.parameters.values(),
+                    inspect.Parameter(
+                        "config", inspect.Parameter.KEYWORD_ONLY, default=None
```
hinthornw, line 647: god bless Guido van Rossum heh - clever

python/langsmith/run_helpers.py
```diff
@@ -970,14 +938,18 @@ def _collect_extra(extra_outer: dict, langsmith_extra: LangSmithExtra) -> dict:
...
+    langsmith_extra: LangSmithExtra, config: Optional[dict] = None
+) -> Optional[run_trees.RunTree]:
     parent = langsmith_extra.get("parent")
     if isinstance(parent, run_trees.RunTree):
         return parent
     if isinstance(parent, dict):
         return run_trees.RunTree.from_headers(parent)
     if isinstance(parent, str):
         return run_trees.RunTree.from_dotted_order(parent)
+    if rt := run_trees.RunTree.from_runnable_config(config):
```
hinthornw, line 951: Wouldn't this raise an import error if the user doesn't have langchain installed? We should probably:

1. catch that error
2. maybe cache whether or not it's installed to avoid repeatedly trying to import

hinthornw: I think this should also be lower precedence than the "run-tree" condition in 953

hinthornw (changes requested):
We need to ensure traceable still works without langchain installed :) - main thing is catching the import error

hinthornw (approved):

PR #750, opened 2024-05-31
fix sig.replace to account for kwargs

mc-marcocheng:
If the `selected_wrapper` function contains a `**kwargs` parameters, `sig.replace(...)` breaks because of the following error:
```
ValueError: wrong parameter order: variadic keyword parameter before keyword-only parameter
```

This PR fixes the above error.


hinthornw:
Thank you! I'll try to look into this and add unit tests for it

PR #772, opened 2024-06-07
feat(js): add `withRunTree` method for updating context, improve types for `toHeader` and `fromHeaders`

dqbd:


agola11 (approved):

agola11:
Before merging, please add a `:::note` to indicate which version of the SDK these methods will be available in

PR #788, opened 2024-06-14
feat(python, js): add anonymizer API

dqbd:
Expected TypeScript usage:
```ts
import { createAnonymizer } from "langsmith/anonymizer"

// provide list of regex and replacement values
const anonymizer = createAnonymizer([
  { pattern: "...", replace: "[value]" }
])

const client = new Client({ anonymizer })

// replace string values
const anonymizer = createAnonymizer((value) => value.replace("...", "[value]"))
```

Expected Python usage:
```python
from langsmith.anonymizer import create_anonymizer

# provide list of regex and replacement values
anonymizer = create_anonymizer([{"pattern": r"...", "replace": "[value]"}])

client = Client(anonymizer=anonymizer)

# replace string values
anonymizer = create_anonymizer(lambda text: r"...".sub("[value]", text))
```

### DIFF
--- a/python/langsmith/anonymizer.py
+++ b/python/langsmith/anonymizer.py
@@ -7,7 +7,7 @@ from typing import Any, Callable, List, Optional, Tuple, TypedDict, Union
 
 
 class _ExtractOptions(TypedDict):
-    maxDepth: Optional[int]
+    max_depth: Optional[int]
     """
     Maximum depth to traverse to to extract string nodes
     """
@@ -26,7 +26,7 @@ class StringNode(TypedDict):
 def _extract_string_nodes(data: Any, options: _ExtractOptions) -> List[StringNode]:
     max_depth = options.get("maxDepth") or 10
 
-    queue: List[Tuple[Any, int, List[Union[str, int]]]] = [(data, 0, list())]
+    queue: List[Tuple[Any, int, List[Union[str, int]]]] = [(data, 0, [])]
     result: List[StringNode] = []
 
     while queue:
@@ -35,7 +35,7 @@ def _extract_string_nodes(data: Any, options: _ExtractOptions) -> List[StringNod
             continue
         value, depth, path = task
 
-        if isinstance(value, dict) or isinstance(value, defaultdict):
+        if isinstance(value, (dict, defaultdict)):
             if depth >= max_depth:
                 continue
             for key, nested_value in value.items():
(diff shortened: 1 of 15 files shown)


python/langsmith/client.py
```diff
@@ -461,6 +461,7 @@ def __init__(
         web_url: Optional[str] = None,
         session: Optional[requests.Session] = None,
         auto_batch_tracing: bool = True,
+        anonymizer: Optional[Callable[[dict], dict]] = None,
```
hinthornw, line 464: If we're implementing a full anonymizer API, it may make sense to have it take in a `Run` object I think, or at least accept other values like name, metadata, etc.

Reason being - some people may want to have different behavior depending on the span, and we already have the single `hide` funciton that only accepts a dict

hinthornw (commented):
Will test more - had some comments. I think main thing is that I *think* this wouldn't step through custom objects / pydantic objects / etc. , so it wouldn't apply to chat model output, for instance, which feels a bit limiting

python/langsmith/anonymizer.py
```diff
@@ -0,0 +1,182 @@
...
+    if isinstance(replacer, list):
+        return RuleNodeProcessor(rules=replacer)
+    elif callable(replacer):
+        return CallableNodeProcessor(func=replacer)
+    else:
+        return replacer
+
+
+def create_anonymizer(
+    replacer: ReplacerType, options: Optional[ReplacerOptions] = None
```
hinthornw, line 155: it seems more pythonic to have the options as separate keyword args (would also let the user avoid having to decide "deep_clone")

python/langsmith/anonymizer.py
```diff
@@ -0,0 +1,182 @@
...
+) -> Callable[[Any], Any]:
+    """Create an anonymizer function."""
+    processor = _get_node_processor(replacer)
+
+    def anonymizer(data: Any) -> Any:
+        nodes = _extract_string_nodes(
+            data, {"max_depth": (options.get("max_depth") if options else None) or 10}
+        )
+        mutate_value = (
```
hinthornw, line 164: Do we need to deep clone if we're already doing json serialization before this?

python/langsmith/client.py
```diff
@@ -1241,17 +1247,23 @@ def _create_run(self, run_create: dict):
...
+            return inputs
         return self._hide_inputs(inputs)
 
     def _hide_run_outputs(self, outputs: dict):
-        if self._hide_outputs is False:
-            return outputs
         if self._hide_outputs is True:
             return {}
+        if self._anonymizer:
+            json_outputs = orjson.loads(_dumps_json(outputs))
```
hinthornw, line 1263: we're doing a lot of copying and re-loading and dumping at this point... If a user says they want to deepcopy in the anonymizer, we

1. Deep(ish) copy the object to avoid mutations in the queue
2. copy/serialize and re-load here 
3. Maybe copy within the anonymizer?
4. and finally serialize at the end

hinthornw (approved):

python/langsmith/client.py
dqbd: Will address this in a follow-up PR!

PR #792, opened 2024-06-14
unit tests: test helpers

eyurtsev:


python/tests/combined_tracing/utils.py
```diff
@@ -0,0 +1,164 @@
...
+    def get_children(self, node: Span) -> List[Span]:
+        """Return all nodes that have the given node as a parent."""
+        child_ids = self.parent_id_to_child_ids.get(node["id"], [])
+        return [child for child in self.children if child["id"] in child_ids]
+
+    def get_breadth_first_travel(self) -> List[Tuple[Span, int]]:
+        """Return a list of nodes in the order they were added."""
+        # First we assert there's only a single root node
+        root_nodes = self.get_root_nodes()
+        assert len(root_nodes) == 1
```
hinthornw, line 140: Maybe we let this be configurable?

eyurtsev: +1 we'll need it for independent traces, will refactor, wanted to iterate on existing tests quickly to see what i discover

python/tests/combined_tracing/utils.py
```diff
@@ -0,0 +1,164 @@
...
+
+    return calls
+
+
+def _extract_lineage_information(requests: Sequence[Request]):
+    lineages: List[Span] = []
+    for request in requests:
+        if request["verb"] != "POST":
+            continue
+        if request["path"].endswith("/runs/batch)"):
```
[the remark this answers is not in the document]
eyurtsev: nope it's wrong, but looks like it was fine b/c it was double wrong and it was the only POST example

hinthornw (approved):

PR #793, opened 2024-06-14
js[patch]: Store local storage instance on globalThis

jacoblee93:
Matches LangChain.js, lays groundwork for easier trace handoffs since both can now reference the same `AsyncLocalStorage` instance.

https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/singletons/index.ts#L24


jacoblee93:
Wait, no we don't want to do this yet, sorry. Needs to be after a potential new JS tracer that uses a RunTree

PR #796, opened 2024-06-17
[Python] Pass through IO

hinthornw:
When loading from runnable config

### DIFF
--- a/.github/workflows/python_test.yml
+++ b/.github/workflows/python_test.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Install dependencies
         run: |
           poetry install --with dev,lint
-          poetry run pip install -U langchain
+          poetry run pip install -U langchain langchain-core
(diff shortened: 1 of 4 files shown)


.github/workflows/python_test.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       - name: Install dependencies
         run: |
           poetry install --with dev,lint
-          poetry run pip install -U langchain
+          poetry run pip install -U langchain langchain-core
```
eyurtsev, line 45: Are there unit tests that are internal only without any extras from langchain/langchian-core?

eyurtsev (approved):

.github/workflows/python_test.yml
hinthornw: Most of the unit tests are that way - only a handful use langchain code

PR #799, opened 2024-06-18
Add pytest-sockets to test dependencies

eyurtsev:
- Add pytest-sockets
- Run unit tests with --disable-socket and --allow-unix-socket


hinthornw (approved):

python/pyproject.toml
```diff
@@ -57,6 +57,10 @@ pytest-rerunfailures = "^14.0"
 [tool.poetry.group.lint.dependencies]
 openai = "^1.10"
 
+
+[tool.poetry.group.test.dependencies]
+pytest-socket = "^0.7.0"
```
hinthornw, line 62: probably fine to just put in `dev` dependencies

PR #827, opened 2024-06-27
Update list examples

hinthornw:


python/langsmith/client.py
```diff
@@ -3203,6 +3203,8 @@ def list_examples(
         as_of: Optional[Union[datetime.datetime, str]] = None,
         splits: Optional[Sequence[str]] = None,
         inline_s3_urls: bool = True,
+        *,
```
hinthornw, line 3206: Could move this further up... realized we want these keyword only

agola11 (approved):

PR #831, opened 2024-07-01
[RFC] Add `LLMEvaluator` to create LLM-as-a-judge evaluators

agola11:
It's currently quite annoying to use LLM-as-a-judge evaluators in code, and there is a bit of a disconnect between the SDK and the UI. 

Our [off-the-shelf evaluators](https://docs.smith.langchain.com/how_to_guides/evaluation/use_langchain_off_the_shelf_evaluators) don't even use tool calling

In the UI, you can specify the prompt and output schema. With LangSmith, you have to use `.with_structured_output` within a custom function. This can be a lot of boilerplate for the user.

Additionally, a JSON schema or generic pydantic model is likely not the best interface for allowing people to specify the score format for their LLM evaluators. Opted for something more opinionated, `ContinuousScoreConfig` and `CategoricalScoreConfig`

**Important detail**: I map each score to a tool as opposed to each argument of a tool. This allows other attributes, like explanation to be extracted and mapped to the same feedback entry.

Future work:
* Allow people to load these from a file
* Create off-the-shelf evaluators based off of `LLMEvaluator`
* async


python/langsmith/evaluation/llm_evaluator.py
```diff
@@ -0,0 +1,221 @@
...
+        except ImportError as e:
+            raise ImportError(
+                "LLMEvaluator requires langchain-core to be installed. "
+                "Please install langchain-core by running `pip install langchain-core`."
+            ) from e
+        try:
+            from langchain.chat_models import init_chat_model
+        except ImportError as e:
+            raise ImportError(
+                "LLMEvaluator requires langchain to be installed. "
```
hinthornw, line 110: Maybe do this one first since that would give us core for free? Reduce the number of times people install these deps one by one

python/langsmith/evaluation/llm_evaluator.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def evaluate_run(
+        self, run: Run, example: Optional[Example] = None
+    ) -> Union[EvaluationResult, EvaluationResults]:
+        """Evaluate a run."""
+        if self.map_variables:
+            variables = self.map_variables(run, example)
+            if set(self.prompt.input_variables) - set(variables.keys()):
+                raise ValueError(
+                    "map_variables must return a dictionary with keys for all of the "
```
hinthornw, line 155: Wouldn't we get this for free when we invoke the model?

python/langsmith/evaluation/llm_evaluator.py
```diff
@@ -0,0 +1,221 @@
...
+class LLMEvaluator(RunEvaluator):
+    """A class for building LLM-as-a-judge evaluators."""
+
+    def __init__(
+        self,
+        *,
+        prompt_template: Union[str, List[Tuple[str, str]]],
+        score_config: Union[CategoricalScoreConfig, ContinuousScoreConfig],
+        map_variables: Optional[Callable[[Run, Optional[Example]], dict]] = None,
+        model: str = "gpt-3.5-turbo",
```
hinthornw, line 78: Probably shouldn't default to turbo right? Maybe don't even have a default

hinthornw: We should probably let them provide a chat model directly also, since langsmith users frequently have their own custom proxy, etc. that wouldn't be supported just with args to init_chat_model

hinthornw (commented):
Really like the idea of having an opinionated, easy evaluator def using tool calling

python/langsmith/evaluation/llm_evaluator.py
agola11: yeah but thought it would be better to fail fast at construction

python/langsmith/evaluation/llm_evaluator.py
agola11: I think it helps to have a default, esp if most people are using oai anyways. Will add something to init from chat model.

python/langsmith/evaluation/llm_evaluator.py
```diff
@@ -0,0 +1,285 @@
...
+from langsmith.schemas import Example, Run
+
+
+class CategoricalScoreConfig(BaseModel):
+    """Configuration for a categorical score."""
+
+    key: str
+    choices: List[str]
+    description: str
+    include_explanation: bool = False
```
hinthornw, line 17: I personally usually hyper-prompt engineer the explanation variable for my evals so don't love this being just a bool

agola11: will have this be a list of attributes or dictionary where you can add descriptions

samnoyes (approved):

PR #843, opened 2024-07-03
js[patch]: Inline lodash.set functionality to patch vulnerability

jacoblee93:
The `lodash.set` package is unmaintained and has a CVE: https://security.snyk.io/package/npm/lodash.set

The current `lodash` version has no CVEs: https://security.snyk.io/package/npm/lodash

I've inlined the relevant set functionality from the current version of `lodash`.

Let's be more careful with adding deps in the future!

@dqbd @nfcampos @agola11

### DIFF
--- a/js/.eslintrc.cjs
+++ b/js/.eslintrc.cjs
@@ -14,6 +14,7 @@ module.exports = {
   ignorePatterns: [
     ".eslintrc.cjs",
     "scripts",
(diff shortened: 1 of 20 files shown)


nfcampos (approved):

nfcampos:
We don't want loadsh as a dependency in python?

dqbd (approved):

PR #857, opened 2024-07-09
Add dataset split update + list methods

hinthornw:


samnoyes (approved):
might add a test? other than that lgtm

PR #864, opened 2024-07-11
feat: add prompt sdk functionality to langsmith sdk

madams0013:


_scripts/_fetch_schema.py
```diff
@@ -1,4 +1,5 @@
 """Fetch and prune the Langsmith spec."""
```
madams0013, line 1: only format

python/Makefile
```diff
@@ -18,6 +18,9 @@ doctest:
 evals:
 	poetry run python -m pytest tests/evaluation
 
+prompts:
```
hinthornw, line 21: these won't be run in CI right now

agola11 (approved):

python/langsmith/client.py
```diff
@@ -4560,6 +4560,543 @@ def _evaluate_strings(
...
+
+        try:
+            from langchain_core.load.dump import dumps
+        except ImportError:
+            raise ImportError(
+                "The client.create_commit function requires the langchain_core"
+                "package to run.\nInstall with `pip install langchain_core`"
+            )
+
+        json_object = dumps(object)
```
agola11, line 4853: probably should be more defensive here as the typing allows for any object

madams0013: added some tests for improper manifest formats, but through this noticed that we need more validation in the backend anyway as to what we accept for prompt manifests. will add to the backend.

python/langsmith/client.py
```diff
@@ -4664,6 +4664,538 @@ def _evaluate_strings(
             **kwargs,
         )
 
+    def _get_settings(self) -> dict:
+        """Get the settings for the current tenant.
+
+        Returns:
+            dict: The settings for the current tenant.
+        """
+        response = self.request_with_retries("GET", "/settings")
```
hinthornw, line 4673: Would be good to cache this so we don't have to send 2 api requests on every pull

python/langsmith/client.py
```diff
@@ -4664,6 +4664,538 @@ def _evaluate_strings(
...
+            limit (int): The maximum number of commits to fetch. Defaults to 1.
+            offset (int): The number of commits to skip. Defaults to 0.
+
+        Returns:
+            Optional[str]: The latest commit hash, or None if no commits are found.
+        """
+        response = self.request_with_retries(
+            "GET",
+            f"/commits/{prompt_owner_and_name}/",
+            params={"limit": limit, "offset": offset},
```
hinthornw, line 4713: ooc: so we're guaranteed to always get these in reverse order by time? Would that ever change?

python/langsmith/client.py
```diff
@@ -4664,6 +4664,538 @@ def _evaluate_strings(
...
+
+        """
+        owner, prompt_name, commit_hash = ls_utils.parse_prompt_identifier(
+            prompt_identifier
+        )
+
+        if not self._current_tenant_is_owner(owner):
+            return f"{self._host_url}/hub/{owner}/{prompt_name}:{commit_hash[:8]}"
+
+        settings = self._get_settings()
```
hinthornw, line 4758: Ya caching seems important. We're calling this twice in this method alone!

python/langsmith/client.py
```diff
@@ -4664,6 +4664,538 @@ def _evaluate_strings(
...
+                Defaults to "latest".
+
+        Returns:
+            str: The url of the prompt commit.
+
+        Raises:
+            HTTPError: If the server request fails.
+            ValueError: If the prompt does not exist.
+        """
+        if not self._prompt_exists(prompt_identifier):
```
hinthornw, line 4939: Why isn't this handled in the backend? Seems extraneous to add an additional API call here

python/langsmith/client.py
```diff
@@ -4664,6 +4664,538 @@ def _evaluate_strings(
...
+
+        Raises:
+            ValueError: If the prompt_identifier is empty.
+            HTTPError: If the server request fails.
+        """
+        settings = self._get_settings()
+        if is_public and not settings.get("tenant_handle"):
+            raise ValueError(
+                "Cannot create a public prompt without first\n"
+                "creating a LangChain Hub handle. "
```
hinthornw, line 5003: Is it still called the LangChain Hub??

python/langsmith/client.py
```diff
@@ -4664,6 +4664,538 @@ def _evaluate_strings(
...
+            json["tags"] = tags
+
+        owner, prompt_name, _ = ls_utils.parse_prompt_identifier(prompt_identifier)
+        response = self.request_with_retries(
+            "PATCH", f"/repos/{owner}/{prompt_name}", json=json
+        )
+        response.raise_for_status()
+        return response.json()
+
+    def delete_prompt(self, prompt_identifier: str) -> Any:
```
hinthornw, line 5028: Is the return type really  Any?

python/langsmith/client.py
```diff
@@ -4664,6 +4664,538 @@ def _evaluate_strings(
...
+
+        response = self.request_with_retries("DELETE", f"/repos/{owner}/{prompt_name}")
+
+        return response
+
+    def pull_prompt_object(
+        self,
+        prompt_identifier: str,
+        *,
+        include_model: Optional[bool] = False,
```
hinthornw, line 5052: ooc: why would we exclude by default?

python/langsmith/client.py
```diff
@@ -4664,6 +4664,538 @@ def _evaluate_strings(
...
+                "package to run.\nInstall with `pip install langchain_core`"
+            )
+
+        prompt_object = self.pull_prompt_object(
+            prompt_identifier, include_model=include_model
+        )
+        prompt = loads(json.dumps(prompt_object.manifest))
+
+        if isinstance(prompt, BasePromptTemplate) or isinstance(
+            prompt.first, BasePromptTemplate
```
hinthornw, line 5118: Is there always going to be a `first` value if it's not a prompt template? (i.e., do we always enforce it'll be a runnable sequence then)?

python/langsmith/client.py
```diff
@@ -4664,6 +4664,538 @@ def _evaluate_strings(
...
+              Defaults to an empty string.
+            tags (Optional[List[str]]): A list of tags for the prompt.
+              Defaults to an empty list.
+
+        Returns:
+            str: The URL of the prompt.
+
+        """
+        # Create or update prompt metadata
+        if self._prompt_exists(prompt_identifier):
```
hinthornw, line 5171: would prefer a single upsert endpoint eventually

python/langsmith/client.py
madams0013: backwards compatibility -- historically pull_prompt returns a PromptTemplate. but if you want the RunnableSequence including the model, you can specify that

python/langsmith/client.py
madams0013: yes

python/langsmith/client.py
madams0013: right now it's hardcoded in the backend as `ORDER BY created_at DESC`
should probably accept sort type and direction params, but will do that at a later date

python/langsmith/client.py
madams0013: added better checking

python/langsmith/client.py
madams0013: good catch it's None

python/langsmith/schemas.py
```diff
@@ -744,3 +744,96 @@ def metadata(self) -> dict[str, Any]:
         if self.extra is None or "metadata" not in self.extra:
             return {}
         return self.extra["metadata"]
+
+
+class PromptObject(BaseModel):
```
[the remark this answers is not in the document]
madams0013: renamed to PromptCommit

fetching the prompt with metadata is quite slow since it involves some computations and sometimes you just want the manifest, not the number of downloads or something

pull_prompt fetches the manifest, we could add an arg like "include_metadata" to get all information

python/langsmith/client.py
madams0013: yeah that makes sense

python/langsmith/client.py
```diff
@@ -4664,6 +4664,553 @@ def _evaluate_strings(
             **kwargs,
         )
 
+    @functools.lru_cache(maxsize=1)
+    def _get_settings(self) -> dict:
```
hinthornw, line 4668: This will impact the expected refcount to this client instance, which would make it so the tracer thread isn't garbage collected.

Would implement this in a similar way to how we handle the `def info()` caching with an explicit private attribute

hinthornw (changes requested):

python/langsmith/client.py
hinthornw: but this is a new method right?

python/langsmith/client.py
madams0013: have updated!

python/langsmith/client.py
```diff
@@ -701,6 +704,19 @@ def info(self) -> ls_schemas.LangSmithInfo:
...
 
+    def _get_settings(self) -> dict:
+        """Get the settings for the current tenant.
+
+        Returns:
+            dict: The settings for the current tenant.
+        """
+        if self._settings is None:
+            response = self.request_with_retries("GET", "/settings")
+            settings: dict = response.json()
```
hinthornw, line 715: would be kinda nice to have a better error with `ls_utils.raise_for_status_with_text(response)`

python/langsmith/client.py
```diff
@@ -4664,6 +4680,543 @@ def _evaluate_strings(
...
+        """Check if the current workspace has the same handle as owner.
+
+        Args:
+            owner (str): The owner to check against.
+
+        Returns:
+            bool: True if the current tenant is the owner, False otherwise.
+        """
+        settings = self._get_settings()
+        return owner == "-" or settings["tenant_handle"] == owner
```
hinthornw, line 4693: Would be kinda nice to TypedDict the settings at some opint

python/langsmith/client.py
```diff
@@ -4664,6 +4680,543 @@ def _evaluate_strings(
...
+        except ls_utils.LangSmithNotFoundError:
+            return None
+
+    def create_prompt(
+        self,
+        prompt_identifier: str,
+        *,
+        description: Optional[str] = None,
+        readme: Optional[str] = None,
+        tags: Optional[List[str]] = None,
```
hinthornw, line 4878: Nit; the inputs should be `Sequence[str]` rather than `list[str]`

python/langsmith/client.py
```diff
@@ -4664,6 +4680,543 @@ def _evaluate_strings(
...
+        response = self.request_with_retries("POST", "/repos/", json=json)
+        response.raise_for_status()
+        return ls_schemas.Prompt(**response.json()["repo"])
+
+    def create_commit(
+        self,
+        prompt_identifier: str,
+        object: Any,
+        *,
+        parent_commit_hash: Optional[str] = "latest",
```
hinthornw, line 4929: ooc: does this fetch if this is None?

python/langsmith/client.py
```diff
@@ -4664,6 +4680,543 @@ def _evaluate_strings(
...
+        return prompt
+
+    def push_prompt(
+        self,
+        prompt_identifier: str,
+        *,
+        object: Optional[Any] = None,
+        parent_commit_hash: Optional[str] = "latest",
+        is_public: bool = False,
+        description: Optional[str] = "",
```
hinthornw, line 5163: If this is optional, why do we default to `""`? Is there a difference between empty string and `None`?

python/langsmith/client.py
```diff
@@ -4664,6 +4680,543 @@ def _evaluate_strings(
...
+        readme: Optional[str] = "",
+        tags: Optional[List[str]] = [],
```
hinthornw, line 5165: We shouldn't default to a list (a mutable type) in the type signature. Should initialize the list within the function itself

python/langsmith/client.py
```diff
@@ -4811,3 +5364,72 @@ def _tracing_sub_thread_func(
...
+
+    try:
+        return openai._get_request_payload(messages, stop=stop, **kwargs)
+    except Exception as e:
+        raise ls_utils.LangSmithError(f"Error converting to OpenAI format: {e}")
+
+
+def convert_prompt_to_anthropic_format(
+    messages: Any,
+    model_name: str = "claude-2",
```
hinthornw, line 5402: Why default to cluade-2? Feels like it's gonna be deprecated soon.

Also why make only `stop` configurable and not the base url, etc.? Would probably be better tojust ahve a `model_kwargs` or something

hinthornw (commented):
Looks mostly good! Some comments/quesitons about the method signatures

python/langsmith/client.py
madams0013: changed it to model_kwargs and defaulted to "claude-3-haiku-20240307"

python/langsmith/client.py
madams0013: changed to None

python/langsmith/client.py
madams0013: good point, it won't work, took away the optional, left the default

python/langsmith/client.py
madams0013: added

hinthornw (approved):

PR #865, opened 2024-07-11
Fix up contextvar propagation

hinthornw:


python/tests/evaluation/test_evaluation.py
```diff
@@ -41,7 +41,6 @@ def predict(inputs: dict) -> dict:
         },
         num_repetitions=3,
     )
-    results.wait()
```
agola11, line 44: why was this removed?

hinthornw: Wasn't necessary since we were blocking in the function anyway

PR #869, opened 2024-07-12
feat: methods to convert prompt to openai and anthropic format

madams0013:


python/langsmith/client.py
```diff
@@ -4928,6 +4929,61 @@ def push_prompt(
...
+            dict: The prompt in OpenAI format.
+        """
+        from langchain_openai import ChatOpenAI
+
+        openai = ChatOpenAI()
+
+        try:
+            return openai._get_request_payload(messages, stop=stop, **kwargs)
+        except Exception as e:
+            print(e)
```
hinthornw, line 4954: Why not raise an error?

agola11 (approved):

PR #878, opened 2024-07-17
fix(js): pass other traceable options in wrapSDK

dqbd:


js/src/wrappers/openai.ts
```diff
@@ -312,6 +312,10 @@ const _wrapClient = <T extends object>(
   });
 };
 
+type WrapSDKOptions = Partial<
+  Omit<RunTreeConfig, "name"> & { runName: string }
```
jacoblee93, line 316: Should we accept both and mark `runName` as deprecated?

js/src/wrappers/openai.ts
```diff
@@ -325,9 +329,14 @@ const _wrapClient = <T extends object>(
...
 export const wrapSDK = <T extends object>(
   sdk: T,
-  options?: { client?: Client; runName?: string }
+  options?: WrapSDKOptions
 ): T => {
-  return _wrapClient(sdk, options?.runName ?? sdk.constructor?.name, {
-    client: options?.client,
-  });
+  const traceableOptions = options ? { ...options } : undefined;
+  if (traceableOptions != null) delete traceableOptions.runName;
```
jacoblee93, line 335: Just use `!== undefined`?

js/src/wrappers/openai.ts
```diff
@@ -280,18 +280,17 @@ export const wrapOpenAI = <T extends OpenAIType>(
...
     get(target, propKey, receiver) {
       const originalValue = target[propKey as keyof T];
       if (typeof originalValue === "function") {
         return traceable(
           originalValue.bind(target),
-          Object.assign(
-            { name: [runName, propKey.toString()].join("."), run_type: "llm" },
-            options
-          )
+          Object.assign({ run_type: "llm" }, options, {
```
jacoblee93, line 291: OOC why this over a spread operator?

jacoblee93 (approved):

PR #884, opened 2024-07-18
feat(datasets): add support for bulk updating examples

samnoyes:
Do not land until the langsmith pr hits prod


python/langsmith/client.py
```diff
@@ -3336,6 +3336,72 @@ def update_example(
...
+                "id": id_,
+                "inputs": in_,
+                "outputs": out_,
+                "dataset_id": dataset_id_,
+                "metadata": metadata_,
+                "split": split_,
+            }
+            for id_, in_, out_, metadata_, split_, dataset_id_ in zip(
+                example_ids,
+                inputs or [None] * len(example_ids),
```
hinthornw, line 3382: Just to double check, sending `None` doesn't erase the value does it? 

I'm guessing in this framing, the way you'd erase the value is with an empty dict then

hinthornw (approved):

Issue #882, opened 2024-07-18
Add an `atrace` context manager? Or handle serialization/copying in `trace` in a separate thread?


ldorigo:
### Feature request

`trace` runs everything on the main thread which means that filtering, serialization etc. all has to happen synchronously with the main logic and can significantly slow things down. As far as I understand this is not the case with the tracing that happens in langchain. Could you add an `atrace` version of `trace` or implement `trace` such that it doesn't block the main thread?

### Motivation

Tracing shouldn't have massive impacts on the application performance.

hinthornw:
Ya I was thinking of adding this actually. Makes sense to me

ldorigo:
I'm not sure how exactly it's implemented so far but it would perhaps make sense to push all of langsmith's processing in a whole separate process (so run it in parallel, not concurrently). `deepcopy`'s and serializations are CPU-bound and can be really slow for big objects, so doing it asynchronously will still impact the traced application's performance (in the sense that it will consume a lot of CPU cycles which are shared by all async functions).

hinthornw:
Ya we've talked about it. It would add complexity since multiprocessing isn't supported on all machines and we'd still have to be using ipc. 
First step is definitely making sure io-bound tasks aren't run in the main thread

ldorigo:
Sure, I understand. Do you have a rough idea of when the first non-blocking version might be released? (this is currently blocking us, I need to decide whether to implement a workaround myself or just wait for it to be supported :-) )

ldorigo:
I'm not sure this really solved the problem - consider the following setup:


```
def slow_langsmith(inputs: dict) -> dict:
    for i in range(30):
        time.sleep(1)
        print("waiting")
    return inputs


LANGSMITH_CLIENT = LangSmithClient(
    hide_inputs=filter_langsmith, hide_outputs=filter_langsmith, timeout_ms=(250000, 250000)
)

```

If I do 

```
async with trace(inputs={}):
    print("tracing!")
```

This still seems to wait for the entire 30 seconds before actual "tracing" gets printed.

ldorigo:
(@hinthornw ; not sure you get notifs for closed issues)

PR #887, opened 2024-07-19
Async trace context manager

hinthornw:
exposed via the same `trace` CM.

Would resolve https://github.com/langchain-ai/langsmith-sdk/issues/882

### DIFF
--- a/python/langsmith/run_helpers.py
+++ b/python/langsmith/run_helpers.py
@@ -699,8 +699,6 @@ class trace:
         Type of run (e.g., "chain", "llm", "tool"). Defaults to "chain".
     inputs : Optional[Dict], optional
         Initial input data for the run
-    extra : Optional[Dict], optional
-        Additional run-specific data not covered by other parameters
     project_name : Optional[str], optional
         Associates the run with a specific project, overriding defaults
     parent : Optional[Union[run_trees.RunTree, str, Mapping]], optional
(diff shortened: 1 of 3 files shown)


python/langsmith/run_helpers.py
```diff
@@ -685,6 +686,256 @@ def generator_wrapper(
...
+        self.metadata = metadata
+        self.client = client
+        self.run_id = run_id
+        self.reference_example_id = reference_example_id
+        self.exceptions_to_handle = exceptions_to_handle
+        self.new_run: Optional[run_trees.RunTree] = None
+        self.old_ctx: Optional[dict] = None
+
+    def _setup(self) -> run_trees.RunTree:
+        """Set up the tracing context and create a new run.
```
hinthornw, line 789: All copied over from before

python/langsmith/run_helpers.py
```diff
@@ -685,6 +686,256 @@ def generator_wrapper(
...
+        """
+        self._teardown(exc_type, exc_value, traceback)
+
+    async def __aenter__(self) -> run_trees.RunTree:
+        """Enter the context manager asynchronously.
+
+        Returns:
+            run_trees.RunTree: The newly created run.
+        """
+        return await aitertools.aio_to_thread(self._setup)
```
hinthornw, line 921: main change (here and aexit)

samnoyes (approved):

PR #893, opened 2024-07-23
feat: add hub js sdk to langsmith js sdk

madams0013:


js/src/client.ts
```diff
@@ -746,6 +758,13 @@ export class Client {
     return true;
   }
 
+  protected async _getSettings() {
+    if (!this.settings) {
+      this.settings = await this._get("/settings");
```
dqbd, line 763: Should we store a promise instead? This does not necessarily dedupe settings requests

js/src/client.ts
```diff
@@ -2921,4 +2940,566 @@ export class Client {
...
+    return this._likeOrUnlikePrompt(promptIdentifier, true);
+  }
+
+  public async unlikePrompt(
+    promptIdentifier: string
+  ): Promise<LikePromptResponse> {
+    return this._likeOrUnlikePrompt(promptIdentifier, false);
+  }
+
+  public async listCommits(
```
dqbd, line 3046: Shouldn't we follow the same convention of using AsyncIterable?

js/src/client.ts
```diff
@@ -2921,4 +2940,566 @@ export class Client {
...
+        originalRepoFullName: result.original_repo_full_name,
+        upstreamRepoFullName: result.upstream_repo_full_name,
+      })),
+      total: res.total,
+    };
+  }
+
+  public async getPrompt(promptIdentifier: string): Promise<Prompt | null> {
+    const [owner, promptName, _] = parsePromptIdentifier(promptIdentifier);
+    const response = await this.caller.call(
```
dqbd, line 3151: I think we should use `raiseForStatus` after receiving every response to properly test of `!response.ok`

js/src/client.ts
```diff
@@ -2921,4 +2946,527 @@ export class Client {
...
+      `${this.apiUrl}/repos/${owner}/${promptName}`,
+      {
+        method: "GET",
+        headers: this.headers,
+        signal: AbortSignal.timeout(this.timeout_ms),
+        ...this.fetchOptions,
+      }
+    );
+
+    if (response.status === 404) {
```
dqbd, line 3176: Should we also handle `!res.ok` as well?

js/src/client.ts
```diff
@@ -2921,4 +2946,527 @@ export class Client {
...
+      is_public: !!options?.isPublic,
+    };
+
+    const response = await this.caller.call(fetch, `${this.apiUrl}/repos/`, {
+      method: "POST",
+      headers: { ...this.headers, "Content-Type": "application/json" },
+      body: JSON.stringify(data),
+      signal: AbortSignal.timeout(this.timeout_ms),
+      ...this.fetchOptions,
+    });
```
dqbd, line 3180: _nit_: should we handle `!response.ok` as well?

js/src/client.ts
```diff
@@ -2921,4 +2946,527 @@ export class Client {
...
+    return {
+      owner,
+      repo: promptName,
+      commit_hash: result.commit_hash,
+      manifest: result.manifest,
+      examples: result.examples,
+    };
+  }
+
+  public async _pullPrompt(
```
dqbd, line 3375: _nit_: maybe worth adding `@private` JSDoc tag + indicate what user should do?

js/src/client.ts
```diff
@@ -2921,4 +2946,527 @@ export class Client {
...
+    } else {
+      await this.createPrompt(promptIdentifier, {
+        description: options?.description,
+        readme: options?.readme,
+        tags: options?.tags,
+        isPublic: options?.isPublic,
+      });
+    }
+
+    if (options?.object === null) {
```
dqbd, line 3416: Don't we need to check for undefined as well?

dqbd (approved):

PR #907, opened 2024-08-02
fix: pushing new manifest without metadata update

madams0013:
When pushing an updated manifest to an existing prompt and not specifying any other options, update_prompt threw an error that we need to be updating something, but we're updating the manifest. This fixes.


python/langsmith/client.py
```diff
@@ -5036,7 +5036,7 @@ def create_prompt(
         description: Optional[str] = None,
         readme: Optional[str] = None,
         tags: Optional[Sequence[str]] = None,
-        is_public: bool = False,
+        is_public: Optional[bool] = False,
```
hinthornw, line 5039: what's the need for this?

python/langsmith/client.py
```diff
@@ -5350,13 +5350,20 @@ def push_prompt(
...
         # Create or update prompt metadata
         if self._prompt_exists(prompt_identifier):
-            self.update_prompt(
-                prompt_identifier,
-                description=description,
-                readme=readme,
-                tags=tags,
-                is_public=is_public,
-            )
+            if not (
```
hinthornw, line 5353: nit: easier to read if we flip the conditions:

```
if (all this stuff is none) then create prompt, else update

Comment (madams0013) [reply python/langsmith/client.py:5353]: we don't want to create the prompt at all in this case, but if we have an existing prompt and no metadata is passed we don't want to call update_prompt since there's nothing to update

this is just the case where we're updating the manifest and not the metadata

Comment (hinthornw) [reply python/langsmith/client.py:5372]: I'm just saying it's hard to read a double negative (not x is None); the logic is right

File: python/langsmith/client.py [hunk 3]
```diff
@@ -5093,7 +5093,7 @@ def create_prompt(
             "description": description or "",
             "readme": readme or "",
             "tags": tags or [],
-            "is_public": is_public,
+            "is_public": is_public or False,
```
hinthornw, line 5096: This isn't needed right?

hinthornw (approved):
Awesome!

PR #921, opened 2024-08-12
[Python] fix: handle vals that can't have truthiness checks

hinthornw:


python/langsmith/run_helpers.py
```diff
@@ -429,9 +429,12 @@ def _on_run_end(
         error: Optional[BaseException] = None,
     ) -> None:
         """Handle the end of run."""
-        if outputs and outputs_processor is not None:
-            outputs = outputs_processor(outputs)
-        _container_end(container, outputs=outputs, error=error)
+        try:
+            if outputs_processor is not None:
```
hinthornw, line 433: No need fo rthe truthiness checks

PR #922, opened 2024-08-13
feat: schema validation in langsmith sdk

jakerachleff:


python/tests/integration_tests/test_client.py
```diff
@@ -312,11 +313,7 @@ def test_error_surfaced_invalid_uri(monkeypatch: pytest.MonkeyPatch, uri: str) -
         client.create_run("My Run", inputs={"text": "hello world"}, run_type="llm")
 
 
-def test_create_dataset(
-    monkeypatch: pytest.MonkeyPatch, langchain_client: Client
-) -> None:
-    """Test persisting runs and adding feedback."""
-    monkeypatch.setenv("LANGCHAIN_ENDPOINT", "https://dev.api.smith.langchain.com")
```
jakerachleff, line 319: This was a quirk where we were overriding a single test to use dev. If we wanna test against dev, we should just configure the suite to run against dev in addition

python/langsmith/client.py
```diff
@@ -2529,6 +2529,8 @@ def create_dataset(
         *,
         description: Optional[str] = None,
         data_type: ls_schemas.DataType = ls_schemas.DataType.kv,
+        inputs_schema_definition: Optional[Dict[str, Any]] = None,
```
hinthornw, line 2532: is "definition" necessary? My fingers hurt just looking at this

jakerachleff: I matched the API exactly. Do we mismatch ever otherwise?

hinthornw: We do mismatch - our apis are pretty confusing sometimes

python/langsmith/schemas.py
```diff
@@ -128,10 +128,17 @@ class DatasetBase(BaseModel):
...
+        None, alias="inputs_schema_definition"
+    )
+    outputs_schema: Optional[Dict[str, Any]] = Field(
+        None, alias="outputs_schema_definition"
+    )
 
     class Config:
         """Configuration class for the schema."""
 
+        allow_population_by_field_name = True
```
jakerachleff, line 141: @hinthornw I had to do some pydantic magic to make all this work. Do we do this in the SDK? I see this pattern in runtree, but I know pydantic stuff is frowned upon other parts of the code base.

hinthornw: Maybe we just don't use pydantic in thecreate_dataset method anymore?

jakerachleff: we'll need to remove it from all dataset related areas then, because we'll need to do conversion on any read/create/etc

python/langsmith/client.py
```diff
@@ -2546,18 +2548,28 @@ def create_dataset(
         Dataset
             The created dataset.
         """
-        dataset = ls_schemas.DatasetCreate(
-            name=dataset_name,
-            description=description,
-            data_type=data_type,
-        )
+        dataset = {
```
jakerachleff, line 2551: byebye pydantic

python/langsmith/schemas.py
```diff
@@ -163,6 +158,12 @@ def __init__(
         **kwargs: Any,
     ) -> None:
         """Initialize a Dataset object."""
+        if "inputs_schema_definition" in kwargs:
```
jakerachleff, line 161: this is disgusting

hinthornw: Would this actually be applied?

jakerachleff: wdym? yeah, there's a new integration test showing this works

python/tests/integration_tests/test_client.py
```diff
@@ -360,6 +357,59 @@ def test_create_dataset(
...
+    with pytest.raises(LangSmithError):
+        langchain_client.create_example(
+            inputs={"input": "hello world"},
+            outputs={"john": 1},
+            dataset_id=dataset.id,
+        )
+
+    # assert read API includes the schema definition
+    read_dataset = langchain_client.read_dataset(dataset_id=dataset.id)
+    assert read_dataset.inputs_schema == InputSchema.model_json_schema()
```
jakerachleff, line 407: @hinthornw here's the integration test for reading the input schema back out

python/langsmith/schemas.py
hinthornw: So this is for the dataset copying case ya?

I think it's fine. I also wouldn't resist if we just marked it as "inputs_schema_definition" itself here

hinthornw (approved):

PR #924, opened 2024-08-15
python: add Client.similar_examples() and Client.index_dataset()

baskaryan:
Add methods for interacting with `datasets/{dataset_id}/index` and `datasets/{dataset_id}/search` endpoints.

Indexing takes a second so the integration test is skipped by default. Could turn it on or could update to use an existing index instead of creating one as part of the test.


python/langsmith/client.py
```diff
@@ -3412,6 +3414,57 @@ def list_examples(
...
+        )
+        ls_utils.raise_for_status_with_text(few_shot_resp)
+        examples = []
+        for res in few_shot_resp.json()["examples"]:
+            examples.append(
+                ls_schemas.ExampleSearch(
+                    **res,
+                    dataset_id=dataset_id,
+                    _host_url=self._host_url,
+                    _tenant_id=self._get_optional_tenant_id(),
```
hinthornw, line 3463: I'd maybe not include this here if we want it to be fast.

python/langsmith/client.py
```diff
@@ -3412,6 +3414,57 @@ def list_examples(
...
+        dataset_name: Optional[str] = None,
+        **kwargs: Any,
+    ) -> List[ls_schemas.ExampleBase]:
+        """Retrieve the dataset examples whose inputs best match the query.
+
+        **Note**: Must have few-shot indexing enabled for the dataset. See (TODO) method
+        for how to enable indexing.
+
+        Args:
+            query (dict): The query to search against. Must be JSON serializable.
```
hinthornw, line 3433: Maybe note that it must match the schema of the dataset?

python/langsmith/client.py
```diff
@@ -3412,6 +3414,57 @@ def list_examples(
             if limit is not None and i + 1 >= limit:
                 break
 
+    @ls_utils.xor_args(("dataset_name", "dataset_id"))
+    def search_examples(
+        self,
+        query: dict,
```
jakerachleff, line 3420: can we call this input? There are many types of queries we might want to support (metadata filtering, actual query if we make this a retriever eventually, etc) so would rather be explicit for few shot for now?

python/langsmith/client.py
```diff
@@ -3412,6 +3414,57 @@ def list_examples(
...
+            dataset_name (str, optional): The name of the dataset to filter by.
+                Defaults to None. Must specify one of ``dataset_id`` or
+                ``dataset_name``.
+            kwargs (Any): Additional keyword args to pass as part of request body.
+
+        Returns:
+            List of ExampleSearch.
+        """
+        if dataset_id is None:
+            dataset_id = self.read_dataset(dataset_name=dataset_name).id
```
hinthornw, line 3447: Thoughts on not even supporting the name arg for now to avoid 2 API calls?

python/langsmith/client.py
hinthornw: How about make it positional only?

hinthornw: oh it is lol

hinthornw: i like

baskaryan: updated

python/langsmith/client.py
```diff
@@ -3412,6 +3412,43 @@ def list_examples(
             if limit is not None and i + 1 >= limit:
                 break
 
+    # dataset_name explicitly not supported to avoid extra API calls.
+    def search_examples(
```
baskaryan, line 3416: could rename it "few_shot_search" or something to make it more clear that the query is inputs and the output is meant to be injected into a prompt (ie the examples are bare, missing metadata that full Examples have)

jakerachleff: makes sense tor me, but also defer to @hinthornw for approval

hinthornw (approved):

python/tests/integration_tests/test_client.py
```diff
@@ -269,6 +269,50 @@ def test_list_examples(langchain_client: Client) -> None:
...
+            "required": ["response"],
+            "additionalProperties": False,
+        },
+    )
+    langchain_client.create_examples(
+        inputs=inputs, outputs=outputs, dataset_id=dataset.id
+    )
+    langchain_client.index_dataset(dataset_id=dataset.id)
+    # Need to wait for indexing to finish.
+    time.sleep(5)
```
jakerachleff, line 307: we need to figure out a way to not make people need to do this...

baskaryan: yea ideally you could poll it till its done. but also this is a one-time thing for any dataset so wonder if it's that high prio.

hinthornw: Is there an indicator in the output to `read_dataset` we could use?

hinthornw (approved):

PR #925, opened 2024-08-15
tmp

hinthornw:


python/langsmith/async_client.py
```diff
@@ -0,0 +1,714 @@
...
+)
+
+import httpx
+
+from langsmith import client as ls_client
+from langsmith import schemas as ls_schemas
+from langsmith import utils as ls_utils
+from langsmith._internal import _beta_decorator as ls_beta
+
+logger = logging.getLogger(__name__)
```
jakerachleff, line 30: nit: should we use an async logger here? We do that with `structlog` on langsmith

python/langsmith/utils.py
```diff
@@ -123,7 +124,9 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
     return decorator
 
 
-def raise_for_status_with_text(response: requests.Response) -> None:
+def raise_for_status_with_text(
+    response: Union[requests.Response, httpx.Response],
+) -> None:
```
jakerachleff, line 129: Do we need to `except httpx.HTTPError` as well here?

python/tests/integration_tests/test_async_client.py
```diff
@@ -0,0 +1,240 @@
...
+
+        example = await client.create_example(
+            inputs={"name": "Alice", "age": 30},
+            outputs={"hi": "hello"},
+            dataset_id=dataset.id,
+        )
+
+        await client.index_dataset(dataset_id=dataset.id)
+        for _ in range(10):
+            examples = await client.similar_examples(
```
jakerachleff, line 34: did you mean to use `wait_for` here?

jakerachleff (approved):
approved with some comments

python/tests/integration_tests/test_async_client.py
hinthornw: maybe - can add!

python/langsmith/async_client.py
hinthornw: TIL. I actually am not using this anymore so will just remove. Think async logger would add another pckg right?

python/langsmith/utils.py
hinthornw: Oh ya!

PR #930, opened 2024-08-20
python[patch]: Return cloned dataset

baskaryan:


hinthornw (approved):

baskaryan:
decided it makes more sense to mirror `create_dataset` api

python/langsmith/client.py
```diff
@@ -2969,11 +2973,15 @@ def clone_public_dataset(
...
         ds = source_client.read_shared_dataset(token_uuid)
         dataset_name = dataset_name or ds.name
-        if self.has_dataset(dataset_name=dataset_name):
+        try:
+            ds = self.read_dataset(dataset_name=dataset_name)
             logger.info(
                 f"Dataset {dataset_name} already exists in your tenant. Skipping."
             )
-            return
+            return str(ds.id)
```
hinthornw, line 2981: This would need updated ya

PR #933, opened 2024-08-20
feat: support few shot in js

jakerachleff:
Please review closely because I am not a JS dev

Adds two few shot related methods for indexing and finding similar examples to the js langsmith client. Mirrors implementations in python here: https://github.com/langchain-ai/langsmith-sdk/pull/925/files#diff-619399e55cdfbfd5de40fb5e942606b34a9d5e16c289ae44e02d7a4a46d82e4fR14


js/src/client.ts
```diff
@@ -2148,6 +2162,80 @@ export class Client {
...
+    tag,
+  }: {
+    datasetId?: string;
+    datasetName?: string;
+    tag?: string;
+  }): Promise<void> {
+    let datasetId_ = datasetId;
+    if (datasetId_ === undefined && datasetName === undefined) {
+      throw new Error("Must provide either datasetName or datasetId");
+    } else if (datasetId_ !== undefined && datasetName !== undefined) {
```
bracesproul, line 2177: I would do `if (!datasetId_ && ...)` instead of undefined checks because someone could pass in two empty strings, which would pass the `undefined` checks but not actually be defined

Same idea for the other checks too

js/src/schemas.ts
```diff
@@ -251,6 +251,11 @@ export interface ExampleUpdate {
 export interface ExampleUpdateWithId extends ExampleUpdate {
   id: string;
 }
+
+export interface SimilarExample extends BaseExample {
```
hinthornw, line 255: nit: we called it ExampleSearch in python

js/src/client.ts
```diff
@@ -2148,6 +2162,81 @@ export class Client {
...
+    );
+    if (!response.ok) {
+      throw new Error(
+        `Failed to index dataset ${datasetId_}: ${response.status} ${response.statusText}`
+      );
+    }
+    await response.json();
+  }
+
+  public async similarExamples(
```
hinthornw, line 2207: Want typedoc

js/src/client.ts
```diff
@@ -1892,7 +1893,14 @@ export class Client {
     {
       description,
       dataType,
-    }: { description?: string; dataType?: DataType } = {}
+      inputsSchema,
```
hinthornw, line 1896: do we want to add to the update method as well?

jakerachleff: eh I can add later? I only need this for passing integration tests

js/src/client.ts
jakerachleff: added

agola11 (approved):

PR #937, opened 2024-08-21
js[minor]: Add clonePublicDataset and listSharedExamples

bracesproul:


js/src/schemas.ts
```diff
@@ -270,6 +270,8 @@ export interface Dataset extends BaseDataset {
   example_count?: number;
   session_count?: number;
   last_session_start_time?: number;
+  inputs_schema_definition?: KVMap;
+  outputs_schema_definition?: KVMap;
```
bracesproul, line 274: cc @jakerachleff is this the proper interface to add these fields to? Or should they be on `BaseDataset`?

hinthornw: It can be on the base i believe (that way in BE)

js/src/client.ts
```diff
@@ -1528,6 +1528,53 @@ export class Client {
...
+        method: "GET",
+        headers: this.headers,
+        signal: AbortSignal.timeout(this.timeout_ms),
+        ...this.fetchOptions,
+      }
+    );
+    const result = await response.json();
+    if (!response.ok) {
+      throw new Error(
+        `Failed to list shared examples: ${response.status} ${response.statusText}`
```
hinthornw, line 1569: Possible to include the full error message?

js/src/client.ts
```diff
@@ -1528,6 +1528,53 @@ export class Client {
     return dataset as Dataset;
   }
 
+  /**
+   * Get shared examples.
+   * @param {string} shareToken The share token to get examples for.
```
hinthornw, line 1533: Maybe say what a share token is / how to get it?

bracesproul: To confirm a share token is the UUID (or UUID w/ smith url) generated when you share something (e.g a run)?

hinthornw: Ya!

hinthornw (approved):

PR #938, opened 2024-08-21
[JS] More Informative Errors

hinthornw:
1. Add error type for 409's
2. Include response body text in error message content


jakerachleff:
nice!! will let someone who knows js a bit better do signoff but this will be a huge improvement

jacoblee93 (approved):
Awesome!

dqbd (approved):

PR #940, opened 2024-08-22
[Py] Handle experiment name conflicts

hinthornw:


python/langsmith/evaluation/_runner.py
```diff
@@ -958,24 +957,28 @@ def _get_experiment_metadata(self):
...
         return project_metadata
 
-    def _get_project(self, first_example: schemas.Example) -> schemas.TracerSession:
-        if self._experiment is None:
+    def _create_experiment(
+        self, dataset_id: uuid.UUID, metadata: dict
+    ) -> schemas.TracerSession:
+        # There is a chance of name collision, so we'll retry
+        starting_name = self._experiment_name
+        while True:
```
agola11, line 965: probably should be a limit here

agola11 (approved):

PR #944, opened 2024-08-23
[Python] Fix incremental streaming of eval steps

hinthornw:


python/langsmith/evaluation/_runner.py
```diff
@@ -1084,7 +1084,7 @@ def dataset_id(self) -> str:
     @property
     def evaluation_results(self) -> Iterable[EvaluationResults]:
         if self._evaluation_results is None:
-            return [{"results": []} for _ in self.examples]
+            return ({"results": []} for _ in self.examples)
```
hinthornw, line 1087: This is the change

python/langsmith/evaluation/_arunner.py
```diff
@@ -329,6 +329,7 @@ async def aevaluate_existing(
         max_concurrency=max_concurrency,
         client=client,
         blocking=blocking,
+        experiment=project,
```
hinthornw, line 332: Not releated to incrementality - just realized in my UT we were doing an extra unnecessary call

agola11 (approved):

PR #957, opened 2024-08-28
Optimize CPU-bound Task Execution by Replacing Thread Lock with Producer-Consumer Pattern

adham-elarabawy:
## Issue
Seeing large slowdowns when using `evaluate()` compared to naive `ThreadPoolExecutor`. The specific slowdown happens on a cpu-bound call to BeautifulSoup, which balloons from 0.1s -> 5 seconds (and sometimes up to 60+ seconds based on concurrency).

## Profiling Investigation
First, I noticed that the majority of the time in bs4 init was actually empty space, which made me think that: (1) there was GIL contention, or (2) there was some lock being placed on the thread.
CleanShot 2024-08-28 at 13 09 14

I took a look at the `ExperimentResults` implementation, and noted that in the infinite while loop, there was an `is_active` call inside of the thread lock, and since `is_active` has nontrivial overhead, and this loop was running even if there were no results, we were incurring this overhead over and over.
CleanShot 2024-08-28 at 13 10 33

To confirm, I noticed that in the main thread profile, about half the time is spent in `is_alive` in aggregate.
CleanShot 2024-08-28 at 13 13 04

## Fix
To improve the implementation without blocking CPU-bound operations, we can use a producer-consumer pattern with a queue. This will allow the main thread to continue processing while the data is being fetched and processed in the background.

### DIFF
--- a/python/langsmith/evaluation/_runner.py
+++ b/python/langsmith/evaluation/_runner.py
@@ -377,7 +377,7 @@ class ExperimentResults:
     ):
         self._manager = experiment_manager
         self._results: List[ExperimentResultRow] = []
-        self._queue = queue.Queue()
+        self._queue: queue.Queue[ExperimentResultRow] = queue.Queue()
         self._processing_complete = threading.Event()
         self._thread = threading.Thread(target=self._process_data)
         self._thread.start()
--- a/python/langsmith/evaluation/_runner.py
+++ b/python/langsmith/evaluation/_runner.py
@@ -376,7 +376,7 @@ class ExperimentResults:
         experiment_manager: _ExperimentManager,
     ):
         self._manager = experiment_manager
-        self._results: List[ExperimentResultRow]= []
+        self._results: List[ExperimentResultRow] = []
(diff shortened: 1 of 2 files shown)


python/langsmith/evaluation/_runner.py
```diff
@@ -376,38 +377,45 @@ def __init__(
...
             with self._lock:
                 self._results.append(item)
-        summary_scores = manager.get_summary_scores()
+
+        summary_scores = self._manager.get_summary_scores()
         with self._lock:
             self._summary_results = summary_scores
 
+        self._queue.put(None)  # Sentinel value to indicate completion
+        self._processing_complete.set()
```
calvinqi, line 414: sorry noob question as an outsider coming in 😆: is this None sentinel redundant with `_processing_complete`? Seems like they both just serve as a check for the same thing

adham-elarabawy: Great point! These can be collapsed into one condition -- made the fix.

hinthornw (approved):

PR #971, opened 2024-09-04
[Python] Cache env var check

hinthornw:


python/langsmith/client.py
```diff
@@ -1115,7 +1115,7 @@ def _run_transform(
 
     @staticmethod
     def _insert_runtime_env(runs: Sequence[dict]) -> None:
-        runtime_env = ls_env.get_runtime_and_metrics()
+        runtime_env = ls_env.get_runtime_environment()
```
hinthornw, line 1118: We also don't currently plot the runtime metrics anywhere, so it doesn't seem useful to collect them

nfcampos (approved):

PR #972, opened 2024-09-04
[Python] Log content-length during connection errors

hinthornw:


python/langsmith/client.py
```diff
@@ -814,6 +827,9 @@ def request_with_retries(
                         f"Connection error caused failure to {method} {pathname}"
                         f"  in LangSmith API. {recommendation}."
                         f" {repr(e)}"
+                        f"\nContent-Length: {content_length}"
+                        f"\nAPI Key: {masked_api_key}"
```
barberscott, line 831: Great idea.  I assume this will look something like what we show in the LS UI?

python/langsmith/client.py
```diff
@@ -829,6 +845,7 @@ def request_with_retries(
                         emsg = msg
                     raise ls_utils.LangSmithError(
                         f"Failed to {method} {pathname} in LangSmith API. {emsg}"
+                        f"{_context}"
```
barberscott, line 848: Should we gate this behind some kind of flag to prevent unintended leakage into logs in error cases?  Thinking about whether there'd be PII in here (potentially?) and that we'd maybe want someone to set an easy flag to surface this?  I might be barking up wrong tree here.

hinthornw: Hm we could, though I think I'd want it to be opt-out since we're only returning run/trace IDs for the context?

barberscott: Ah OK, this is not the full payload.  Maybe *that* (pretty printed) is what we'd (eventually) want to add in as an opt-in.

hinthornw: Ya that could be an interesting opt-in bit. Right now I'm just doing the IDs so we can correlate them; anything else you'd want to see by default?

python/langsmith/client.py
hinthornw: Example from me:

```
WARNING:langsmith.client:Failed to batch ingest runs: langsmith.utils.LangSmithConnectionError: Connection error caused failure to POST https://api.smith.langchain.com/runs/batch  in LangSmith API. Please confirm your internet connection.. SSLError(MaxRetryError("HTTPSConnectionPool(host='api.smith.langchain.com', port=443): Max retries exceeded with url: /runs/batch (Caused by SSLError(SSLEOFError(8, 'EOF occurred in violation of protocol (_ssl.c:2393)')))"))
Content-Length: 105000932
API Key: lsv2*********************************************33
post: trace=fec2a5d9-bd40-4238-a7f0-ed4f23096d36,id=fec2a5d9-bd40-4238-a7f0-ed4f23096d36
```

PR #974, opened 2024-09-04
feat: metadata filters for few shot

jakerachleff:


python/langsmith/async_client.py
```diff
@@ -853,6 +854,9 @@ async def similar_examples(
                 input schema. Must be JSON serializable.
             limit (int): The maximum number of examples to return.
             dataset_id (str or UUID): The ID of the dataset to search over.
+            filter (str, optional): A filter string to apply to the search results. Uses
```
hinthornw, line 857: Could we add examples to this docstring?

The syntax is kinda hard to grok

hinthornw (approved):

python/langsmith/client.py
```diff
@@ -3500,6 +3501,12 @@ def similar_examples(
                 input schema. Must be JSON serializable.
             limit (int): The maximum number of examples to return.
             dataset_id (str or UUID): The ID of the dataset to search over.
+            filter (str, optional): A filter string to apply to the search results. Uses
+            the same syntax as the `filter` parameter in `list_runs()`. Only a subset
+            of operations are supported. Defaults to None.
+
+            For example, you can use `and(eq(metadata.some_tag, 'some_value'), neq(metadata.env, 'dev'))`
```
hinthornw, line 3508: The string formatting of this would need to be updateed right? Needs to have '' around metadata.some_tag"?

jakerachleff: updated lark to make this not necessary in my last pr :)

PR #975, opened 2024-09-04
feat: few shot example filtering in js

jakerachleff:


js/src/client.ts
```diff
@@ -2238,13 +2238,22 @@ export class Client {
   public async similarExamples(
     inputs: KVMap,
     datasetId: string,
-    limit: number
+    limit: number,
```
hinthornw, line 2241: Could we update teh typedoc :)

hinthornw (approved):

PR #977, opened 2024-09-05
[Python] Evals: submit feedback in thread

hinthornw:
And then will be even better once we add batch endpoint

### DIFF
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -3938,8 +3938,17 @@ class Client:
         run: Optional[ls_schemas.Run] = None,
         source_info: Optional[Dict[str, Any]] = None,
         project_id: Optional[ID_TYPE] = None,
+        *,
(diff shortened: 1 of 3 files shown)


python/langsmith/client.py
```diff
@@ -3938,8 +3938,17 @@ def _log_evaluation_feedback(
         run: Optional[ls_schemas.Run] = None,
         source_info: Optional[Dict[str, Any]] = None,
         project_id: Optional[ID_TYPE] = None,
+        *,
+        _executor: Optional[cf.ThreadPoolExecutor] = None,
     ) -> List[ls_evaluator.EvaluationResult]:
         results = self._select_eval_results(evaluator_response)
+
+        def _submit(**kwargs):
```
agola11, line 3946: Probably should be called _submit_feedback?

agola11 (approved):

PR #984, opened 2024-09-06
Faster Serialization

hinthornw:
1. Stop serializing arbitrary python objects
2. Stop recursing upon hitting unserializable object

This helps reduce overhead of tracing when you have large or unserializable content


hinthornw:
huh fun interaction ain't it

python/langsmith/client.py
```diff
@@ -151,74 +165,89 @@ def _default_retry_config() -> Retry:
...
+    buf: bytes
 
 
-def _simple_default(obj: Any) -> Any:
-    # Don't traverse into nested objects
+def _simple_default(obj):
     try:
+        # Only need to handle types that orjson doesn't serialize by default
+        # https://github.com/ijl/orjson#serialize
+        if isinstance(obj, _Fragment):
```
nfcampos, line 176: this shouldn't be here I think?

nfcampos (approved):

PR #985, opened 2024-09-07
fix(js): Adds fix for circular references in inputs and outputs

### LINKED ISSUE #962: bug(js): Errors thrown when serializing run data or trace data with circular references
I often see errors of this nature in my local JS agent:

```
TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Environment'
    |     property 'application' -> object with constructor 'Application'
    |     property 'environments' -> object with constructor 'Array'
    --- index 0 closes the circle
    at JSON.stringify (<anonymous>)
    at Client.createRun (/Users/airhorns/Code/gadget/node_modules/.pnpm/langsmith@0.1.43_patch_hash=5u7bmtv3vjzuhwq77bsoqw2f5u_@langchain+core@0.2.28_langchain@0.2.17_openai@4.56.0/node_modules/langsmith/dist/client.cjs:582:24)
    at runNextTicks (node:internal/process/task_queues:60:5)
    at processTimers (node:internal/timers:511:9)
    at async Client.batchIngestRuns (/Users/airhorns/Code/gadget/node_modules/.pnpm/langsmith@0.1.43_patch_hash=5u7bmtv3vjzuhwq77bsoqw2f5u_@langchain+core@0.2.28_langchain@0.2.17_openai@4.56.0/node_modules/langsmith/dist/client.cjs:635:17)
    at async Client.drainAutoBatchQueue (/Users/airhorns/Code/gadget/node_modules/.pnpm/langsmith@0.1.43_patch_hash=5u7bmtv3vjzuhwq77bsoqw2f5u_@langchain+core@0.2.28_langchain@0.2.17_openai@4.56.0/node_modules/langsmith/dist/client.cjs:490:17)
```

The `traceable` helper and automatic tracing of tools forcibly captures all the data returned from a tool or a traceable function. We use a bunch of rich objects for all these functions that sometimes have circular references within them. These objects throw errors when returned from traceable functions because they are forcibly JSON serialized.

It'd be great if we could return objects with circular references from our traceable functions. FWIW, OpenTelemetry doesn't attempt to serialize the results of arbitrary functions as it is hard to do right in the general case, insecure, and kinda expensive -- maybe LangSmith should consider doing the same for non-LLM-token return values. 

To work around this, I have patched langsmith locally to use https://www.npmjs.com/package/safe-stable-stringify instead of plain old `JSON.stringify`, but it is much less performant and I still get the heebie jeebies that a function call can leak all it's inputs and outputs out of the system while looking entirely innocent. I also went down the path of adding `.toJSON` definitions on all my objects, but that was laborious for the ones I own and hard/sketchy for the ones I'm importing from libraries.

Versions:
  - JS LangSmith and LangChain
  - `langsmith 0.1.43`
  - `@langchain/core 0.2.28`

### Suggestion:

_No response_

jacoblee93:
~~This does add a few stringify calls - we could do this more efficiently if needed but I am not sure it's worth prematurely optimizing~~

Fixes #962

### DIFF
--- a/js/src/utils/serde.ts
+++ b/js/src/utils/serde.ts
@@ -1,0 +1,22 @@
+export const CIRCULAR_VALUE_REPLACEMENT_STRING = "[Circular]";
+
+/**
+ * JSON.stringify version that handles circular references by replacing them
+ * with an object marking them as such ({ result: "[Circular]" }).
+ */
+export const stringifyForTracing = (value: any): string => {
+  const seen = new WeakSet();
+
+  const serializer = (_: string, value: any): any => {
+    if (typeof value === "object" && value !== null) {
+      if (seen.has(value)) {
+        return {
+          result: CIRCULAR_VALUE_REPLACEMENT_STRING,
+        };
+      }
+      seen.add(value);
+    }
+    return value;
+  };
+  return JSON.stringify(value, serializer);
+};
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -344,6 +344,29 @@ const handle429 = async (response?: Response) => {
   return false;
 };
 
+const sanitizeParams = (run: Record<string, any>) => {
+  const runCopy = { ...run };
+  try {
+    JSON.stringify(run.inputs);
+  } catch (e: any) {
+    runCopy.inputs = {
+      error: {
+        message: `Failed to serialize run inputs: ${e.message}`,
+      },
+    };
+  }
+  try {
+    JSON.stringify(run.outputs);
+  } catch (e: any) {
+    runCopy.outputs = {
+      error: {
+        message: `Failed to serialize run outputs: ${e.message}`,
+      },
+    };
+  }
+  return runCopy;
+};
+
 export class Queue<T> {
   items: [T, () => void][] = [];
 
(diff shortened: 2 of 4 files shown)


hinthornw (approved):
Think this should work. @dqbd could you also TAL?

airhorns:
What do you think about using https://www.npmjs.com/package/safe-json-stringify or the `util` module from node instead? When data has circular references I'd still like to see it in the tracing dashboard, you know?

jacoblee93:
Am open to that too - can have a look later.

dqbd (approved):

jacoblee93:
Updated to preserve the shape when encountering circular references

js/src/utils/serde.ts
```diff
@@ -0,0 +1,22 @@
...
+ * with an object marking them as such ({ result: "[Circular]" }).
+ */
+export const stringifyForTracing = (value: any): string => {
+  const seen = new WeakSet();
+
+  const serializer = (_: string, value: any): any => {
+    if (typeof value === "object" && value !== null) {
+      if (seen.has(value)) {
+        return {
+          result: CIRCULAR_VALUE_REPLACEMENT_STRING,
```
dqbd, line 14: Shouldn't we just return `CIRCULAR_VALUE_REPLACEMENT_STRING`?

dqbd, line 14: Shouldn't we just return `CIRCULAR_VALUE_REPLACEMENT_STRING`?

jacoblee93: Backend requires `input` and `output` to be wrapped in an object, this is easier than trying to special-case those (in other places we stringify arrays of runs or nested runs).

This seems overall safer

jacoblee93: See below

dqbd (approved):

jacoblee93:
Thanks!

PR #1005, opened 2024-09-15
fix(js): Removes langchain and @langchain/core from peer deps to fix npm installation

jacoblee93:
npm seems to have changed the way they handle circular deps - folks installing even old versions of `@langchain/core` that rely on a LangSmith version with the circular peer dep with npm are now getting:

```
Jacob:foo jacoblee$ npm i @langchain/core@0.1.29
npm ERR! Cannot read properties of null (reading 'edgesOut')
```

This works:

```
{
  "name": "foo",
  "version": "1.0.0",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "description": "",
  "dependencies": {
    "@langchain/core": "^0.1.29"
  },
  "overrides": {
    "langsmith": "0.1.18"
  }
}
```

This doesn't:

```
{
  "name": "foo",
  "version": "1.0.0",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "description": "",
  "dependencies": {
    "@langchain/core": "^0.1.29"
  },
  "overrides": {
    "langsmith": "0.1.55"
  }
}
```

@dqbd I think this will break bundlers so we shouldn't merge but maybe we can otherwise mark as external? Otherwise we will have to split out all non-LangChain related code into e.g. `@langchain/langsmith-core`, have `@langchain/core` depend on that, and have `langsmith` re-export all of `@langchain/langsmith-core` with extra entrypoints, which is not ideal.


dqbd (approved):
Will defer to @bracesproul w.r.t. `evaluation/langchain` entrypoint, but will trust @jacoblee93 that it's not referenced nor used anywhere else.

PR #1009, opened 2024-09-17
Fix SHORT_LIVED_TTL_SECONDS wrong comment on default value

abozaralizadeh:
According to the documentation, the default value of shortlived should be 1209600, which is 14 days. However, there is a typo in the comments, incorrectly describing it as 1 day.


hinthornw:
Thanks for your contribution!

hinthornw (approved):

PR #1010, opened 2024-09-17
[JS] Anno crud

hinthornw:
I need to add tests

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -3142,11 +3142,11 @@ export class Client {
     );
 
     let count = 0;
-    for await (const queue of this._getPaginated(
+    for await (const queues of this._getPaginated<AnnotationQueue>(
       "/annotation-queues",
       params
     )) {
-      yield queue as unknown as AnnotationQueue;
+      yield* queues;
       count++;
       if (limit !== undefined && count >= limit) break;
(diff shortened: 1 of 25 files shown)


js/src/client.ts
```diff
@@ -3103,6 +3105,217 @@ export class Client {
...
+  }
+
+  /**
+   * Read an annotation queue with the specified queue ID.
+   * @param queueId - The ID of the annotation queue to read
+   * @returns The AnnotationQueue object
+   */
+  public async readAnnotationQueue(queueId: string): Promise<AnnotationQueue> {
+    // TODO: Replace when actual endpoint is added
+    const queue = await this.listAnnotationQueues({
```
jacoblee93, line 3195: Naming is a bit funny here - `getAnnotationQueue` maybe?

`readAnnotationQueue` sounds like it should just return the items in the queue.

jacoblee93 (commented):
Needs tests but looks fine overall

js/src/client.ts
hinthornw: ya.... I think for sake of consistency I'm gonna keep it though since we use `read` elsewhere

agola11 (approved):

PR #1046, opened 2024-09-26
Handle missing version errors

hinthornw:


python/langsmith/client.py
```diff
@@ -5269,9 +5269,15 @@ def pull_prompt_commit(
...
-        )
+        try:
+            use_optimization = ls_utils.is_version_greater_or_equal(
+                self.info.version, "0.5.23"
+            )
+        except ValueError:
+            logger.exception(
+                "Failed to parse LangSmith API version. Defaulting to using optimization."
+            )
+            use_optimization = True
```
hinthornw, line 5280: Gonna check with @langchain-infra on this one. I think we have very few deployments out < 0.5.23, meaning that more positive EV from defualting to optimized

langchain-infra (approved):

langchain-infra (approved):

PR #1054, opened 2024-10-01
Nc/1oct/py prepare multipart

nfcampos:
- Add requests-toolbelt dep
- Use custom http adapter
  - subclass so that we can override blocksize (default one too small to maximize performance on modern network stacks)
  - override default pool size to match the max nr of batch tarcing threads (otherwise some threads would be left waiting for connections to be available)
  - actually mount the adapter (previous behavior was never mounting the adapter because sessions always have http and https adapters)
  - note before retry policy was being ignored, and no transport level retries were being done
- Add attachments field to run dict/model
  - update _run_transform to collect/drop attachments
  - warn if attachments are passed to an endpoint that can't accept them, in this case inject runs without attachments


python/langsmith/client.py
```diff
@@ -1490,6 +1546,178 @@ def _post_batch_ingest_runs(self, body: bytes, *, _context: str):
...
+        if not pre_sampled:
+            create_dicts = self._filter_for_sampling(create_dicts)
+            update_dicts = self._filter_for_sampling(update_dicts, patch=True)
+        if not create_dicts and not update_dicts:
+            return
+        # insert runtime environment
+        self._insert_runtime_env(create_dicts)
+        self._insert_runtime_env(update_dicts)
+        # check size limit
+        size_limit_bytes = (self.info.batch_ingest_config or {}).get(
```
hinthornw, line 1636: Should we also have an `EMPTY_DICT`? (or just a frozen dict-like) 

```
from types import MappingProxyType

EMPTY_DICT = MappingProxyType({})
```

python/langsmith/client.py
```diff
@@ -5590,6 +5825,9 @@ def _tracing_control_thread_func(client_ref: weakref.ref[Client]) -> None:
     size_limit: int = batch_ingest_config["size_limit"]
     scale_up_nthreads_limit: int = batch_ingest_config["scale_up_nthreads_limit"]
     scale_up_qsize_trigger: int = batch_ingest_config["scale_up_qsize_trigger"]
+    use_multipart = os.getenv("LANGSMITH_FF_MULTIPART") in ["1", "true"]
```
hinthornw, line 5828: Should we cache this? Probably not terribly important since we create the control thread once per client and it's in the background thread, etc.

python/langsmith/client.py
```diff
@@ -781,7 +821,7 @@ def request_with_retries(
             ls_utils.FilterPoolFullWarning(host=str(self._host)),
         ]
         retry_on_: Tuple[Type[BaseException], ...] = (
-            *(retry_on or []),
+            *(retry_on or ()),
```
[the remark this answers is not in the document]
nfcampos: I don't think this one works, because its the wrong type

python/langsmith/client.py
nfcampos: I think fine

hinthornw (approved):

python/langsmith/client.py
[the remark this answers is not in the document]
hinthornw: okok cool

PR #1065, opened 2024-10-05
[Py] Token Counting

hinthornw:


python/langsmith/wrappers/_openai.py
```diff
@@ -160,6 +160,35 @@ def _reduce_completions(all_chunks: List[Completion]) -> dict:
...
+        for k, v in (metadata.get("completion_tokens_details") or _DICT).items()
+    }
+    return metadata
+
+
+def _count_beans(outputs: Any):
+    try:
+        rdict = outputs.model_dump()
+        rdict["usage_metadata"] = _flatten_metadata(rdict.pop("usage", None))
+        return rdict
```
agola11, line 186: just flagging that this will have the potentially unintended consequence of making the output a dict instead of a `ChatCompletion` object -- the output will be inlined in `outputs` as opposed to nested under `outputs.output`

hinthornw: Just in the run.outputs right? (not in the actual application response)

That's what we do in the backend anyway, right?

PR #1077, opened 2024-10-08
Add optional crash dumping on ConnectionError

nhuang-lc:
Crash dump on ConnectionError if a global variable is provided.

The goal is to see inside of a request without logging sensitive fields

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "langsmith",
-  "version": "0.1.64",
+  "version": "0.1.65",
   "description": "Client library to connect to the LangSmith LLM Tracing and Evaluation Platform.",
   "packageManager": "yarn@1.22.19",
   "files": [
--- a/js/src/index.ts
+++ b/js/src/index.ts
@@ -14,4 +14,4 @@ export { RunTree, type RunTreeConfig } from "./run_trees.js";
 export { overrideFetchImplementation } from "./singletons/fetch.js";
 
 // Update using yarn bump-version
-export const __version__ = "0.1.64";
+export const __version__ = "0.1.65";
--- a/js/src/run_trees.ts
+++ b/js/src/run_trees.ts
@@ -8,6 +8,7 @@ import {
 import { Client } from "./client.js";
 import { isTracingEnabled } from "./env.js";
 import { warnOnce } from "./utils/warn.js";
+import { _LC_CONTEXT_VARIABLES_KEY } from "./singletons/constants.js";
 
 function stripNonAlphanumeric(input: string) {
   return input.replace(/[-:.]/g, "");
@@ -172,7 +173,12 @@ export class RunTree implements BaseRun {
   execution_order: number;
   child_execution_order: number;
 
-  constructor(originalConfig: RunTreeConfig) {
+  constructor(originalConfig: RunTreeConfig | RunTree) {
+    // If you pass in a run tree directly, return a shallow clone
+    if (isRunTree(originalConfig)) {
+      Object.assign(this, { ...originalConfig });
+      return;
+    }
     const defaultConfig = RunTree.getDefaultConfig();
     const { metadata, ...config } = originalConfig;
     const client = config.client ?? RunTree.getSharedClient();
@@ -247,6 +253,13 @@ export class RunTree implements BaseRun {
       execution_order: child_execution_order,
       child_execution_order: child_execution_order,
     });
+
+    // Copy context vars over into the new run tree.
+    if (_LC_CONTEXT_VARIABLES_KEY in this) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (child as any)[_LC_CONTEXT_VARIABLES_KEY] =
+        this[_LC_CONTEXT_VARIABLES_KEY];
+    }
 
     type ExtraWithSymbol = Record<string | symbol, unknown>;
     const LC_CHILD = Symbol.for("lc:child_config");
--- a/js/src/singletons/constants.ts
+++ b/js/src/singletons/constants.ts
@@ -1,0 +1,1 @@
+export const _LC_CONTEXT_VARIABLES_KEY = Symbol.for("lc:context_variables");
--- a/js/src/tests/traceable.test.ts
+++ b/js/src/tests/traceable.test.ts
@@ -1,9 +1,11 @@
 import { jest } from "@jest/globals";
 import { RunTree, RunTreeConfig } from "../run_trees.js";
+import { _LC_CONTEXT_VARIABLES_KEY } from "../singletons/constants.js";
 import { ROOT, traceable, withRunTree } from "../traceable.js";
 import { getAssumedTreeFromCalls } from "./utils/tree.js";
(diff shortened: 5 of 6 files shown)


python/langsmith/client.py
```diff
@@ -742,6 +746,14 @@ def _content_above_size(self, content_length: Optional[int]) -> Optional[str]:
         if size_limit is None:
             return None
         if content_length > size_limit:
+            should_debug_crash_dump = os.getenv("LANGSMITH_DEBUG_CRASH_DUMP") in [
```
hinthornw, line 749: Let's use `get_env_var` in utils to cache this

hinthornw: And define `["1", "true"]` as a tuple and define in outer context

python/langsmith/client.py
```diff
@@ -742,6 +746,14 @@ def _content_above_size(self, content_length: Optional[int]) -> Optional[str]:
...
+                "1",
+                "true",
+            ]
+            if should_debug_crash_dump and request is not None:
+                with open("content_size_limit_crash_dump.jsonl", "a") as f:
```
hinthornw, line 754: need to specify `'utf-8'` as the encoding.
Also, let's use gzip - by definition these will get pretty big pretty fast

python/langsmith/client.py
```diff
@@ -742,6 +746,14 @@ def _content_above_size(self, content_length: Optional[int]) -> Optional[str]:
         if size_limit is None:
             return None
         if content_length > size_limit:
+            should_debug_crash_dump = os.getenv("LANGSMITH_DEBUG_CRASH_DUMP") in [
+                "1",
+                "true",
+            ]
+            if should_debug_crash_dump and request is not None:
```
hinthornw, line 753: This doesn't belong in a function that checks content length - it should go outside as the next step.

Also I think we would benefit from seeing other things in the dump even if the payload size doesn't surpass (since these are also of interest)

python/langsmith/client.py
nhuang-lc: Yea I like that!

python/langsmith/client.py
nhuang-lc: Do you think we should always dump? Had intended this to only dump if we exceeded the content length limit (and also potentially dump here if other errors occur)

hinthornw: Only on error, but this is intended for debugging connection errors, correct?

nhuang-lc: Ah yes it's a connection error but specifically the one triggered by the content length limit.

> Failed to batch ingest runs: langsmith.utils.LangSmithConnectionError: Connection error caused failure to POST https://api.smith.langchain.com/runs/batch in LangSmith API. The content length of 270646872 bytes exceeds the maximum size limit of 20971520 bytes.

Their input and output tokens and traces seem to be pretty small, so we wanted to see what else was in their payload which was causing them to exceed 20mb.

We can also log more generically on ConnectionError though if you think that's better! It does make sense to log for other types of connection errors too

python/langsmith/client.py
```diff
@@ -931,6 +933,24 @@ def request_with_retries(
...
+                        "LANGSMITH_DEBUG_CRASH_DUMP"
+                    )
+                    if debug_crash_dump_file is not None and e.request is not None:
+                        try:
+                            request_data: (
+                                requests.Request | requests.PreparedRequest
+                            ) = copy.deepcopy(e.request)
+                            if "x-api-key" in request_data.headers:
+                                request_data.headers["x-api-key"] = masked_api_key
+                            with gzip.open(debug_crash_dump_file, "ab") as f:
```
hinthornw, line 946:
```suggestion
                            with gzip.open(debug_crash_dump_file, "ab", encoding="utf-8") as f:
```

nhuang-lc: Ah sorry I lost this

python/langsmith/client.py
```diff
@@ -931,6 +933,24 @@ def request_with_retries(
...
                     masked_api_key = f"{prefix}{filler}{suffix}"
 
+                    debug_crash_dump_file = ls_utils.get_env_var(
+                        "LANGSMITH_DEBUG_CRASH_DUMP"
+                    )
+                    if debug_crash_dump_file is not None and e.request is not None:
+                        try:
+                            request_data: (
+                                requests.Request | requests.PreparedRequest
+                            ) = copy.deepcopy(e.request)
```
hinthornw, line 943: We should avoid deep copies. Especially just so we can put a masked API key in the headers

python/langsmith/client.py
nhuang-lc: Actually quick question here @hinthornw 

Previously I was opening the files in byte mode "ab", and then directly writing in byte mode with .encode("utf-8") on my json.dumps.

With the above change, I need to open the file in text mode "at", and then just write strings back. Is this considered safer style?

python/langsmith/client.py
```diff
@@ -931,6 +933,43 @@ def request_with_retries(
                     filler = "*" * (max(0, len(api_key) - 7))
                     masked_api_key = f"{prefix}{filler}{suffix}"
 
+                    debug_crash_dump_file = ls_utils.get_env_var(
+                        "LANGSMITH_DEBUG_CRASH_DUMP"
+                    )
+                    if debug_crash_dump_file is not None and e.request is not None:
+                        try:
+                            headers = e.request.headers.copy()
```
nhuang-lc, line 941: @hinthornw Let me know if you agree with this approach.

Made a shallow copy of the headers so we could get rid of the x-api-key, but log the rest of the headers.

Manually added all of the fields which I think are relevant here. This is to prevent us accidentally logging other sensitive fields (if any).

python/langsmith/client.py
```diff
@@ -931,6 +932,43 @@ def request_with_retries(
...
+                                    if isinstance(e.request, requests.Request)
+                                    else None
+                                ),
+                                "json": (
+                                    e.request.json
+                                    if isinstance(e.request, requests.Request)
+                                    else None
+                                ),
+                            }
+                            with gzip.open(
```
nfcampos, line 962: can you easily open a gzip file that has been appended to multiple times?

python/langsmith/client.py
```diff
@@ -931,6 +932,43 @@ def request_with_retries(
                     filler = "*" * (max(0, len(api_key) - 7))
                     masked_api_key = f"{prefix}{filler}{suffix}"
 
+                    debug_crash_dump_file = ls_utils.get_env_var(
```
nfcampos, line 935: can we read env var once in global scope and then just use that here (reading env vars is expensive)

python/langsmith/client.py
nhuang-lc: Just looked into it a bit, it is a bit of a pain to read from multiple streams.

Do you think an acceptable alternative is to write to a file with a certain timestamp appended to the file path?

We can probably turn the env variable back to a boolean, and then use a fixed static name like `debug_crash_dump_{timestamp}`. wdyt? @nfcampos

nhuang-lc:
No longer necessary due to multi-part endpoint

PR #1079, opened 2024-10-08
track evaluator errors from sdk

isahers1:


python/langsmith/evaluation/_runner.py
```diff
@@ -82,6 +84,27 @@
     ],
 ]
 
+def extract_code_evaluator_feedback_keys(python_code: str) -> list[str]:
```
hinthornw, line 87: If we commit to doing this, we'll need to handle case where you return `EvaluationResult` as well

isahers1: thanks for calling this out. Tried to handle all the cases I could think of:

- return {"key":"foo","score":1} -> should extract ["foo"]
- return {"score":1} -> should extract name of function
- return EvaluationResult(...) should extract key from object
- return EvaluationResults(results=...) should extract keys from all the EvaluationResults objects passed to 'results'. This case is harder, but it works for simple invocations.

PR #1085, opened 2024-10-10
Allow users to add metadata on end()

nhuang-lc:


agola11 (commented):
I would add an integration test in https://github.com/langchain-ai/langsmith-sdk/blob/main/python/tests/integration_tests/test_runs.py (it actually seems like we don't have any for RunTree explicitly there :/ )

Basically, just create a run tree and log in a test, try to fetch the run and check to see that it has the metadata. 

Should also implement in JS

js/src/tests/run_trees.int.test.ts
```diff
@@ -214,3 +214,31 @@ test.concurrent(
   },
   120_000
 );
+
+test.concurrent(
+  "Test end() write to metadata",
+  async () => {
+    const runMeta = uuid.v4();
+    const projectName = `__test_end_metadata_run_tree_js ${runMeta}`;
```
agola11, line 222: If you're creating a project with UUID in the name, you should clean it up after the test. Also should consolidate how you're doing in py and JS. I would honestly just create a unique project and clean up at end of test in py and js

python/tests/integration_tests/test_runs.py
```diff
@@ -455,3 +455,28 @@ async def my_async_generator(num: int) -> AsyncGenerator[str, None]:
...
+
+    run_tree.end(metadata={"final_metadata": run_meta}, outputs={"result": "success"})
+    run_tree.post()
+
+    filter_ = (
+        f'and(eq(metadata_key, "final_metadata"), eq(metadata_value, "{run_meta}"))'
+    )
+    poll_runs_until_count(langchain_client, project_name, 1, filter_=filter_)
+
+    runs_ = list(langchain_client.list_runs(project_name=project_name, filter=filter_))
```
agola11, line 478: I would explicitly set the run_id in `RunTree`, then fetch the run by run id. That guarantees you're looking at the exact same run you just posted. If someone else were to write to this project, it might throw off your test

agola11 (approved):

PR #1093, opened 2024-10-11
feat(js): Adds further required support for context vars

jacoblee93:


js/src/run_trees.ts
```diff
@@ -28,6 +28,8 @@ export function convertToDottedOrderFormat(
   );
 }
 
+export const _LC_CONTEXT_VARIABLES_KEY = Symbol.for("lc:context_variables");
```
dqbd, line 31: _nit_: do we want to export this symbol out? Ideally would avoid exporting these internals out. If no-option is available, maybe `@hidden` jsdoc tag would help (saw that tag somewhere when reading LC.js)

dqbd (approved):

js/src/run_trees.ts
jacoblee93: Yeah can refactor

jacoblee93: Done

jacoblee93:
Thanks!

PR #1095, opened 2024-10-13
feat(JS): add `processInputs` / `processOutputs` to traceable + add usage metadata to `wrapOpenAI`

agola11:


js/src/traceable.ts
```diff
@@ -29,43 +29,64 @@ AsyncLocalStorageProviderSingleton.initializeGlobalInstance(
...
-    return rawOutputs;
+    outputs = rawOutputs;
+  } else {
+    outputs = { outputs: rawOutputs };
+  }
+
+  try {
+    return processOutputs(outputs);
+  } catch (e) {
+    console.error("Error occurred during processOutputs:", e);
```
dqbd, line 72: _nit_: would also add a explainer that raw outputs / inputs are being sent to LangSmith

js/src/traceable.ts
```diff
@@ -293,16 +314,35 @@ export function traceable<Func extends (...args: any[]) => any>(
     getInvocationParams?: (
       ...args: Parameters<Func>
     ) => InvocationParamsSchema | undefined;
+
+    /**
+     * Apply transformations to the inputs before logging.
```
dqbd, line 319: _nit_: would add a disclaimer that the `processInputs` is not inherited in inner `traceable` calls

jacoblee93 (approved):

PR #1106, opened 2024-10-17
feat(js): Use new multipart endpoint when available

jacoblee93:


js/src/utils/env.ts
```diff
@@ -69,7 +69,7 @@ export type RuntimeEnvironment = {
 
 let runtimeEnvironment: RuntimeEnvironment | undefined;
 
-export async function getRuntimeEnvironment(): Promise<RuntimeEnvironment> {
```
nfcampos, line 72: was this exported?

jacoblee93: Only internally

nfcampos (approved):

PR #1107, opened 2024-10-18
feat: cross platform tracing client written in rust

agola11:
todo:
- [x] add more tests for inputs/outputs verification
- [x] actually test integration with langsmith
- [x] attempt switching to mpmc pattern
- [x] stop excessive copying
- [ ] python binding
- [ ] gzip
- [x] add benchmarks 
- [x] stream multipart requests


rust/crates/langsmith_tracing_client/tests/test_run.rs
```diff
@@ -0,0 +1,115 @@
...
+    };
+
+    let run_create = RunCreate {
+        common: run_common,
+        name: String::from("Run Name"),
+        start_time: TimeValue::String("2024-10-16T12:00:00Z".to_string()),
+        end_time: Some(TimeValue::String("2024-10-16T14:00:00Z".to_string())),
+        outputs: serde_json::json!({"output_key": "output_value"}),
+        run_type: String::from("test_run_type"),
+        session_id: None,
```
nfcampos, line 46: session id and name should be in common

rust/crates/langsmith_tracing_client/src/client/tracing_client.rs
```diff
@@ -0,0 +1,70 @@
...
+use crate::client::processor::RunProcessor;
+use crate::client::run::QueuedRun;
+use crate::client::run::{RunCreateWithAttachments, RunUpdateWithAttachments};
+use std::time::Duration;
+use tokio::sync::mpsc::{self, Sender};
+use tokio::task::JoinHandle;
+
+pub struct ClientConfig {
+    pub endpoint: String,
+    pub queue_capacity: usize,
```
nfcampos, line 11: what's the idea with a limited capacity? if it goes over what do we expect should happen

PR #1117, opened 2024-10-23
benchmark create_run

efriis:
for baseline performance

`poetry run pytest bench/create_run.py --disable-socket -k "test_benchmark_runs[500-1000]" -s`


efriis:
going to do separately post module separation, plus include profiling

PR #1118, opened 2024-10-23
integration tests multipart

efriis:
adds integration tests to confirm correct handling for multipart content cases


python/langsmith/client.py
```diff
@@ -1740,6 +1740,8 @@ def _send_multipart_req(
                         logger.warning(f"Failed to multipart ingest runs: {exc_desc}")
                     except Exception:
                         logger.warning(f"Failed to multipart ingest runs: {repr(e)}")
+                    # do not retry by default
+                    return
```
efriis, line 1744: this fixes the issue where 422s were getting retried

python/langsmith/evaluation/_runner.py
efriis, line 1: `make format`

nfcampos (approved):

PR #1124, opened 2024-10-24
fix(js): Limit queue batch concurrency, bump timeout, add maximum wait period for serverless envs

### LINKED ISSUE #1101: Issue: Tracing fails on larger batches

Hello LangSmith team!

## Description

I'm running a small chain that takes an image and returns a transcription and description of the image. I'm doing a [batch operation](https://js.langchain.com/docs/how_to/lcel_cheatsheet/#batch-a-runnable), and LangSmith fails to save the trace for most of the runs. I get this error on ~50 out of 61 runs on a document I've been testing on:

```
DOMException [TimeoutError]: The operation was aborted due to timeout
    at node:internal/deps/undici/undici:12502:13
    at RetryOperation._fn ([REDACTED]/node_modules/.pnpm/p-retry@4.6.2/node_modules/p-retry/index.js:50:12) {
  attemptNumber: 7,
  retriesLeft: 0
}
```

It seems like the traces should all be batched and reported together, but that's not happening - I'm getting ~50 of these errors logged every time I run this batch. It's a terrible dev experience.

Additionally, the reported latency times in the LangSmith console are incorrect. It looks like it's reporting the diff from chain start time to the time it was received by the LangSmith server (it _should_ be from chain start to resolution, ignoring latency between my server and the LangSmith server). For the handful of successful traces getting through, I'm seeing reported times of over 100 seconds, while I can see from local logging that the entire batch takes between 45 and 65 seconds to run.

It should be noted that I tried wrapping the function containing the `.batch` call as a [traceable](https://docs.smith.langchain.com/old/tracing/faq/logging_and_viewing#using-traceable--traceable) in the hopes of coaxing LangSmith to report the whole trace in one batch, but I got this error:
```
Error: Failed to batch create run. Received status [403]: Forbidden. Server response: <!doctype html><meta charset="utf-8"><meta name=viewport content="width=device-width, initial-scale=1"><title>403</title>403 Forbidden
```
## My Setup

I can confirm my env variables are setup properly, as the automatic tracing functionality just works.  `LANGCHAIN_CALLBACKS_BACKGROUND` is set to true.

LangSmith TypeScript SDK version: `0.1.66`

## Potential solutions

- Ideally, calling `.batch` on a LangChain runnable would cause LangSmith to report a single trace.
- Alternatively, I'd like to be able to modify the timeout on LangSmith to see if I can find a configuration that works
- As a last resort, I need the ability to suppress failed trace reporting. These errors are making my console unusable.

jacoblee93:
Fixes #1101 

Bumps timeouts related to large uploads, limit the total amount of bandwidth used, and avoid blocking for too long in the default case

Adds `LANGSMITH_TRACING_BACKGROUND` to set blocking behavior

Bump to 0.2 will come along with this.


js/src/tests/batch_client.int.test.ts
```diff
@@ -241,3 +242,52 @@ test.concurrent(
...
 );
+
+test.skip("very large runs", async () => {
+  const langchainClient = new Client({
+    autoBatchTracing: true,
+    timeout_ms: 120_000,
+  });
+
+  // Use old batch endpoint
+  // jest
```
hinthornw, line 253: nit

js/src/client.ts
```diff
@@ -753,35 +765,44 @@ export class Client {
...
-        if (serverInfo?.batch_ingest_config?.use_multipart_endpoint) {
-          await this.multipartIngestRuns(ingestParams);
-        } else {
-          await this.batchIngestRuns(ingestParams);
+    const batchSizeLimit = await this._getBatchSizeLimitBytes();
+    while (this.autoBatchQueue.items.length > 0) {
+      for (let i = 0; i < this.traceBatchConcurrency; i++) {
+        const [batch, done] = this.autoBatchQueue.pop(batchSizeLimit);
+        if (!batch.length) {
+          done();
```
hinthornw, line 773: should we just let the processBatch handle this case?

hinthornw (approved):

js/src/client.ts
jacoblee93: Yeah could have but figured I wouldn't make assumptions

PR #1129, opened 2024-10-25
feat(py): Add multipart feedback ingestion

akira:
When multipart is enabled, use same tracing_queue and multipart endpoint to send feedback as well.

### DIFF
--- a/python/langsmith/_internal/_background_thread.py
+++ b/python/langsmith/_internal/_background_thread.py
@@ -69,9 +69,10 @@ def _tracing_thread_handle_batch(
 ) -> None:
     create = [it.item for it in batch if it.action == "create"]
     update = [it.item for it in batch if it.action == "update"]
+    feedback = [it.item for it in batch if it.action == "feedback"]
     try:
         if use_multipart:
-            client.multipart_ingest_runs(create=create, update=update, pre_sampled=True)
+            client.multipart_ingest(create=create, update=update, feedback=feedback, pre_sampled=True)
         else:
             client.batch_ingest_runs(create=create, update=update, pre_sampled=True)
     except Exception:
--- a/python/langsmith/_internal/_background_thread.py
+++ b/python/langsmith/_internal/_background_thread.py
@@ -72,7 +72,9 @@ def _tracing_thread_handle_batch(
     feedback = [it.item for it in batch if it.action == "feedback"]
     try:
         if use_multipart:
-            client.multipart_ingest(create=create, update=update, feedback=feedback, pre_sampled=True)
+            client.multipart_ingest(
+                create=create, update=update, feedback=feedback, pre_sampled=True
+            )
         else:
             client.batch_ingest_runs(create=create, update=update, pre_sampled=True)
     except Exception:
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -1471,7 +1471,7 @@ class Client:
             - The run objects MUST contain the dotted_order and trace_id fields
                 to be accepted by the API.
         """
-        if not create and not update and not feedback:
+        if not (create or update or feedback):
             return
         # transform and convert to dicts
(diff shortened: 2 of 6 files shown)


nfcampos (approved):

python/langsmith/client.py
```diff
@@ -4241,66 +4278,87 @@ def create_feedback(
...
+                    "/feedback",
+                    request_kwargs={
+                        "data": feedback_block,
+                    },
+                    stop_after_attempt=stop_after_attempt,
+                    retry_on=(ls_utils.LangSmithNotFoundError,),
+                )
+            return ls_schemas.Feedback(**feedback.dict())
+        except Exception as e:
+            logger.error("Error creating feedback", exc_info=True)
```
hinthornw, line 4360: What do we get from this if we re-raise anyway?

python/langsmith/client.py
```diff
@@ -4241,66 +4278,87 @@ def create_feedback(
...
+
+            feedback_block = _dumps_json(feedback.dict(exclude_none=True))
+            use_multipart = (self.info.batch_ingest_config or {}).get(
+                "use_multipart_endpoint", False
+            )
+
+            if (
+                use_multipart
+                and self.tracing_queue is not None
+                and feedback.trace_id is not None
```
hinthornw, line 4343: So we cannot log session-level feedback using the batch API then?

python/langsmith/client.py
```diff
@@ -4185,6 +4221,7 @@ def create_feedback(
         project_id: Optional[ID_TYPE] = None,
         comparative_experiment_id: Optional[ID_TYPE] = None,
         feedback_group_id: Optional[ID_TYPE] = None,
+        trace_id: Optional[ID_TYPE] = None,
```
hinthornw, line 4224: Need docstring

python/langsmith/schemas.py
```diff
@@ -440,6 +440,8 @@ class FeedbackBase(BaseModel):
     """The time the feedback was last modified."""
     run_id: Optional[UUID]
     """The associated run ID this feedback is logged for."""
+    trace_id: Optional[UUID]
```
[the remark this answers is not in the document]
akira: We require trace_id for the multipart processing for feedback - otherwise would need to do a bunch more processing when accepting them

python/langsmith/client.py
akira: I noticed this was swallowing exceptions (if running in executor)

python/langsmith/client.py
akira: Yeah, currently requires a trace_id so wouldn't work with that

python/langsmith/_internal/_background_thread.py
```diff
@@ -69,9 +69,12 @@ def _tracing_thread_handle_batch(
...
     update = [it.item for it in batch if it.action == "update"]
+    feedback = [it.item for it in batch if it.action == "feedback"]
     try:
         if use_multipart:
-            client.multipart_ingest_runs(create=create, update=update, pre_sampled=True)
+            client.multipart_ingest(
+                create=create, update=update, feedback=feedback, pre_sampled=True
+            )
         else:
             client.batch_ingest_runs(create=create, update=update, pre_sampled=True)
```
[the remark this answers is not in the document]
akira: synced offline - but this would be much slower in that case for evals since this background thread is a single thread vs evals has multiple threads and is actually faster for the single feedback case.

python/langsmith/client.py
```diff
@@ -1442,7 +1471,7 @@ def multipart_ingest_runs(
             - The run objects MUST contain the dotted_order and trace_id fields
                 to be accepted by the API.
         """
-        if not create and not update:
+        if not create and not update and not feedback:
```
agola11, line 1474: very nit: slightly more readable if `not(create or update or feedback)`

akira: Updated

PR #1136, opened 2024-10-26
serialize instead of deepcopy for performance gains

efriis:
code cleanup changes (no performance benefits):
- [x] declares helper methods for working with Multipart requests (_multipart.py)
- [x] declares explicit `Operation` dataclasses representing run/feedback operations in the `tracing_queue` (_operations.py)
- [x] declares helper methods for creating/using `Operation` dataclasses (_operations.py)
- [x] declare new `_batch_ingest_ops` and `_multipart_ingest_ops` methods that take `List[Operation]` as input and call respective endpoints (client.py)
- [x] convert background thread to batch process `Operation` objects in `_x_ingest_ops` functions (_background_thread.py)
- [x] convert `batch_ingest_runs` and `multipart_ingest_runs` into wrappers that create `Operation` objects and pass along to `_x_ingest_ops` functions (client.py)

performance change:
- [x] convert `create_run`, `update_run` to create an `Operation` dataclass instead of deepcopying the input dictionary for performance gains 💪 (client.py)

interface changes:
- `multipart_ingest_runs` no longer accepts feedback (not a public interface yet - shouldn't matter)
- `batch_ingest_runs`: no interface changes


python/langsmith/client.py
```diff
@@ -1212,18 +1215,34 @@ def create_run(
...
+            run_create = self._run_transform(
+                run_create,
+                copy=False,
+                attachments_collector=attachments_collector,
+            )
+            if revision_id is not None:
+                run_create["extra"]["metadata"]["revision_id"] = revision_id
+            acc = convert_to_multipart_parts_and_context(
+                [run_create], [], all_attachments=attachments_collector
+            )
```
efriis, line 1235: this is the "new stuff" in the hot path - except we have a massive removal by doing `_run_transform(copy=False)` instead of `copy=True`

python/langsmith/client.py
```diff
@@ -1212,18 +1215,34 @@ def create_run(
...
         ):
+            attachments_collector: Dict[str, ls_schemas.Attachments] = {}
+            run_create = self._run_transform(
+                run_create,
+                copy=False,
+                attachments_collector=attachments_collector,
+            )
+            if revision_id is not None:
+                run_create["extra"]["metadata"]["revision_id"] = revision_id
+            acc = convert_to_multipart_parts_and_context(
```
efriis, line 1233: need to insert runtime env above this I think

python/langsmith/client.py
```diff
@@ -1212,18 +1215,35 @@ def create_run(
...
-            run_create["extra"]["metadata"]["revision_id"] = revision_id
+
         if (
             self.tracing_queue is not None
             # batch ingest requires trace_id and dotted_order to be set
             and run_create.get("trace_id") is not None
             and run_create.get("dotted_order") is not None
         ):
+            attachments_collector: Dict[str, ls_schemas.Attachments] = {}
+            run_create = self._run_transform(
```
efriis, line 1226: need to confirm we're duplicating any shallow references we're setting things on - don't want to modify objects that are used upstream

alternatively, could add these fields during serialization

python/langsmith/_internal/_background_thread.py
```diff
@@ -36,8 +44,19 @@ class TracingQueueItem:
...
 
     priority: str
-    action: str
-    item: Any = field(compare=False)
+    item: Union[SerializedRunOperation, SerializedFeedbackOperation]
+
+    def __lt__(self, other: TracingQueueItem) -> bool:
+        return (self.priority, self.item.__class__) < (
+            other.priority,
+            other.item.__class__,
```
hinthornw, line 52: We're no longer comparing based on operation fwiw. I don't think that actually matters just noting.

python/langsmith/_internal/_multipart.py
```diff
@@ -0,0 +1,23 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Dict, Iterable, Tuple
+
+MultipartPart = Tuple[str, Tuple[None, bytes, str, Dict[str, str]]]
+
+
+@dataclass
+class MultipartPartsAndContext:
```
hinthornw, line 10: Do we need all these to be dataclasses? 

I'm sure this is dwarfed by the other improvements, but dataclass creation is much slower than dict/slotted class creation. So while we're on the topic of perf:
```shell
(prompt-optimizer) ➜  langchainplus-sdk git:(main) ✗ python foo.py d # dataclass
Time taken to create 10000000 instances: 6.2248 seconds
(prompt-optimizer) ➜  langchainplus-sdk git:(main) ✗ python foo.py t # typed dict
Time taken to create 10000000 instances: 1.9094 seconds
(prompt-optimizer) ➜  langchainplus-sdk git:(main) ✗ python foo.py slt # regular class with __slots__
Time taken to create 10000000 instances: 1.0796 seconds
(prompt-optimizer) ➜  langchainplus-sdk git:(main) ✗ python foo.py std # stdlib dict
Time taken to create 10000000 instances: 1.0785 seconds
```

python/langsmith/_internal/_operations.py
```diff
@@ -0,0 +1,199 @@
...
+
+import orjson
+
+from langsmith import schemas as ls_schemas
+from langsmith._internal._multipart import MultipartPart, MultipartPartsAndContext
+from langsmith._internal._serde import dumps_json as _dumps_json
+
+
+@dataclass
+class SerializedRunOperation:
```
hinthornw, line 16: Do we actually need these to be dataclasses?

python/langsmith/client.py
```diff
@@ -1325,22 +1367,13 @@ def batch_ingest_runs(
         if not create and not update:
             return
         # transform and convert to dicts
-        create_dicts = [self._run_transform(run) for run in create or EMPTY_SEQ]
+        create_dicts = [
+            self._run_transform(run, copy=True) for run in create or EMPTY_SEQ
+        ]  # still copy create dicts because manipulated in create_by_id
```
hinthornw, line 1372: Is this just fixing a bug (since we hadn't been copying these before)

python/langsmith/client.py
```diff
@@ -1444,7 +1458,6 @@ def multipart_ingest(
         update: Optional[
             Sequence[Union[ls_schemas.Run, ls_schemas.RunLikeDict, Dict]]
         ] = None,
```
hinthornw, line 1460: Why remove?

python/langsmith/_internal/_background_thread.py
```diff
@@ -18,14 +20,20 @@
...
 
 if TYPE_CHECKING:
     from langsmith.client import Client
 
 logger = logging.getLogger("langsmith.client")
 
 
-@dataclass(order=True)
+@functools.total_ordering
+@dataclass
```
hinthornw, line 36: I bet we could just do a normal class

python/langsmith/_internal/_operations.py
```diff
@@ -0,0 +1,199 @@
...
+        trace_id=payload["trace_id"],
+        _none=_dumps_json(payload),
+        inputs=_dumps_json(inputs) if inputs is not None else None,
+        outputs=_dumps_json(outputs) if outputs is not None else None,
+        events=_dumps_json(events) if events is not None else None,
+        attachments=attachments if attachments is not None else None,
+    )
+
+
+def combine_serialized_run_operations(
```
hinthornw, line 80: nit: if this is also handling feedback making `combine_serialized_queue_operations`?

python/langsmith/client.py
```diff
@@ -1288,6 +1262,74 @@ def _hide_run_outputs(self, outputs: dict):
             return outputs
         return self._hide_outputs(outputs)
 
+    def _batch_ingest_ops(
```
hinthornw, line 1265:
```suggestion
    def _batch_ingest_run_ops(
```

If this only actually is intended to handle runs

python/langsmith/client.py
```diff
@@ -1352,64 +1385,26 @@ def batch_ingest_runs(
...
-                "patch": self._filter_for_sampling(update_dicts, patch=True),
-            }
-        if not raw_body["post"] and not raw_body["patch"]:
-            return
-
-        self._insert_runtime_env(raw_body["post"] + raw_body["patch"])
-        info = self.info
+        if not pre_sampled:
+            create_dicts = self._filter_for_sampling(create_dicts)
+            update_dicts = self._filter_for_sampling(update_dicts, patch=True)
```
hinthornw, line 1390: Could check & end early here

python/langsmith/client.py
```diff
@@ -1288,6 +1262,74 @@ def _hide_run_outputs(self, outputs: dict):
...
+                if op.attachments:
+                    logger.warning(
+                        "Attachments are not supported in non-multipart mode"
+                    )
+                ids_and_partial_body[op.operation].append(
+                    (f"trace={op.trace_id},id={op.id}", orjson.dumps(curr_dict))
+                )
+            elif isinstance(op, SerializedFeedbackOperation):
+                logger.warning(
+                    "Feedback operations are not supported in non-multipart mode"
```
hinthornw, line 1295: `s/in non-multipart mode/use_multipart_endpoint is False/g`

hinthornw (commented):
Nice i like it. Looks pretty good to me - we should get this in with a release cut to test out in a few projects for now

python/langsmith/_internal/_multipart.py
efriis: interesting. relative to the other things I think this might be worth the overhead vs. maintaining init/slot stuff. Could also make these immutable at some point if we want

in the benchmark, we spend 99.8% of the `serialize_run_dict` time on `_dumps_json` and creating these classes is part of the remainder
ScreenShot 2024-10-31 at 12 55 27PM

python/langsmith/_internal/_operations.py
efriis: discuss in other one

python/langsmith/_internal/_operations.py
efriis: done

python/langsmith/client.py
efriis: great catch removing. Ended up replacing with serialization path

python/langsmith/client.py
efriis: can add back if we want this part of public interface. removes some tech debt. basically just keeping this function around for backwards compat with tests to make sure they were working - would have removed otherwise

python/langsmith/_internal/_multipart.py
hinthornw: fair fair

python/langsmith/_internal/_background_thread.py
efriis: yep should be fine

python/langsmith/_internal/_background_thread.py
```diff
@@ -67,16 +94,20 @@ def _tracing_thread_handle_batch(
...
-                create=create, update=update, feedback=feedback, pre_sampled=True
-            )
+            client._multipart_ingest_ops(ops)
         else:
-            client.batch_ingest_runs(create=create, update=update, pre_sampled=True)
+            if any(isinstance(op, SerializedFeedbackOperation) for op in ops):
+                logger.warn(
+                    "Feedback operations are not supported in non-multipart mode"
+                )
+                ops = [
```
hinthornw, line 106: silly but you could do this first and then compare lenghts if you only wanted to go over once

hinthornw (approved):

PR #1141, opened 2024-10-28
add ability to run experiments in parallel with different hyper params

isahers1:


python/langsmith/evaluation/_runner.py
```diff
@@ -870,48 +878,134 @@ def _evaluate(
...
-        runs,
-        client,
-    )
+    if hyper_params is None:
+        runs = None if _is_callable(target) else cast(Iterable[schemas.Run], target)
+        experiment_, runs = _resolve_experiment(
+            experiment,
+            runs,
+            client,
+        )
```
isahers1, line 891: @hinthornw not sure about this part, feel like we might need more checks and also corresponding ones inside of `aevaluate`

python/Makefile
```diff
@@ -36,7 +36,7 @@ doctest:
 	poetry run python -m pytest -n auto --durations=10 --doctest-modules langsmith
 
 evals:
-	poetry run python -m pytest tests/evaluation
+	poetry run python -m pytest tests/evaluation/test_evaluation.py::test_aevaluate_with_hyper_params
```
akira, line 39: typo?

isahers1: yup, fixed

PR #1143, opened 2024-10-28
integration tests against staging

efriis:


python/tests/integration_tests/test_client.py
```diff
@@ -356,11 +356,9 @@ def test_persist_update_run(langchain_client: Client) -> None:
 
 
 @pytest.mark.parametrize("uri", ["http://localhost:1981", "http://api.langchain.minus"])
-def test_error_surfaced_invalid_uri(monkeypatch: pytest.MonkeyPatch, uri: str) -> None:
+def test_error_surfaced_invalid_uri(uri: str) -> None:
     get_env_var.cache_clear()
-    monkeypatch.setenv("LANGCHAIN_ENDPOINT", uri)
```
efriis, line 361: these were impacting the value in other tests because env is shared

PR #1158, opened 2024-10-30
[python] Remove optional session calls for tenant_id

akira:


hinthornw (approved):
FYI This will make the .url properties likely not work  in the returned resources but that's mainly a datascience nicety

PR #1160, opened 2024-10-31
infra: use benchmark-fast in pybench ci

efriis:


.github/workflows/py-bench.yml
```diff
@@ -43,7 +43,7 @@ jobs:
         run: |
           {
             echo 'OUTPUT<<EOF'
-            make -s benchmark
+            make -s benchmark-fast
```
hinthornw, line 46: Maybe make configurable & run full one on cron?

efriis: this only runs on pull requests, not cron

let's do cron separate if we want it. would want that one hooked up to some sort of slack notify probably

PR #1167, opened 2024-11-04
js[patch]: make evaluate,traceable importable from top-level

baskaryan:
- match python sdk
- make our most-used classes and functions more easily accessible


hinthornw:
@dqbd , any concerns around import paths here?

dqbd:
pretty sure that moving evaluate and traceable into root will cause bundling issues for browser environments

hinthornw:
Gotcha ya then maybe we can't land

PR #1168, opened 2024-11-04
fix(vercel): add more debug logs, fix tracing detection for vercel, bump to 0.2.4

dqbd:
https://github.com/langchain-ai/langsmith-docs/pull/506

### DIFF
--- a/js/src/tests/vercel.test.ts
+++ b/js/src/tests/vercel.test.ts
@@ -134,6 +134,7 @@ test("generateText", async () => {
       }),
     },
     experimental_telemetry: AISDKExporter.getSettings({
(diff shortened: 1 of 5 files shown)


js/src/tests/vercel.test.ts
```diff
@@ -134,6 +134,7 @@ test("generateText", async () => {
       }),
     },
     experimental_telemetry: AISDKExporter.getSettings({
+      isEnabled: true,
```
jacoblee93, line 137: nit: generally prefer default as `false`, so `disabled: false` as the default

jacoblee93 (commented):
See comment

PR #1169, opened 2024-11-04
js feat: Support attachments in traceable

hinthornw:
Also add warning & handling of invalid attachments.

Proposed API:

getAttachments(args) => [attachments, otherArgs];

Returning the filtered args means that you're less likely to accidentally double-log attachments in inputs.


js/src/tests/traceable.int.test.ts
```diff
@@ -651,9 +651,105 @@ test.concurrent(
...
-    await waitUntilRunFound(langchainClient, runId);
-    const storedRun3 = await langchainClient.readRun(runId);
+    await waitUntilRunFound(langsmithClient, runId);
+    const storedRun3 = await langsmithClient.readRun(runId);
     expect(storedRun3.id).toEqual(runId);
     expect(storedRun3.status).toEqual("error");
   }
 );
+
+test.only("Test upload attachments and process inputs.", async () => {
```
dqbd, line 661: Remove the scoped test

js/src/schemas.ts
```diff
@@ -587,3 +593,5 @@ export type UsageMetadata = {
    */
   output_token_details?: OutputTokenDetails;
 };
+
+export type Attachments = Record<string, [string, Uint8Array]>;
```
dqbd, line 597: _question_: do we want to support ArrayBuffer as well? Might save a unnecessary array copy

js/src/tests/traceable.int.test.ts
hinthornw: oops

js/src/schemas.ts
hinthornw: Oh - I think that makes sense? Will have to look into it

hinthornw: Added!

js/src/traceable.ts
```diff
@@ -309,6 +338,15 @@ export function traceable<Func extends (...args: any[]) => any>(
     argsConfigPath?: [number] | [number, string];
     __finalTracedIteratorKey?: string;
 
+    /**
+     * Specify which inputs should be treated as attachments and return the remaining arguments.
+     * @param args Arguments of the traced function
+     * @returns A tuple containing the Attachments object and the remaining arguments
+     */
+    getAttachments?: (
```
agola11, line 346: calling this `getAttachments` makes it seem like this should return _just_ the attachments? `separateAttachments` or something similar would be better IMO

agola11 (approved):

PR #1170, opened 2024-11-04
[Python] to_pandas() method for ExperimentResults

baskaryan:
Screenshot 2024-11-04 at 2 56 09 PM

### DIFF
--- a/python/langsmith/evaluation/_runner.py
+++ b/python/langsmith/evaluation/_runner.py
@@ -480,7 +480,7 @@ class ExperimentResults:
             ]
             return pd.DataFrame(head)._repr_html_()  # type: ignore[operator]
         else:
-            return f"<ExperimentResults {self.experiment_name}>"
+            return self.__repr__()
 
     def __repr__(self) -> str:
         return f"<ExperimentResults {self.experiment_name}>"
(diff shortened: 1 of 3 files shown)


python/langsmith/evaluation/_runner.py
```diff
@@ -452,6 +452,36 @@ def _process_data(self) -> None:
...
+                    **{
+                        r.key: r.score if r.score is not None else r.value
+                        for r in x["evaluation_results"]["results"]
+                    },
+                }
+                for x in self._results[:5]
+            ]
+            return pd.DataFrame(head)._repr_html_()  # type: ignore[operator]
+        else:
+            return f"<ExperimentResults {self.experiment_name}>"
```
hinthornw, line 483:
```suggestion
            return self.__repr__()
```

python/langsmith/evaluation/_runner.py
```diff
@@ -452,6 +452,36 @@ def _process_data(self) -> None:
     def __len__(self) -> int:
         return len(self._results)
 
+    def _repr_html_(self) -> str:
+        import importlib.util
+
+        if self._results and importlib.util.find_spec("pandas"):
```
hinthornw, line 458: Can we make a `to_pandas` method and then re-use that here?

Would be useful in general

hinthornw (approved):

PR #1184, opened 2024-11-06
Remove initial timeout, the batch timeout is now always 250ms

nhuang-lc:
Standardize with the Python SDK, the delay for batch flushing in multipart JS should also be 250ms

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -466,9 +466,7 @@ export class Client {
 
   private autoBatchTimeout: ReturnType<typeof setTimeout> | undefined;
 
-  private autoBatchInitialDelayMs = 250;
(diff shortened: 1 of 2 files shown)


js/src/client.ts
```diff
@@ -814,15 +811,10 @@ export class Client {
       this.drainAutoBatchQueue(sizeLimitBytes);
     }
     if (this.autoBatchQueue.items.length > 0) {
```
jacoblee93, line 813: We could clean up this logic more, but let's minimize changes for now

jacoblee93 (approved):

jacoblee93:
Thank you!

PR #1185, opened 2024-11-06
fix: trace LLMEvaluator runs

isahers1:


python/langsmith/evaluation/llm_evaluator.py
```diff
@@ -278,15 +332,23 @@ def _prepare_variables(self, run: Run, example: Optional[Example]) -> dict:
 
     def _parse_output(self, output: dict) -> Union[EvaluationResult, EvaluationResults]:
         """Parse the model output into an evaluation result."""
+        run_info = output.get(RUN_KEY)
+        source_run_id = run_info.run_id if run_info is not None else None
         if isinstance(self.score_config, CategoricalScoreConfig):
             value = output["score"]
-            explanation = output.get("explanation", None)
+            explanation = output.get(self.reasoning_key, None)
```
hinthornw, line 339: I think we should handle if any of (self.reasoning_key, "reasoning", "explanation", "comment") are provided

python/langsmith/evaluation/llm_evaluator.py
```diff
@@ -1,14 +1,35 @@
 """Contains the LLMEvaluator class for building LLM-as-a-judge evaluators."""
 
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union, cast
+from uuid import UUID
 
+from langchain.schema import RUN_KEY
```
hinthornw, line 6: This isn't relevant here

python/langsmith/evaluation/llm_evaluator.py
```diff
@@ -1,14 +1,35 @@
...
+from langchain_core.outputs import RunInfo
 from pydantic import BaseModel
 
 from langsmith._internal._beta_decorator import warn_beta
 from langsmith.evaluation import EvaluationResult, EvaluationResults, RunEvaluator
+from langsmith.run_helpers import traceable
 from langsmith.schemas import Example, Run
 
 
+class RunIdCapturingHandler(BaseCallbackHandler):
```
hinthornw, line 18: Synced offline but we can just generate our own:
```
import uuid

from langchain_openai import ChatOpenAI

response = ChatOpenAI(model="gpt-4o").invoke(
    "What is the capital of France?", config={"run_id": (uuid.uuid4()}
)

print(response.content)
```

python/langsmith/evaluation/llm_evaluator.py
```diff
@@ -278,15 +332,23 @@ def _prepare_variables(self, run: Run, example: Optional[Example]) -> dict:
 
     def _parse_output(self, output: dict) -> Union[EvaluationResult, EvaluationResults]:
```
hinthornw, line 333: I'd just put the source_run_id as an argument here since it's a private method (unles this is being used elsewhere for some reason)

python/langsmith/evaluation/llm_evaluator.py
isahers1: am slightly confused on how something other than self.score_config.reasoning_key would exist. In the `_create_score_json_schema` don't we only set score_config.reasoning_key? I am probably just missing something simple

python/langsmith/evaluation/llm_evaluator.py
```diff
@@ -10,32 +11,66 @@
...
+    """
 
     key: str
     min: float = 0
     max: float = 1
     description: str
-    include_explanation: bool = False
-    explanation_description: Optional[str] = None
+    reasoning_key: Optional[str] = None
+    reasoning_description: Optional[str] = None
```
hinthornw, line 54: This is a breaking change, so we'll need to be able to handle accepting the other value in the __init__ and log a deprecation warning

python/langsmith/evaluation/llm_evaluator.py
```diff
@@ -10,32 +11,66 @@
...
+                else score_config.reasoning_description
+            ),
+        }
+
     if isinstance(score_config, CategoricalScoreConfig):
-        properties["score"] = {
+        properties["value"] = {
             "type": "string",
             "enum": score_config.choices,
             "description": f"The score for the evaluation, one of "
```
hinthornw, line 76: It isn't a score anymore, is it? It's the selected category.
The descriptions here matter a lot

python/langsmith/evaluation/llm_evaluator.py
```diff
@@ -52,23 +87,24 @@ def _create_score_json_schema(
...
         "title": score_config.key,
         "description": score_config.description,
         "type": "object",
         "properties": properties,
         "required": (
-            ["score", "explanation"] if score_config.include_explanation else ["score"]
+            [
+                (
+                    "value"
+                    if isinstance(score_config, CategoricalScoreConfig)
```
hinthornw, line 99: nit: easier to read if we just separate this out above (since nested python ternary expressions are a bit daunting)

python/langsmith/evaluation/llm_evaluator.py
```diff
@@ -206,18 +241,27 @@ def evaluate_run(
...
+        source_run_id = uuid4()
         variables = self._prepare_variables(run, example)
-        output: dict = cast(dict, await self.runnable.ainvoke(variables))
-        return self._parse_output(output)
+        output: dict = cast(
+            dict,
+            await self.runnable.ainvoke(variables, config={"run_id": source_run_id}),
+        )
+
+        return self._parse_output(output, str(source_run_id))
```
hinthornw, line 264: nit: Could leave as a UUID

python/tests/integration_tests/test_llm_evaluator.py
```diff
@@ -15,7 +15,7 @@ def test_llm_evaluator_init() -> None:
             key="vagueness",
             choices=["Y", "N"],
             description="Whether the response is vague. Y for yes, N for no.",
-            include_explanation=True,
```
hinthornw, line 18: Should keep a couple of these tests with include_explanation around for backward compat testing

hinthornw (commented):
Nice! Just a few nits left hten should be good

python/langsmith/evaluation/llm_evaluator.py
```diff
@@ -10,32 +11,66 @@
...
-    """Configuration for a categorical score."""
+    """Configuration for a categorical score.
+
+    Attributes:
+        key (str): The feedback key for the evaluator.
+        choices (List[str]): List of valid categorical values that can be assigned.
+        description (str): Detailed description provided to the LLM judge of what
+            this score evaluates.
+        reasoning_key (Optional[str]): Key used to store the reasoning/explanation
+            for the score. Defaults to None.
```
hinthornw, line 22: Say what None means (don't include a reasoning/ CoT field)
Ditto with description (Defaults to ""The explanation for the score."")
I still think we should have a better default than "The explanation for the score." - like "Think step-by-step about what the correct score should be."

python/tests/integration_tests/test_llm_evaluator.py
isahers1: added some backward compatibility tests

python/langsmith/evaluation/llm_evaluator.py
```diff
@@ -10,35 +12,133 @@
...
+    reasoning_description: Optional[str] = None
+    include_explanation: bool = False  # Deprecated
+    explanation_description: Optional[str] = None  # Deprecated
+
+    def __init__(self, **data):
+        """Initialize CategoricalScoreConfig."""
+        if data.get("include_explanation") and data.get("reasoning_key"):
+            raise ValueError(
+                "Cannot include both include_explanation and reasoning_key, "
+                "please just use reasoning_key - include_explanation has been deprecated"  # noqa: E501
```
hinthornw, line 42: nit: If we're already splitting across lines let's split them within the limit rather than using noqa

python/langsmith/evaluation/llm_evaluator.py
```diff
@@ -10,35 +12,133 @@
...
     choices: List[str]
     description: str
-    include_explanation: bool = False
-    explanation_description: Optional[str] = None
+    reasoning_key: Optional[str] = None
+    reasoning_description: Optional[str] = None
+    include_explanation: bool = False  # Deprecated
+    explanation_description: Optional[str] = None  # Deprecated
+
+    def __init__(self, **data):
```
hinthornw, line 37: A sad byproduct of us doing this is you lose IDE completion. Could we add the keys here?
Ditto with below.

python/tests/integration_tests/test_llm_evaluator.py
```diff
@@ -207,3 +196,83 @@ async def apredict(inputs: dict) -> dict:
         evaluators=[reference_accuracy, accuracy],
         experiment_prefix=__name__ + "::test_evaluate.aevaluate",
     )
+
+
+@pytest.mark.parametrize(
+    "config_class", [CategoricalScoreConfig, ContinuousScoreConfig]
+)
+def test_backwards_compatibility(config_class) -> None:
```
hinthornw, line 204: nit: Thse could probably go in `unit_tests/*` ratehr than in integration tests

hinthornw (commented):
Almost there!

PR #1186, opened 2024-11-07
add LLMEvaluator to JS

isahers1:


js/src/evaluation/llm_evaluator.ts
```diff
@@ -0,0 +1,248 @@
+import { initChatModel } from "langchain/chat_models/universal";
+import { ChatPromptTemplate } from "@langchain/core/prompts";
+import * as uuid from "uuid";
+import { EvaluationResult, EvaluationResults, RunEvaluator } from "./evaluator.js";
+import type { Run, Example } from "../schemas.js";
+export class CategoricalScoreConfig {
```
hinthornw, line 6: We should add good typedocs :)

js/src/evaluation/llm_evaluator.ts
```diff
@@ -0,0 +1,248 @@
...
+  key: string;
+  choices: string[];
+  description: string;
+  includeExplanation: boolean;
```
hinthornw, line 10: Not a fan of boolean flags. Would prefer something like "reasoningKey" that if not provided is undefined and not included (if that's what we want for default behavior)

js/src/evaluation/llm_evaluator.ts
```diff
@@ -0,0 +1,248 @@
...
+    this.includeExplanation = params.includeExplanation ?? false;
+    this.explanationDescription = params.explanationDescription;
+  }
+}
+
+type ScoreConfig = CategoricalScoreConfig | ContinuousScoreConfig;
+
+function createScoreJsonSchema(
+  scoreConfig: ScoreConfig,
+  reasoningKey: string
```
hinthornw, line 57: Ditto above, but seems like a better API to make opting in to CoT only take one argument rather than 2 (split across two things)

js/src/evaluation/llm_evaluator.ts
```diff
@@ -0,0 +1,248 @@
...
+      type: "string",
+      description: scoreConfig.explanationDescription || "The explanation for the score.",
+    };
+  }
+
+  if ("choices" in scoreConfig) {
+    properties.score = {
+      type: "string",
+      enum: scoreConfig.choices,
+      description: `The score for the evaluation, one of ${scoreConfig.choices.join(", ")}.`,
```
hinthornw, line 72: This is really confusing to me tbh. 

Why do we put categorical in the numeric score field? Why call it a score? Why not call it "category" or something obvious?

Score implies sortability / comparability to me

js/src/evaluation/llm_evaluator.ts
```diff
@@ -0,0 +1,248 @@
...
+function createScoreJsonSchema(
+  scoreConfig: ScoreConfig,
+  reasoningKey: string
+): Record<string, any> {
+  const properties: Record<string, any> = {};
+
+  if (scoreConfig.includeExplanation) {
+    properties[reasoningKey] = {
+      type: "string",
+      description: scoreConfig.explanationDescription || "The explanation for the score.",
```
hinthornw, line 64: I would say "First, think step by step to explain your score."

JS doesn't really guarantee order of the schema that's sent up

js/src/evaluation/llm_evaluator.ts
isahers1: I changed the key to "value", but we also do this in the output parsing which I think solves this a little? not 100% sure, lmk what you think:
```
if ("choices" in this.scoreConfig) {
      const value = output.score;
      const explanation = output[this.reasoningKey];
      return {
        key: this.scoreConfig.key,
        value,
        comment: explanation,
        sourceRunId: output.runId
      };
    }
```

js/package.json
```diff
@@ -106,14 +106,15 @@
     "p-queue": "^6.6.2",
     "p-retry": "4",
     "semver": "^7.6.3",
-    "uuid": "^10.0.0"
+    "uuid": "^10.0.0",
+    "langchain": "^0.3.3",
```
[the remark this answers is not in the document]
hinthornw: Synced with Jacob to double check. Steps to handle this:

1. Rm from dependencies
2. Make the llm evaluator a separate entrypoint (see traceablea nd the create_entrypoints.js script)
3. We should probably refrain from importing init chat model from the top since that requires langchain to be installed as well as the sub modules. Maybe accept a chat model as an argument and have them initialize, and in our docs we show using that. Core I think more OK

js/package.json
```diff
@@ -132,13 +135,14 @@
...
     "jest": "^29.5.0",
-    "langchain": "^0.3.3",
     "openai": "^4.67.3",
     "prettier": "^2.8.8",
     "ts-jest": "^29.1.0",
     "ts-node": "^10.9.1",
     "typescript": "^5.4.5",
-    "zod": "^3.23.8"
+    "zod": "^3.23.8",
+    "@langchain/core": "^0.3.14",
```
jacoblee93, line 144: Good to keep these sorted and avoid touching unnecessary lines

js/scripts/create-entrypoints.js
```diff
@@ -12,6 +12,7 @@ const entrypoints = {
   traceable: "traceable",
   evaluation: "evaluation/index",
   "evaluation/langchain": "evaluation/langchain",
+  "evaluation/llm": "evaluation/llm_evaluator",
```
jacoblee93, line 15: Naming should match filepath

So rename file below to `evaluation/llm.ts`

js/src/evaluation/llm_evaluator.ts
```diff
@@ -0,0 +1,290 @@
...
+  }
+  private async initialize(
+    promptTemplate: string | [string, string][],
+    scoreConfig: ScoreConfig,
+    chatModel: BaseLanguageModel,
+    mapVariables?: (run: Run, example?: Example) => Record<string, any>
+  ) {
+    try {
+      // Store the configuration
+      this.scoreConfig = scoreConfig;
```
jacoblee93, line 174: More conventional to make this a constructor instead of a factory method

js/src/evaluation/llm_evaluator.ts
```diff
@@ -0,0 +1,290 @@
...
+          "Multiple output keys are present in example.outputs. Please provide a map_variables function."
+        );
+      }
+      variables.expected = Object.values(example.outputs)[0];
+    }
+
+    return variables;
+  }
+
+  private parseOutput(
```
jacoblee93, line 265: Prefer protected instead of private to make subclassing possible

js/src/evaluation/llm_evaluator.ts
```diff
@@ -0,0 +1,290 @@
...
+export class LLMEvaluator implements RunEvaluator {
+  prompt: any;
+  mapVariables?: (run: Run, example?: Example) => Record<string, any>;
+  scoreConfig: ScoreConfig;
+  scoreSchema: Record<string, any>;
+  runnable: any;
+
+  constructor() {}
+
+  static async create(params: LLMEvaluatorParams): Promise<LLMEvaluator> {
```
jacoblee93, line 156: You definitely shouldn't have two methods for this, ideally you just use the constructor

js/src/evaluation/llm_evaluator.ts
```diff
@@ -0,0 +1,290 @@
...
+}
+
+interface LLMEvaluatorParams {
+  promptTemplate: string | [string, string][];
+  scoreConfig: ScoreConfig;
+  chatModel: BaseLanguageModel;
+  mapVariables?: (run: Run, example?: Example) => Record<string, any>;
+}
+
+export class LLMEvaluator implements RunEvaluator {
```
jacoblee93, line 147: No docstring?

js/src/evaluation/llm_evaluator.ts
```diff
@@ -0,0 +1,290 @@
...
+  prompt: any;
+  mapVariables?: (run: Run, example?: Example) => Record<string, any>;
```
jacoblee93, line 149: What are mapVariables?

js/src/tests/llm_evaluator.int.test.ts
```diff
@@ -0,0 +1,183 @@
...
+      ["human", "Rate this response from 0 to 1: {response}"],
+    ],
+    scoreConfig: new ContinuousScoreConfig({
+      key: "rating",
+      description: "Quality rating from 0 to 1",
+      min: 0,
+      max: 1,
+    }),
+    chatModel: CHAT_MODEL,
+    mapVariables: (run: any, _example?: any) => ({
```
jacoblee93, line 177: Is there a better name for this we can use? `mapVariables` sounds like a map of variables or something and `.map()` has a specific meaning in JS (and most languages?)

js/src/evaluation/llm_evaluator.ts
```diff
@@ -0,0 +1,290 @@
...
+    required: scoreConfig.reasoningKey
+      ? ["choices" in scoreConfig ? "value" : "score", scoreConfig.reasoningKey]
+      : ["choices" in scoreConfig ? "value" : "score"],
+  };
+}
+
+interface LLMEvaluatorParams {
+  promptTemplate: string | [string, string][];
+  scoreConfig: ScoreConfig;
+  chatModel: BaseLanguageModel;
```
jacoblee93, line 143: Should be `BaseChatModel`

js/src/tests/llm_evaluator.int.test.ts
isahers1: Open to suggestions, this is what we use in python, agree it is somewhat confusing. The true purposes is "map variables from your example/run to variables used in your evaluator prompt" - not sure how to best condense this into a succinct variable name.

js/src/evaluation/llm_evaluator.ts
isahers1: see comment below

js/src/tests/llm_evaluator.int.test.ts
jacoblee93: `formatPrompt`?

jacoblee93: Obviously would need an update on Python side but I think this name is not good

js/src/evaluation/llm_evaluator.ts
jacoblee93, line 1: Delete this

js/src/evaluation/llm.ts
```diff
@@ -0,0 +1,295 @@
...
+
+    // Output handling
+    if (!run.outputs || Object.keys(run.outputs).length === 0) {
+      throw new Error(
+        "No output keys are present in run.outputs but the prompt requires 'output'."
+      );
+    }
+    if (Object.keys(run.outputs).length !== 1) {
+      throw new Error(
+        "Multiple output keys are present in run.outputs. Please provide a map_variables function."
```
jacoblee93, line 247: This uses the same function as inputs? Is that a typo?

Sidebar: I generally dislike this `keys` pattern but maybe it's more popular in Python

PR #1188, opened 2024-11-07
infra: drop py3.8 CI

baskaryan:


python/langsmith/run_helpers.py
```diff
@@ -1034,14 +1032,10 @@ async def __aexit__(
         ctx = copy_context()
         if exc_type is not None:
             await asyncio.shield(
-                aitertools.aio_to_thread(
-                    self._teardown, exc_type, exc_value, traceback, __ctx=ctx
```
baskaryan, line 1038: @hinthornw ok to remove __ctx here? afaict to_thread already calls copy_context()

hinthornw: i think so but would need to check

PR #1196, opened 2024-11-08
Add six to requirements.txt

Tayamarn:
Six is used at least in this file:
https://github.com/langchain-ai/langsmith-sdk/blob/main/python/langsmith/_internal/_patch.py#L4

So, the library causes errors, if six is not already installed. Better add it to required libs.


efriis:
fixed in #1197 instead

PR #1199, opened 2024-11-08
rfc: @ls.pytest.mark.parametrize interface

baskaryan:
almost certainly not handling lazy eval correctly, but what do we think of interface?
- automatically logs a pass/fail feedback based on test pass/fail
- can return whatever other feedback you want as well

```python
@ls.pytest.mark.parametrize("Sample Dataset 3", (lambda x: x))
def test_parametrize(inputs, outputs, reference_outputs) -> list:
    assert inputs == outputs
    return [{"key": "foo", "value": "bar"}]
```

some example experiments here https://dev.smith.langchain.com/public/e7782ea0-3de5-4352-8cd4-7b2cdbb03e4c/d

### DIFF
--- a/python/langsmith/pytest/mark.py
+++ b/python/langsmith/pytest/mark.py
@@ -1,0 +1,93 @@
+from __future__ import annotations
+
+import inspect
+from typing import Any, Callable, Optional
+
+import pytest
+
+from langsmith import evaluate
+from langsmith.evaluation._runner import TARGET_T
+
+
+def parametrize(
+    dataset_name: str,
+    target_fn: TARGET_T,
+    *,
+    client: Optional[Any] = None,
+    max_concurrency: Optional[int] = None,
+) -> Callable:
+    """Decorator to parametrize a test function with LangSmith dataset examples.
+
+    Args:
+        dataset_name: Name of the LangSmith dataset to use
+        target_fn: Function to test that takes inputs dict and returns outputs dict
+        client: Optional LangSmith client to use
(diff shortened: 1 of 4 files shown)


hinthornw:
Things I like about this:
1. Can connect to dataset
2. outputs are fairly localized/transpernt
3. Trace seems sensical (has outputs by default)


Things I don't looove about this relative to `@unit`

1. Seems harder to check multi-step things
2. The actual system is run "outside" the test function

python/langsmith/pytest/mark.py
```diff
@@ -0,0 +1,93 @@
...
+            """Wrapped test function that gets parametrized with results."""
+            # Fail the test if the evaluation failed
+            eval_results = example_result["evaluation_results"]["results"]
+            if not eval_results:
+                pytest.fail("No evaluation results")
+
+            pass_result = [r for r in eval_results if r.key == "pass"][0]
+            if not pass_result.score:
+                error = pass_result.comment
+                pytest.fail(
```
jakerachleff, line 87: How would you set failure conditions? I assume people don't want to actually fail if any evaluation fails?

jakerachleff: which might mean allowing customizability on the interface on this

baskaryan: this only fails if the actual test raises an error (we need to add a manual pytest.fail for that bc we catch and log all errors in the wrapper L48). so it is customizable by default

PR #1200, opened 2024-11-09
rfc: accept simple evaluators

baskaryan:
can write evaluators like this:

```python

def simp(inputs: dict, outputs: dict, reference_outputs: dict) -> dict:
    return {"results": [
        {"score": inputs == outputs, "key": 'identity'}, 
        {"score": outputs == reference_outputs, "key": "correct"}
    ]}

evaluate(
    (lambda x: x),
    data="Sample Dataset 3",
    evaluators=[simp],
)
```

example experiment: left-tray-86  https://dev.smith.langchain.com/public/e7782ea0-3de5-4352-8cd4-7b2cdbb03e4c/d

### DIFF
--- a/python/langsmith/evaluation/evaluator.py
+++ b/python/langsmith/evaluation/evaluator.py
@@ -17,7 +17,7 @@ from typing import (
     cast,
 )
 
-from typing_extensions import TypedDict
+from typing_extensions import TypedDict, get_type_hints
 
 try:
     from pydantic.v1 import (  # type: ignore[import]
@@ -194,6 +194,8 @@ class DynamicRunEvaluator(RunEvaluator):
             func (Callable): A function that takes a `Run` and an optional `Example` as
             arguments, and returns a dict or `ComparisonEvaluationResult`.
         """
+        func = _normalize_evaluator_func(func)
+
         wraps(func)(self)
         from langsmith import run_helpers  # type: ignore
 
@@ -632,3 +634,43 @@ def comparison_evaluator(
 ) -> DynamicComparisonRunEvaluator:
     """Create a comaprison evaluator from a function."""
     return DynamicComparisonRunEvaluator(func)
+
+
+def _normalize_evaluator_func(
+    func: Callable,
+) -> Callable[[Run, Optional[Example]], _RUNNABLE_OUTPUT]:
+    # for backwards compatibility, if args are untyped we assume they correspond to
+    # Run and Example:
+    if not (type_hints := get_type_hints(func)):
+        return func
+    elif {Run, Example, Optional[Example]}.intersection(type_hints.values()):
+        return func
+    else:
+        sig = inspect.signature(func)
+        num_positional = len(
+            [
+                p
+                for p in sig.parameters.values()
+                if p.kind in (p.POSITIONAL_OR_KEYWORD, p.POSITIONAL_ONLY)
+            ]
+        )
+        has_positional_var = any(
+            p.kind == p.VAR_POSITIONAL for p in sig.parameters.values()
+        )
+        if not (
+            num_positional in (2, 3) or (num_positional <= 3 and has_positional_var)
+        ):
+            msg = ""
+            raise ValueError(msg)
+
+        def wrapper(run: Run, example: Example) -> _RUNNABLE_OUTPUT:
+            args = (example.inputs, run.outputs, example.outputs)
+            if has_positional_var:
+                return func(*args)
+            else:
+                return func(*args[:num_positional])
+
+        wrapper.__name__ = (
+            getattr(func, "__name__") if hasattr(func, "__name__") else wrapper.__name__
+        )
+        return wrapper
--- a/python/langsmith/evaluation/_runner.py
+++ b/python/langsmith/evaluation/_runner.py
@@ -86,6 +86,7 @@ EVALUATOR_T = Union[
         [schemas.Run, Optional[schemas.Example]],
         Union[EvaluationResult, EvaluationResults],
     ],
+    Callable[..., Union[dict, EvaluationResults, EvaluationResult]],
 ]
 AEVALUATOR_T = Union[
     Callable[


python/langsmith/evaluation/evaluator.py
```diff
@@ -632,3 +634,43 @@ def comparison_evaluator(
...
+                if p.kind in (p.POSITIONAL_OR_KEYWORD, p.POSITIONAL_ONLY)
+            ]
+        )
+        has_positional_var = any(
+            p.kind == p.VAR_POSITIONAL for p in sig.parameters.values()
+        )
+        if not (
+            num_positional in (2, 3) or (num_positional <= 3 and has_positional_var)
+        ):
+            msg = ""
```
hinthornw, line 663: nit

python/langsmith/evaluation/evaluator.py
```diff
@@ -632,3 +634,43 @@ def comparison_evaluator(
...
+            raise ValueError(msg)
+
+        def wrapper(run: Run, example: Example) -> _RUNNABLE_OUTPUT:
+            args = (example.inputs, run.outputs, example.outputs)
```
hinthornw, line 667:
```suggestion
            args = (example.inputs, run.outputs or {}, example.outputs or {})
```

hinthornw (commented):
Nice i like the start will re-review this morning

python/langsmith/evaluation/evaluator.py
```diff
@@ -632,3 +636,70 @@ def comparison_evaluator(
 ) -> DynamicComparisonRunEvaluator:
     """Create a comaprison evaluator from a function."""
     return DynamicComparisonRunEvaluator(func)
+
+
+def _normalize_evaluator_func(
```
jakerachleff, line 641: might be nice to add like a couple unit tests on this to make it obvious it's working

python/langsmith/evaluation/evaluator.py
```diff
@@ -632,3 +636,70 @@ def comparison_evaluator(
...
+    func: Callable,
+) -> Union[
+    Callable[[Run, Optional[Example]], _RUNNABLE_OUTPUT],
+    Callable[[Run, Optional[Example]], Awaitable[_RUNNABLE_OUTPUT]],
+]:
+    # for backwards compatibility, if args are untyped we assume they correspond to
+    # Run and Example:
+    if not (type_hints := get_type_hints(func)):
```
jakerachleff, line 649: Do we want to add debug logs letting you know what function type is being used? Might be helpful since we tell people to enable debug logs for debugging issues in the SDK?

hinthornw:
Do we want it to be like pytest where it's all by name?
run, example, inputs, predictions, reference

baskaryan:
> Do we want it to be like pytest where it's all by name? run, example, inputs, predictions, reference

yea i like that. for backwards compat can't enforce run/example but can enforce the others

python/langsmith/evaluation/evaluator.py
agola11, line 649: shouldn't we check the number of args here? traditional evaluators have run and example whereas the simple evaluators take 3 args

python/langsmith/evaluation/evaluator.py
```diff
@@ -632,3 +636,70 @@ def comparison_evaluator(
...
+        )
+        if not (
+            num_positional in (2, 3) or (num_positional <= 3 and has_positional_var)
+        ):
+            msg = (
+                "Invalid evaluator function. Expected to take either 2 or 3 positional "
+                "arguments. Please see "
+                "https://docs.smith.langchain.com/evaluation/how_to_guides/evaluation/evaluate_llm_application#use-custom-evaluators"  # noqa: E501
+            )
+            raise ValueError(msg)
```
agola11, line 673: seems like this check on arg length should be moved up

python/langsmith/evaluation/evaluator.py
```diff
@@ -645,3 +649,74 @@ def comparison_evaluator(
...
+]:
+    supported_args = ("run", "example", "inputs", "outputs", "reference_outputs")
+    sig = inspect.signature(func)
+    positional_no_default = [
+        pname
+        for pname, p in sig.parameters.items()
+        if p.kind in (p.POSITIONAL_OR_KEYWORD, p.POSITIONAL_ONLY)
+        and p.default is p.empty
+    ]
+    if not positional_no_default or (
```
hinthornw, line 668: ooc, why do we require at least one positional one?

python/langsmith/evaluation/evaluator.py
```diff
@@ -645,3 +649,74 @@ def comparison_evaluator(
...
+
+            def wrapper(run: Run, example: Example) -> _RUNNABLE_OUTPUT:
+                arg_map = {
+                    "run": run,
+                    "example": example,
+                    "inputs": example.inputs,
+                    "outputs": run.outputs or {},
+                    "reference_outputs": example.outputs or {},
+                }
+                args = (arg_map[arg] for arg in positional_no_default)
```
hinthornw, line 714: If I put a default in an arg this silently never provides the matching value. Would either want to validate ahead of time that no default is provided (preferred) or pass it in anyway (think not preferred)

python/langsmith/evaluation/evaluator.py
```diff
@@ -645,3 +649,74 @@ def comparison_evaluator(
...
+        if p.kind in (p.POSITIONAL_OR_KEYWORD, p.POSITIONAL_ONLY)
+        and p.default is p.empty
+    ]
+    if not positional_no_default or (
+        not all(pname in supported_args for pname in positional_no_default)
+        and len(positional_no_default) != 2
+    ):
+        msg = (
+            f"Invalid evaluator function. Must have at least one positional "
+            f"argument. Supported positional arguments are {supported_args}. Please "
```
hinthornw, line 674: Could we include a description of what each argument is here?

python/langsmith/evaluation/evaluator.py
```diff
@@ -645,3 +649,74 @@ def comparison_evaluator(
...
+            f"see https://docs.smith.langchain.com/evaluation/how_to_guides/evaluation/evaluate_llm_application#use-custom-evaluators"
```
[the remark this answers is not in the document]
baskaryan: ye im updating it as we speak, but there will be redirects

python/langsmith/evaluation/evaluator.py
baskaryan: we only pass in the supported args as positional args, so equivalent to enforcing that there's at least one supported arg

python/langsmith/evaluation/evaluator.py
baskaryan: feels like ppl should just look at api ref for that?

python/langsmith/evaluation/evaluator.py
baskaryan: think im just going to remove the check for whether it has defaults or not, doesn't seem necessary

PR #1203, opened 2024-11-11
python[patch]: evaluators can return primitives

baskaryan:
```python

def foo(run, example):
    return 0

def bar(run, example):
    return "long"

# removed, needs to be list of dict
# def baz(run, example):
#     return [0, 0.2, "how are ya"]

def app(inputs):
    return inputs

evaluate(app, data="Sample Dataset 3", evaluators=[foo, bar, baz])
```

### DIFF
--- a/python/langsmith/evaluation/evaluator.py
+++ b/python/langsmith/evaluation/evaluator.py
@@ -224,7 +224,7 @@ class DynamicRunEvaluator(RunEvaluator):
         self,
         result: Union[EvaluationResult, dict],
         source_run_id: uuid.UUID,
-        allow_no_key: bool = False,
+        default_name: str = "",
     ) -> EvaluationResult:
         if isinstance(result, EvaluationResult):
             if not result.source_run_id:
@@ -237,8 +237,7 @@ class DynamicRunEvaluator(RunEvaluator):
                     f" 'key' and optional 'score'; got empty result: {result}"
                 )
             if "key" not in result:
-                if allow_no_key:
-                    result["key"] = self._name
+                result["key"] = default_name
             if all(k not in result for k in ("score", "value", "comment")):
                 raise ValueError(
                     "Expected an EvaluationResult object, or dict with a metric"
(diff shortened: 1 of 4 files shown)


baskaryan:
how ^ example gets logged

Screenshot 2024-11-11 at 10 38 17 AM

hinthornw:
The list return type seems unexpected to me - I wouldn't expect it to make separate keys for each value

baskaryan:
> > > The list return type seems unexpected to me - I wouldn't expect it to make separate keys for each value
> > 
> > 
> > are we able to support list values as feedback?
> 
> Only dict or string it seems.
> 
> Maybe doing the _ix suffix is preferable. I'm not sure honestly. Probalby less surprising than just logging duplicates to the same key.
> 
> It just messes with the experiment averages (we'd be averaging over index 1 and over index 2 )

feel less strongly about list behavior either way, think main use case is supporting int/float/bool/str

hinthornw:
Maybe let's land with the numeric and string value support but hold off on list behavior?

Maybe add support for list[evaluationresultlike] to make the "results" key not necessary

python/langsmith/evaluation/evaluator.py
```diff
@@ -260,32 +259,46 @@ def _coerce_evaluation_results(
...
         self,
-        result: Union[EvaluationResult, EvaluationResults, dict],
+        result: Union[
+            EvaluationResult, EvaluationResults, dict, str, int, bool, float, list
+        ],
         source_run_id: uuid.UUID,
     ) -> Union[EvaluationResult, EvaluationResults]:
-        if isinstance(result, EvaluationResult):
+        if isinstance(result, (bool, float, int)):
+            result = {"score": result}
```
jakerachleff, line 278: So if I have four categories that are numbers for some reason (I'm classifying college class levels and they're 100 level, 200 level, 300 level, etc), then I should do `str(value)` to explicitly use categorical scores?

python/tests/unit_tests/evaluation/test_runner.py
```diff
@@ -347,11 +362,26 @@ async def score_value_first(run, example):
...
 
+    async def eval_float(run, example):
+        ordering_of_stuff.append("evaluate")
+        return 0.2
+
+    async def eval_str(run, example):
+        ordering_of_stuff.append("evaluate")
+        return "good"
+
+    async def eval_list(run, example):
```
jakerachleff, line 373: maybe good to confirm that a list of ints, for example, doesn't work?

python/langsmith/evaluation/evaluator.py
```diff
@@ -260,32 +259,46 @@ def _coerce_evaluation_results(
             cp = results.copy()
             cp["results"] = [
                 self._coerce_evaluation_result(r, source_run_id=source_run_id)
-                for r in results["results"]
+                for i, r in enumerate(results["results"])
```
hinthornw, line 262: is `i` needed anymore?

hinthornw (approved):

PR #1209, opened 2024-11-13
[python] support uploading examples with attachments and running evals on examples with attachments

isahers1:


python/langsmith/schemas.py
```diff
@@ -88,6 +88,9 @@ class ExampleCreate(ExampleBase):
     created_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
     split: Optional[Union[str, List[str]]] = None
 
+class ExampleCreateWithAttachments(ExampleCreate):
+    """Example create with attachments."""
+    attachments: Optional[List[Attachment]] = None
```
agola11, line 93: this should actually be `Attachments` which is defined here as `Attachments = Dict[str, Union[Tuple[str, bytes], Attachment]]` in `schemas.py`

it maps the name of the attachment to the attachment itself, eg

```
{
  "attachment1": ("image/png", b"...")
}
```

python/langsmith/client.py
```diff
@@ -3369,6 +3370,94 @@ def create_example_from_run(
...
+                            {"Content-Length": str(len(attachment.data))},
+                        ),
+                    ),
+                    
+        encoder = rqtb_multipart.MultipartEncoder(parts, boundary=BOUNDARY)
+        if encoder.len <= 20_000_000:  # ~20 MB
+            data = encoder.to_string()
+        else:
+            data = encoder
+
```
agola11, line 3446: I'm assuming this is what we do for runs?

python/langsmith/client.py
```diff
@@ -3369,6 +3370,94 @@ def create_example_from_run(
...
+            ),
+        
+            if example.attachments:
+                for attachment in example.attachments:
+                    parts.append(
+                        f"{example_id}.attachment.{attachment.mime_type}",
+                        (
+                            None,
+                            attachment.data,
+                            "application/json", # I feel like this is wrong
```
agola11, line 3436: yeah, this should be the mime type

python/langsmith/client.py
```diff
@@ -3369,6 +3370,94 @@ def create_example_from_run(
...
+                    outputsb,
+                    "application/json",
+                    {"Content-Length": str(len(outputsb))},
+                ),
+            ),
+        
+            if example.attachments:
+                for attachment in example.attachments:
+                    parts.append(
+                        f"{example_id}.attachment.{attachment.mime_type}",
```
agola11, line 3432: this should be `{example_id}.attachment.{attachment_name}` where `attachment_name` is the key in the dictionary

python/langsmith/client.py
isahers1: yeah this code was copy pasted

python/langsmith/client.py
```diff
@@ -3369,6 +3370,133 @@ def create_example_from_run(
             created_at=created_at,
         )
 
+    def upsert_examples_multipart(
+        self,
+        *,
+        upserts: List[ls_schemas.ExampleCreateWithAttachments] = [],
```
agola11, line 3376: should not initialize this to empty list due to the way python handles mutable default arguments, see here https://nikos7am.com/posts/mutable-default-arguments/

to fix, just remove the default

python/langsmith/client.py
```diff
@@ -3369,6 +3370,133 @@ def create_example_from_run(
...
+    ) -> None:
+        """Upsert examples."""
+        # not sure if the below checks are necessary
+        if not isinstance(upserts, list):
+            raise TypeError(f"upserts must be a list, got {type(upserts)}")
+        for item in upserts:
+            if not isinstance(item, ls_schemas.ExampleCreateWithAttachments):
+                raise TypeError(f"Each item must be ExampleCreateWithAttachments, got {type(item)}")
```
agola11, line 3384: not necessary, we don't check types like this elsewhere

python/langsmith/client.py
```diff
@@ -3369,6 +3370,133 @@ def create_example_from_run(
...
+            
+        parts: list[MultipartPart] = []
+
+        for example in upserts:
+            if example.id is not None:
+                example_id = str(example.id)  # is the conversion to string neccessary?
```
agola11, line 3390: yes,  I believe so

python/langsmith/client.py
```diff
@@ -3369,6 +3370,133 @@ def create_example_from_run(
...
+            else:
+                example_id = str(uuid.uuid4())
+
+            remaining_values = {
```
agola11, line 3394: nit: rename to example_body

python/langsmith/client.py
```diff
@@ -3369,6 +3370,133 @@ def create_example_from_run(
...
+
+            (
+                parts.append(
+                    (
+                        f"{example_id}",
+                        (
+                            None,
+                            valb,
+                            "application/json",
+                            {"Content-Length": str(len(valb))},
```
agola11, line 3412: This should work without content length

python/langsmith/client.py
```diff
@@ -3369,6 +3370,133 @@ def create_example_from_run(
...
+
+            (
+                parts.append(
+                    (
+                        f"{example_id}.inputs",
+                        (
+                            None,
+                            inputsb,
+                            "application/json",
+                            {"Content-Length": str(len(inputsb))},
```
agola11, line 3429: same here, not needed for now

python/langsmith/client.py
```diff
@@ -3369,6 +3370,133 @@ def create_example_from_run(
...
+
+            (
+                parts.append(
+                    (
+                        f"{example_id}.outputs",
+                        (
+                            None,
+                            outputsb,
+                            "application/json",
+                            {"Content-Length": str(len(outputsb))},
```
agola11, line 3443: same here, not needed for now

python/langsmith/client.py
```diff
@@ -3369,6 +3370,133 @@ def create_example_from_run(
...
+                        mime_type, data = attachment
+                        (
+                            parts.append(
+                                (
+                                    f"{example_id}.attachment.{name}",
+                                    (
+                                        None,
+                                        data,
+                                        f"{mime_type}; length={len(data)}",
+                                        {},
```
agola11, line 3461: I think if you add a content length header here, it will be interpreted correctly by the encoder

python/langsmith/client.py
```diff
@@ -3369,6 +3370,133 @@ def create_example_from_run(
...
+
+        encoder = rqtb_multipart.MultipartEncoder(parts, boundary=BOUNDARY)
+        if encoder.len <= 20_000_000:  # ~20 MB
+            data = encoder.to_string()
+        else:
+            data = encoder
+
+        response = self.request_with_retries(
+            "POST",
+            "/v1/platform/examples/multipart",  # No clue what this is supposed to be
```
agola11, line 3489: this is correct! You can rm the comment

python/bench/upload_examples_bench.py
```diff
@@ -0,0 +1,124 @@
...
+
+def create_example_data(dataset_id: str, json_size: int) -> Dict:
+    """Create a single example data object."""
+    return ExampleCreateWithAttachments(**{
+        "dataset_id": dataset_id,
+        "inputs": create_large_json(json_size),
+        "outputs": create_large_json(json_size),
+    })
+
+DATASET_NAME = "TEST DATASET"
```
agola11, line 41: probably worth giving this a more unique name since you delete it at the start of every benchmark

python/bench/upload_examples_bench.py
```diff
@@ -0,0 +1,124 @@
...
+        old_start = time.perf_counter()
+        inputs=[e.inputs for e in examples]
+        outputs=[e.outputs for e in examples]
+        # the create_examples endpoint fails above 20mb
+        try:
+            client.create_examples(inputs=inputs,
+                                outputs=outputs,dataset_id=dataset.id)
+            old_elapsed = time.perf_counter() - old_start
+        except:
+            old_elapsed = 1000000
```
agola11, line 73: would just remove the try except. If the code throws, then the benchmark should just exit

python/bench/upload_examples_bench.py
```diff
@@ -0,0 +1,124 @@
...
+    for metric in metrics:
+        print(f"{metric:<15} {results['old'][metric]:>20.4f} {results['new'][metric]:>20.4f}")
+    
+    print("-" * 60)
+    print(f"{'Throughput':<15} {num_examples / results['old']['mean']:>20.2f} {num_examples / results['new']['mean']:>20.2f}")
+    print("(examples/second)")
+
+
+if __name__ == "__main__":
+    main(json_size, num_examples)
```
agola11, line 124: nit: newline

python/langsmith/client.py
```diff
@@ -3369,6 +3370,134 @@ def create_example_from_run(
             created_at=created_at,
         )
 
+    def upsert_examples_multipart(
+        self,
+        *,
+        upserts: List[ls_schemas.ExampleCreateWithAttachments] = None,
+    ) -> dict: # Should we create an object for the return type - like UpsertExamplesResponse?
```
agola11, line 3377: Yeah, I would just use a typeddict, pydantic overkill for this

python/bench/upload_examples_bench.py
```diff
@@ -0,0 +1,124 @@
...
+json_size = 1000
+num_examples = 1000
+
+def main(json_size: int, num_examples: int):
+    """
+    Run benchmarks with different combinations of parameters and report results.
+    """
+    results = benchmark_example_uploading(num_examples=num_examples, json_size=json_size)
+    
+    print(f"\nBenchmark Results for {num_examples} examples with JSON size {json_size}:")
```
agola11, line 109: nice table printout

python/langsmith/client.py
```diff
@@ -3369,6 +3370,134 @@ def create_example_from_run(
...
+        """Upsert examples."""
+        if not (self.info.instance_flags or {}).get(
+                "examples_multipart_enabled", False
+            ):
+            raise ValueError("Your LangChain version does not allow using the multipart examples endpoint, please update to the latest version.")
```
agola11, line 3382:
```suggestion
            raise ValueError("Your LangSmith version does not allow using the multipart examples endpoint, please update to the latest version.")
```

python/tests/integration_tests/test_client.py
```diff
@@ -368,6 +369,77 @@ def test_error_surfaced_invalid_uri(uri: str) -> None:
...
+        dataset_id=uuid4(), # not a real dataset
+        inputs={"text": "foo bar"},
+        outputs={"response": "baz"},
+        attachments={
+            "my_file": ("text/plain", b"more test content"),
+        },
+    )
+
+    with pytest.raises(LangSmithNotFoundError):
+        langchain_client.upsert_examples_multipart(upserts=[example_3])
```
agola11, line 433: why should this raise a not found?

isahers1: because it can't find the dataset I believe

python/langsmith/schemas.py
```diff
@@ -63,8 +63,22 @@ def my_function(bar: int, my_val: Attachment):
...
 
-Attachments = Dict[str, Union[Tuple[str, bytes], Attachment]]
-"""Attachments associated with the run. Each entry is a tuple of (mime_type, bytes)."""
+Attachments = Dict[str, Union[Tuple[str, bytes], Attachment, str]]
+"""Attachments associated with the run. 
+Each entry is a tuple of (mime_type, bytes), or a fliepath"""
+
+
+@runtime_checkable
+class BinaryIOLike(Protocol):
```
agola11, line 72: What is the point of this?

python/langsmith/schemas.py
```diff
@@ -63,8 +63,22 @@ def my_function(bar: int, my_val: Attachment):
     data: bytes
 
 
-Attachments = Dict[str, Union[Tuple[str, bytes], Attachment]]
-"""Attachments associated with the run. Each entry is a tuple of (mime_type, bytes)."""
+Attachments = Dict[str, Union[Tuple[str, bytes], Attachment, str]]
```
agola11, line 66: A couple comments:

This should be:
```
Attachments = Dict[str, Union[Tuple[str, bytes], Attachment, Tuple[str, Path]]
```

However we need to make sure this updated schema does not break existing code.

python/langsmith/_internal/_operations.py
```diff
@@ -256,18 +257,23 @@ def serialized_run_operation_to_multipart_parts_and_context(
...
-                        {"Content-Length": str(len(valb))},
-                    ),
+                        f"attachment.{op.id}.{n}",
+                        (
+                            None,
+                            valb,
+                            content_type,
+                            {"Content-Length": str(len(valb))},
+                        ),
+                    )
```
isahers1, line 275: I believe this is the only other place we access the attachments, but I could definitely be missing some. I just skipped it in this case for speed of getting something out, but we should fix this quickly after

python/langsmith/evaluation/_arunner.py
```diff
@@ -1021,6 +1046,46 @@ def _get_run(r: run_trees.RunTree) -> None:
...
+        raise ValueError(
+            "Target function must accept at most two positional "
+            "arguments (inputs, attachments)"
+        )
+    elif len(positional_params) == 2:
+        mismatches = []
+        for i, (p, expected) in enumerate(
+            zip(positional_params, ("inputs", "attachments"))
+        ):
+            if p.name != expected:
```
baskaryan, line 1076: wonder if we should only enforce this if p.default is not p.empty (ie if it doesn't have a default value). if it does have a default value then its technically fine that it has other positional args

python/langsmith/client.py
```diff
@@ -3493,6 +3490,56 @@ def upsert_examples_multipart(
...
+                "headers": {
+                    **self._headers,
+                    "Content-Type": encoder.content_type,
+                },
+            },
+        )
+        ls_utils.raise_for_status_with_text(response)
+        return response.json()
+
+    def upsert_examples_multipart(
```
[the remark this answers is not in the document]
isahers1: I don't know what version to mark it deprecated since

python/langsmith/_internal/_operations.py
```diff
@@ -256,17 +257,22 @@ def serialized_run_operation_to_multipart_parts_and_context(
                 )
                 continue
 
-            acc_parts.append(
-                (
-                    f"attachment.{op.id}.{n}",
+            if isinstance(valb, Path):
+                # TODO: actually deal with this case
+                # This is just for speed of getting something out
```
agola11, line 262: why did you need to change this method?

agola11 (approved):

PR #1218, opened 2024-11-15
make evaluate function compatible with attachments

isahers1:


python/langsmith/client.py
```diff
@@ -3725,11 +3727,20 @@ def list_examples(
             params["dataset"] = dataset_id
         else:
             pass
+        if get_attachment_urls == True:
+            params['select'] = ['attachment_urls', 'outputs'] # don't need to add other stuff, because of always_select variable in API endpoint
```
agola11, line 3731: you need to explicitly select metadata as well

python/langsmith/client.py
```diff
@@ -3725,11 +3727,20 @@ def list_examples(
...
         for i, example in enumerate(
             self._get_paginated_list("/examples", params=params)
         ):
+            attachment_urls = {}
+            if example['attachment_urls']:
+                for key, value in example['attachment_urls'].items():
+                    attachment_urls[key.split(".")[1]] = value['presigned_url']
```
agola11, line 3738: what's the point of doing this? also need to fail gracefully if [1] causes an index out of bounds error

python/langsmith/evaluation/_runner.py
```diff
@@ -109,6 +109,7 @@ def evaluate(
     client: Optional[langsmith.Client] = None,
     blocking: bool = True,
     experiment: Optional[Union[schemas.TracerSession, str, uuid.UUID]] = None,
+    with_attachments: bool = False,
```
agola11, line 112: need to add a comment about this arg in the docstring below

python/langsmith/evaluation/_runner.py
```diff
@@ -1237,6 +1244,8 @@ def with_predictions(
             self._predict, target, max_concurrency=max_concurrency
         )
         r1, r2 = itertools.tee(_experiment_results, 2)
+        # QUESTION: Do we need to propogate with_attachments here?
+        # If we don't then the evaluators won't have access to example.attachments
```
agola11, line 1248: yes we should

python/langsmith/evaluation/_runner.py
```diff
@@ -1625,23 +1637,43 @@ def _get_run(r: rt.RunTree) -> None:
...
+                            **metadata,
+                            "example_version": (
+                                example.modified_at.isoformat()
+                                if example.modified_at
+                                else example.created_at.isoformat()
+                            ),
+                        },
+                        client=client,
+                    ),
+                )
```
agola11, line 1658: what happens if `with_attachment` is `True` but the caller sends in a function that doesn't take in the inputs and attachments as arguments? There needs to be some validation as high up into the call path as possible to fail fast.

agola11 (commented):
Some comments, need to add some tests as well

python/langsmith/client.py
isahers1: All the attachments appear as "attachment.name", can keep as is if you want (i.e. don't do the splitting and always include the "attachment." prefix.

isahers1: I believe this is from `ATTACHMENT_S3_PREFIX = "attachment."` line in `ingest.py`

python/langsmith/evaluation/_runner.py
```diff
@@ -1625,23 +1628,43 @@ def _get_run(r: rt.RunTree) -> None:
...
```
baskaryan, line 1667: can we just do
```python
if with_attachments:
  args = (example.inputs, example.attachments)
else:
  args = (example.inputs,)
fn(*args, langsmith_extra=...)

File: python/langsmith/evaluation/_runner.py [hunk 7]
```diff
@@ -1323,14 +1323,15 @@ def get_summary_scores(self) -> Dict[str, List[dict]]:
     # Private methods
 
     def _predict(
-        self, target: TARGET_T, /, max_concurrency: Optional[int] = None
+        self, target: TARGET_T, /, max_concurrency: Optional[int] = None,
     ) -> Generator[_ForwardResults, None, None]:
         """Run the target function on the examples."""
-        fn = _ensure_traceable(target)
+        fn, with_attachments = _ensure_traceable(target)
```
baskaryan, line 1329: can we make the attachments check its own function? doesnt' seem very related to ensure_traceable

baskaryan: also out of curiosity if someone does
```python
@traceable
def target(inputs, attachments): ...
 ```
are we properly tracing attachments? or are they going to be traced as part of the regular run inputs?

File: python/langsmith/evaluation/_runner.py (hunk 6 above)
Comment (isahers1) [reply python/langsmith/evaluation/_runner.py:1667]: yes I am just dumb 🤦‍♂️

File: python/langsmith/evaluation/_runner.py (hunk 7 above)
Comment (isahers1) [reply python/langsmith/evaluation/_runner.py:1329]: I think they are being traced correctly? I ran some experiments and the attachments show up in the run, not sure if that is what you mean

File: python/langsmith/evaluation/_runner.py [hunk 8]
```diff
@@ -1653,13 +1662,14 @@ def _get_run(r: rt.RunTree) -> None:
 
 
 def _resolve_data(
-    data: DATA_T, *, client: langsmith.Client
+    data: DATA_T, *, client: langsmith.Client,
 ) -> Iterable[schemas.Example]:
     """Return the examples for the given dataset."""
+    # TODO: Find a smarter way of determining whether to get attachments (don't just default to true)
```
isahers1, line 1668: would appreciate suggestions. I don't want to pass the target function to the ExperimentManager, but not sure what the best way to do this is

python/langsmith/evaluation/_runner.py
```diff
@@ -1678,12 +1688,31 @@ def _ensure_traceable(
...
+) -> bool:
+    """Whether the target function accepts attachments."""
+    # Check function signature
+    sig = inspect.signature(target)
+    params = list(sig.parameters.values())
+    positional_params = [p for p in params if p.kind in (p.POSITIONAL_ONLY, p.POSITIONAL_OR_KEYWORD)]
+    
+    if len(positional_params) == 0:
+        raise ValueError("Target function must accept at least one positional argument (inputs)")
+    elif len(positional_params) > 2:
```
agola11, line 1709: @hinthornw @baskaryan just calling out that _currently_, an experiment still gets created but none of the runs succeed with >1 positional argument is supplied. so this isn't exactly a breaking change but will change behavior, I think for the better. LMK if you have strong opinions though

python/langsmith/client.py
```diff
@@ -3725,11 +3729,28 @@ def list_examples(
...
             self._get_paginated_list("/examples", params=params)
         ):
+            attachments = {}
+            if example['attachment_urls']:
+                for key, value in example['attachment_urls'].items():
+                    def create_reader(url=value['presigned_url']):
+                        response = requests.get(url, stream=True)
+                        response.raise_for_status()
+                        # Return the raw response object so caller can handle streaming
+                        return response
```
agola11, line 3744: would move this outside the function/loop and make prefix with an underscore for readability

agola11: have you tried using the reader object?

python/langsmith/schemas.py
```diff
@@ -74,6 +76,7 @@ class ExampleBase(BaseModel):
     inputs: Dict[str, Any] = Field(default_factory=dict)
     outputs: Optional[Dict[str, Any]] = Field(default=None)
     metadata: Optional[Dict[str, Any]] = Field(default=None)
+    attachments: Optional[Dict[str, Tuple[str, Callable[[], requests.Response]]]] = Field(default=None)
```
agola11, line 79: Can we add a comment here to explain this type? It's pretty cumbersome. Also I don't think we need to be augmenting the type of examples here? And the typing certainly shouldn't be requests.Response

python/langsmith/client.py
```diff
@@ -3725,11 +3729,28 @@ def list_examples(
...
             pass
+        if get_attachments == True:
+            params['select'] = ['attachment_urls', 'outputs', 'metadata']
         for i, example in enumerate(
             self._get_paginated_list("/examples", params=params)
         ):
+            attachments = {}
+            if example['attachment_urls']:
+                for key, value in example['attachment_urls'].items():
+                    def create_reader(url=value['presigned_url']):
```
agola11, line 3740: need to annotate this with a return type, likely `typing.BinaryIO`

python/langsmith/client.py
agola11: (don't have to read obvsiouly, just return the reader)

python/langsmith/client.py
```diff
@@ -3725,11 +3729,28 @@ def list_examples(
             params["dataset"] = dataset_id
         else:
             pass
+        if get_attachments == True:
```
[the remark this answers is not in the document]
isahers1: added a test

python/langsmith/client.py
```diff
@@ -3664,6 +3665,7 @@ def list_examples(
         limit: Optional[int] = None,
         metadata: Optional[dict] = None,
         filter: Optional[str] = None,
+        get_attachments: Optional[bool] = True,
```
baskaryan, line 3668: no need to be Optional

python/langsmith/evaluation/_runner.py
```diff
@@ -1655,15 +1672,19 @@ def _get_run(r: rt.RunTree) -> None:
...
+    data: DATA_T,
+    *,
+    client: langsmith.Client,
 ) -> Iterable[schemas.Example]:
     """Return the examples for the given dataset."""
+    # TODO: Find a smarter way of determining whether
+    # to get attachments (don't just default to true)
     if isinstance(data, str):
-        return client.list_examples(dataset_name=data)
+        return client.list_examples(dataset_name=data, get_attachments=True)
```
baskaryan, line 1683: how expensive can getting attachments be? should we make `get_attachments` an arg on resolve_data and turn `self.examples` into a method instead of a property that takes `get_attachments` as well? and then in with _predict and _score we pass get_attachments into self.examples(get_attachments)

python/langsmith/client.py
isahers1: yeah I like include_attachments

python/langsmith/client.py
```diff
@@ -3664,6 +3665,7 @@ def list_examples(
         limit: Optional[int] = None,
         metadata: Optional[dict] = None,
         filter: Optional[str] = None,
+        include_attachments: bool = True,
```
baskaryan, line 3668: mm should this default to False?

python/tests/integration_tests/test_client.py
```diff
@@ -1119,3 +1120,99 @@ def create_encoder(*args, **kwargs):
...
+
+
+@pytest.mark.skip(
+    reason="Need to land https://github.com/langchain-ai/langsmith-sdk/pull/1209 first"
+)
+def test_list_examples_attachments_keys() -> None:
+    """Test list_examples returns same keys with and without attachments."""
+    dataset_name = "__test_list_examples_attachments" + uuid4().hex[:4]
+    langchain_client = Client(
+        api_key="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"
```
baskaryan, line 1132: assume not meant to be checked in?

python/langsmith/schemas.py
```diff
@@ -67,18 +67,35 @@ def my_function(bar: int, my_val: Attachment):
...
     metadata: Optional[Dict[str, Any]] = Field(default=None)
+    attachment_urls: Optional[Dict[str, Tuple[str, BinaryIOLike]]] = Field(default=None)
+    """Dictionary with attachment names as keys and a tuple of the S3 url
+    and a reader of the data for the file."""
 
     class Config:
         """Configuration class for the schema."""
 
         frozen = True
+        arbitrary_types_allowed = True
```
baskaryan, line 98: ooc why's this needed

baskaryan: for BinaryIOLike?

python/langsmith/client.py
```diff
@@ -3713,11 +3715,26 @@ def list_examples(
...
             pass
+        if include_attachments:
+            params["select"] = ["attachment_urls", "outputs", "metadata"]
         for i, example in enumerate(
             self._get_paginated_list("/examples", params=params)
         ):
+            attachment_urls = {}
+            if example["attachment_urls"]:
+                for key, value in example["attachment_urls"].items():
+                    response = requests.get(value["presigned_url"], stream=True)
```
jakerachleff, line 3726: Should we do this in parallel under a semaphore or something? Seems like a lot of data to sequentially wait for

python/langsmith/schemas.py
isahers1: yes

python/langsmith/evaluation/_runner.py
```diff
@@ -1692,6 +1730,38 @@ def _ensure_traceable(
     return fn
 
 
+def _include_attachments(
```
baskaryan, line 1733: should add some unit tests for this

PR #1219, opened 2024-11-15
nit: stop calling importlib on every api req

jakerachleff:
Each call is like 1-3ms, and we call to get the version as part of every client request's headers. If you're running an eval over a lot of traces (say 10 evaluators on 1k examples), that's slow. Esp if just doing custom code evaluators which should run fast. 

### Flame Graphs before and after
(ran on 2k smallish examples w 10 evaluators)
Screenshot 2024-11-14 at 5 25 43 PM

Screenshot 2024-11-14 at 5 25 53 PM

### DIFF
--- a/python/langsmith/__init__.py
+++ b/python/langsmith/__init__.py
@@ -1,8 +1,7 @@
 """LangSmith Client."""
 
-from typing import TYPE_CHECKING, Any
-
 from importlib import metadata
+from typing import TYPE_CHECKING, Any
 
 if TYPE_CHECKING:
     from langsmith._expect import expect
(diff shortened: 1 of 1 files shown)


python/langsmith/__init__.py
```diff
@@ -21,15 +22,17 @@
         ContextThreadPoolExecutor,
     )
 
+# Avoid calling into importlib on every call to __version__
+version = ""
+try:
+    metadata.version(__package__)
```
hinthornw, line 28: need to assign this

hinthornw (approved):

PR #1223, opened 2024-11-18
orjson optional wip

efriis:


python/langsmith/_internal/_background_thread.py
```diff
@@ -155,13 +155,24 @@ def tracing_control_thread_func(client_ref: weakref.ref[Client]) -> None:
...
+        # if `client.cleanup()` was called, stop thread
+        if client and client._manual_cleanup:
+            return False
+        if not threading.main_thread().is_alive():
+            # main thread is dead. should not be active
+            return False
+        try:
+            # check if client refs count indicates we're the only remaining
+            # reference to the client
+            return sys.getrefcount(client) > num_known_refs + len(sub_threads)
```
hinthornw, line 168: Could we cache whether this exists so we don't have to continuously hit this error ?

hinthornw (commented):
one comment but looks painfuily acceptable otherwise

python/langsmith/_internal/_background_thread.py
efriis: changed to `hasattr`

hinthornw (approved):

PR #1224, opened 2024-11-18
python[patch]: `evaluate` local mode

baskaryan:
Adds an `upload_results` flag to avoid tracing any runs (target or evaluator) or creating an experiment in langsmith:

```python

results = evaluate(
    lambda x: x, 
    data="Sample Dataset 3", 
    evaluators=[lambda inputs: {"score": 1, "key": "correct"}],
    upload_results=False
)
```


python/tests/unit_tests/test_run_helpers.py
```diff
@@ -1201,14 +1201,14 @@ def my_run(foo: str):
 
     my_run(foo="bar", langsmith_extra={"parent": headers, "client": mock_client})
     mock_calls = _get_calls(mock_client)
-    assert len(mock_calls) == 1
```
baskaryan, line 1204: @hinthornw not sure i understand this behavior, do we need/want to maintain it?

python/langsmith/evaluation/_runner.py
```diff
@@ -1438,7 +1460,7 @@ def _score(
         (e.g. from a previous prediction step)
         """
         with ls_utils.ContextThreadPoolExecutor(
-            max_workers=max_concurrency
+            max_workers=max_concurrency or 1
```
jakerachleff, line 1463: This will change default behavior, no? Do we want to do in a separate PR?

python/langsmith/run_helpers.py
```diff
@@ -935,8 +938,9 @@ def _setup(self) -> run_trees.RunTree:
                 attachments=self.attachments or {},
             )
 
-        if enabled:
+        if enabled is True:
             self.new_run.post()
+        if enabled:
```
jakerachleff, line 943: maybe a dumb question, but why still set all of this?

jakerachleff: ah, intermediate steps

python/langsmith/evaluation/_arunner.py
```diff
@@ -83,6 +83,7 @@ async def aevaluate(
     client: Optional[langsmith.Client] = None,
     blocking: bool = True,
     experiment: Optional[Union[schemas.TracerSession, str, uuid.UUID]] = None,
+    upload_results: bool = True,
```
agola11, line 86: flyby comment: the fact that we have to update so many call sites feels pretty problematic

python/langsmith/evaluation/_arunner.py
```diff
@@ -675,7 +685,7 @@ async def _arun_evaluators(
                 **current_context,
                 "project_name": "evaluators",
                 "metadata": metadata,
-                "enabled": True,
+                "enabled": "local" if not self._upload_results else True,
```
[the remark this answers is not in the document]
agola11: In that case, where do they get stored?

python/langsmith/evaluation/_runner.py
```diff
@@ -182,6 +183,7 @@ def evaluate(
         ...     summary_evaluators=[precision],
         ...     experiment_prefix="My Experiment",
         ...     description="Evaluating the accuracy of a simple prediction model.",
+        ...     max_concurrency=0,
```
agola11, line 186: what's the rationale here?

python/langsmith/evaluation/_arunner.py
```diff
@@ -770,16 +783,17 @@ async def _aapply_summary_evaluators(
...
+                        for result in flattened_results:
+                            feedback = result.dict(exclude={"target_run_id"})
+                            evaluator_info = feedback.pop("evaluator_info", None)
+                            await aitertools.aio_to_thread(
+                                self.client.create_feedback,
+                                **feedback,
+                                run_id=None,
+                                project_id=project_id,
+                                source_info=evaluator_info,
+                            )
```
agola11, line 796: A couple of comments here:
1. What's the rationale of `await`ing in the for loop over `gather`ing the results with some max concurrency?
2. Second, in most cases now, (when multipart is enabled) we now enqueue the feedback to the TracingQueue for asyncronous background processing. So it really doesn't make sense to keep this logic to run this in an executor? cc @hinthornw for this too

python/langsmith/evaluation/_runner.py
```diff
@@ -1508,16 +1530,17 @@ def _apply_summary_evaluators(
...
-                            feedback = result.dict(exclude={"target_run_id"})
-                            evaluator_info = feedback.pop("evaluator_info", None)
-                            executor.submit(
-                                self.client.create_feedback,
-                                **feedback,
-                                run_id=None,
-                                project_id=project_id,
-                                source_info=evaluator_info,
-                            )
+                        if self._upload_results:
```
agola11, line 1533: Same comment as above here

python/langsmith/evaluation/_runner.py
baskaryan: think accidentally committed, will undo

python/langsmith/evaluation/_runner.py
baskaryan: cc @hinthornw, dont think i made this one. was this on purpose?

python/tests/unit_tests/test_run_helpers.py
hinthornw: Ya. At one point what was happening was we'd create a RunTree from headers withoutt he overriding client, and then use that, meaning the user configuration was broken

python/langsmith/evaluation/_arunner.py
hinthornw: Ya we can asyncio.gather()

hinthornw: re: background processing, it doesn't get put in the background queue because we lack a trace_id in this case

python/langsmith/evaluation/_arunner.py
hinthornw: wdym call sites?

Our data model for evals is indeed too complicated lol -tracer sessions, feedback, runs in multiple sessions, ay caramba

hinthornw (approved):
Think this looks good. I do today feel more that upload_results=False shouldn't enforce local mode but rather just shouldn't create a project and then should use the user's environment setup to see whether or not to trace

PR #1231, opened 2024-11-19
support create_examples(..., attachments)

baskaryan:


python/langsmith/client.py
```diff
@@ -3537,12 +3480,37 @@ def create_examples(
...
+                    id=id_ or uuid.uuid4(),
+                )
+                for in_, out_, attach_, metadata_, split_, id_ in zip(
+                    inputs,
+                    outputs or [None] * len(inputs),
+                    attachments,
+                    metadata or [None] * len(inputs),
+                    splits or [None] * len(inputs),
+                    ids or [None] * len(inputs),
+                )
```
isahers1, line 3509: I think this silently cuts off data? Might be good to throw a warning/error if any of inputs/outputs/etc. have different (non-0) lengths, idk

python/tests/unit_tests/test_client.py
```diff
@@ -447,7 +447,7 @@ def test_upsert_examples_multipart(mock_session_cls: mock.Mock) -> None:
             ),
         },
     )
-    client.upsert_examples_multipart(upserts=[example])
+    client._upsert_examples_multipart(upserts=[example])
```
isahers1, line 451: Should we add tests that create_examples works with attachments now?

PR #1235, opened 2024-11-20
fix(js): Fix multipart upload with overridden node fetch

jacoblee93:


js/src/client.ts
```diff
@@ -1150,33 +1150,51 @@ export class Client {
...
         _getFetchImplementation(),
         `${this.apiUrl}/runs/multipart`,
         {
           method: "POST",
           headers: {
             ...this.headers,
+            "Content-Type": `multipart/form-data; boundary=${boundary}`,
           },
-          body: formData,
+          body: arrayBuffer,
```
nfcampos, line 1189: Will fetch automatically set content length header from array buffer length?

js/src/client.ts
```diff
@@ -1150,33 +1150,51 @@ export class Client {
...
+      const chunks: Blob[] = [];
+
       for (const part of parts) {
-        formData.append(part.name, part.payload);
+        // Add field boundary
+        chunks.push(new Blob([`--${boundary}\r\n`]));
+        chunks.push(
+          new Blob([
+            `Content-Disposition: form-data; name="${part.name}"\r\n`,
+            `Content-Type: ${part.payload.type}\r\n\r\n`,
```
nfcampos, line 1164: Missing content length for each part

jacoblee93: No it's set above:

https://github.com/langchain-ai/langsmith-sdk/pull/1235/files#diff-a4957c11dcb337079c4682e6d0069736d061422f086fea164792577b977e7149R1100

js/src/client.ts
jacoblee93: Claude says:

```
Note: In many cases, modern HTTP clients and the Fetch API will automatically set the correct Content-Length header when you provide a body. You could also remove the Content-Length header entirely and let the client handle it automatically.
```

It seems to work in tests with native fetch and Node fetch, and I wasn't doing it before - can manually populate if preferred

js/src/client.ts
nfcampos: Ah got it

nfcampos (approved):

PR #1236, opened 2024-11-20
Bagatur/rfc pred eval inline

baskaryan:


python/langsmith/evaluation/_arunner.py
```diff
@@ -482,8 +492,27 @@ def __init__(
         self._summary_results = summary_results
         self._num_repetitions = num_repetitions
         self._include_attachments = include_attachments
+        self._store_results = store_results
+
+    @property
+    def _example_count(self) -> int:
```
jakerachleff, line 498: we have a count endpoint now you could prob use for this?

python/langsmith/evaluation/_arunner.py
```diff
@@ -635,6 +713,45 @@ async def aget_summary_scores(self) -> Dict[str, List[dict]]:
...
+                        **forward_result,
+                        "evaluation_results": {"results": []},
+                    }
+                    experiment_row = self._arun_evaluators(
+                        evaluators, current_results, executor
+                    )
+                    yield experiment_row
+
+            async for result in aitertools.aiter_with_concurrency(
+                max_concurrency, predict_all(), _eager_consumption_timeout=0.001
```
jakerachleff, line 749: Is there a way to push this concurrency down to the predict_all layer, so we grab an example, then concurrently run over each example? Does the `aforward` and `arun_evaluators` get run sequentially per thread?

PR #1237, opened 2024-11-20
add attachments to evaluate

isahers1:


python/tests/integration_tests/test_client.py
```diff
@@ -1160,20 +1157,18 @@ def test_list_examples_attachments_keys(langchain_client: Client) -> None:
...
 
-@pytest.mark.skip(
-    reason="Need to land https://github.com/langchain-ai/langsmith-sdk/pull/1209 first"
-)
 def test_evaluate_with_attachments(langchain_client: Client) -> None:
     """Test evaluating examples with attachments."""
     dataset_name = "__test_evaluate_attachments" + uuid4().hex[:4]
-    # 1. Create dataset
+    langchain_client = Client(
+        api_key="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",
```
hinthornw, line 1164: You ought to delete this key now

isahers1:
need to add to summary evaluators as well I think

python/tests/integration_tests/test_client.py
```diff
@@ -1381,6 +1370,106 @@ def evaluator(run: Run, example: Example) -> Dict[str, Any]:
...
+        return {"answer": "4"}
+
+    async def evaluator(run: Run, example: Example) -> Dict[str, Any]:
+        return {
+            "score": float(
+                run.outputs.get("answer") == example.outputs.get("answer")  # type: ignore
+            )
+        }
+
+    results = await aevaluate(
```
baskaryan, line 1464: can now do langchain_client.aevaluate

python/langsmith/client.py
```diff
@@ -140,6 +140,16 @@ class ZoneInfo:  # type: ignore[no-redef]
 URLLIB3_SUPPORTS_BLOCKSIZE = "key_blocksize" in signature(PoolKey).parameters
 
 
+class AutoSeekBytesIO(io.BytesIO):
```
baskaryan, line 143: is this related?

baskaryan (commented):
do we need to inspect the evaluator signature ahead of time so that if user passes in just a dataset name/id, we know if we need to load attachments?

python/langsmith/client.py
isahers1: no I need to remove this, this was to solve the issue that reading from the attachment in the target function made it unaccessible (until you seek back to 0) in the evaluator, but we solved this in a different way

python/langsmith/evaluation/evaluator.py
```diff
@@ -624,7 +624,14 @@ def _normalize_evaluator_func(
...
     Callable[[Run, Optional[Example]], Awaitable[_RUNNABLE_OUTPUT]],
 ]:
-    supported_args = ("run", "example", "inputs", "outputs", "reference_outputs")
+    supported_args = (
+        "run",
+        "example",
+        "inputs",
+        "outputs",
+        "reference_outputs",
+        "attachments",
```
baskaryan, line 633: should we rename input_attachments

agola11 (approved):

python/langsmith/evaluation/_runner.py
```diff
@@ -1913,7 +1914,30 @@ def _ensure_traceable(
...
+        return False
+    return any(
+        any(
+            p.kind in (p.POSITIONAL_ONLY, p.POSITIONAL_OR_KEYWORD)
+            and p.name == "attachments"
+            for p in (
+                inspect.signature(
+                    e.__call__ if hasattr(e, "__call__") else e
+                ).parameters.values()
+                if callable(e) or hasattr(e, "__call__")
```
baskaryan, line 1930: ooc what kinds of objects have `__call__` but aren't callable?

baskaryan (approved):

PR #1238, opened 2024-11-20
sdk patch: raise errors on update/create examples when passing mismatching length sequences

isahers1:


hinthornw (approved):
Could you add a test please :)

PR #1241, opened 2024-11-21
python[patch]: summary evaluator simpler returns

baskaryan:


python/langsmith/evaluation/evaluator.py
```diff
@@ -724,3 +706,90 @@ def wrapper(run: Run, example: Example) -> _RUNNABLE_OUTPUT:
...
+        for pname, p in sig.parameters.items()
+        if p.kind in (p.POSITIONAL_OR_KEYWORD, p.POSITIONAL_ONLY)
+    ]
+    if not positional_args or (
+        not all(pname in supported_args for pname in positional_args)
+        and len(positional_args) != 2
+    ):
+        msg = (
+            f"Invalid evaluator function. Must have at least one positional "
+            f"argument. Supported positional arguments are {supported_args}."
```
jakerachleff, line 765: nit, because I'm dumb, would be nice to pass in the incorrect names to the error message. 

Got confused for a minute too long when I passed `input` instead of `inputs`

jakerachleff (approved):

PR #1247, opened 2024-11-21
LangSmith 0.2.x

jakerachleff:
fmt


hinthornw:
haha we should remove pydantic if we do 0.2.x

python/langsmith/evaluation/_runner.py
```diff
@@ -1597,7 +1600,7 @@ def _score(
         (e.g. from a previous prediction step)
         """
         with ls_utils.ContextThreadPoolExecutor(
-            max_workers=max_concurrency
+            max_workers=max_concurrency or 1
```
baskaryan, line 1603: cc @hinthornw

js/src/tests/evaluate.int.test.ts
```diff
@@ -1219,6 +1223,79 @@ test("evaluate enforces correct evaluator types for comparative evaluation at ru
   ).rejects.toThrow(); // You might want to be more specific about the error message
 });
 
+test("scratch", async () => {
```
baskaryan, line 1226: delete

hinthornw (approved):

PR #1248, opened 2024-11-21
recompile pattern if needed before its used to mask sensitive info

davidx33:
when the processor is initialized, I recompile the pattern if necessary

### DIFF
--- a/python/langsmith/anonymizer.py
+++ b/python/langsmith/anonymizer.py
@@ -90,7 +90,15 @@ class RuleNodeProcessor(StringNodeProcessor):
 
     def __init__(self, rules: List[StringNodeRule]):
         """Initialize the processor with a list of rules."""
-        self.rules = rules
+        self.rules = [
+            {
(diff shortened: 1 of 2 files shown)


python/langsmith/anonymizer.py
```diff
@@ -90,7 +90,15 @@ class RuleNodeProcessor(StringNodeProcessor):
 
     def __init__(self, rules: List[StringNodeRule]):
         """Initialize the processor with a list of rules."""
-        self.rules = rules
+        self.rules = [
+            {
+                "pattern": rule["pattern"]
+                if isinstance(rule["pattern"], re.Pattern)
+                else re.compile(rule["pattern"]),
+                "replace": rule.get("replace"),
```
hinthornw, line 98: Could we just assign to "[redacted]" if not provided here as well?

hinthornw (commented):
Could we add some tests please? :)

hinthornw (approved):

PR #1259, opened 2024-11-26
Ensure the wheel-building workflow runs if it's changed.

obi1kenobi:
Otherwise, PRs that touch *only* the workflow file aren't tested on the workflow, and might break it.


agola11 (approved):

obi1kenobi:
Flaky Python integration test run here.

PR #1264, opened 2024-11-27
js[patch]: simple evaluator args

baskaryan:
js equivalent of https://github.com/langchain-ai/langsmith-sdk/pull/1200


js/src/evaluation/evaluator.ts
```diff
@@ -225,3 +240,45 @@ export class DynamicRunEvaluator<Func extends (...args: any[]) => any>
...
+  const funcStr = func.toString();
+
+  // Check if the function accepts a single object parameter
+  if (funcStr.match(/^\s*(?:async\s+)?(?:function\s*)?\(?\s*{\s*[a-zA-Z_]/)) {
+    return { type: "object" };
+  }
+
+  // Check if the function accepts run and example parameters
+  if (
+    funcStr.match(/^\s*(?:async\s+)?(?:function\s*)?\(?\s*run\s*,\s*example/)
```
jacoblee93, line 277: You definitely don't want to do this in general - will break if someone does something as simple as renames parameters

js/src/evaluation/evaluator.ts
```diff
@@ -110,6 +110,21 @@ export class DynamicRunEvaluator<Func extends (...args: any[]) => any>
       langSmithRunAndExample: { run: Run; example: Example };
     }) => {
       const { run, example } = input.langSmithRunAndExample;
+
+      // Check if the evaluator expects the new argument format
+      const params = getEvaluatorParameters(evaluator);
```
jacoblee93, line 115: I think the better/safer way to do this is to shim it with no source code inspection:

```
return evaluator({
         ...run,
          run,
          example,
          inputs: example?.inputs,
          outputs: run?.outputs,
          referenceOutputs: example?.outputs,,
        }, example);
```

Convention in JS is to ignore extra kwargs and args so it should work without breaking anyone.

Downside is it makes logging a little bit funkier but we can deprecate the old style(?)

jacoblee93: CC @nfcampos I think I heard you should try to avoid `Function.toString()` but I'm not actually sure why (other than the fact that the regexes for some of the below checks should be tweaked)

Perf probably isn't super important given this will only run once

jacoblee93: Googling around and tinkering in my terminal - implementation seems unclear, I think it may have some funny edge cases (at the very least arrow functions would definitely need a separate regex, along with `async`, maybe others I'm not thinking about)

Would prefer to do something like the shim above in order to avoid any potential gotchas

js/src/evaluation/evaluator.ts
```diff
@@ -225,3 +240,45 @@ export class DynamicRunEvaluator<Func extends (...args: any[]) => any>
...
+      params: EvaluatorObjectParams
+    ) => Promise<EvaluationResult | EvaluationResults>);
+
+function getEvaluatorParameters(func: EvaluatorFunction): {
+  type: "tuple" | "object";
+} {
+  const funcStr = func.toString();
+
+  // Check if the function accepts a single object parameter
+  if (funcStr.match(/^\s*(?:async\s+)?(?:function\s*)?\(?\s*{\s*[a-zA-Z_]/)) {
```
jacoblee93, line 271: Would also need to account for at the very least arrow-declared functions `() => {}` and `async () => {}`

js/src/evaluation/evaluator.ts
baskaryan: i think thats fine, in py we expect specific params

jacoblee93: Positional args aren't named in JS

jacoblee93 (approved):

js/src/evaluation/evaluator.ts
```diff
@@ -110,7 +124,18 @@ export class DynamicRunEvaluator<Func extends (...args: any[]) => any>
       langSmithRunAndExample: { run: Run; example: Example };
     }) => {
       const { run, example } = input.langSmithRunAndExample;
-      return evaluator(run, example);
+
+      return evaluator(
+        {
+          ...run,
```
jacoblee93, line 130: Should we just do the arity check for both of these?

Is there a world where there are some crazies out there declaring evaluators like this:

```ts
async (run) => {
}
```

I would prefer the shim in both places unless we want to drop support for the two arg syntax in the long run, in which case we should deprecate the other signature:

```ts
export type RunEvaluatorLike =
 /** @deprecated NOTE */
  | ((
      run: Run,
      example?: Example
    ) => Promise<EvaluationResult | EvaluationResults>)
 /** @deprecated NOTE */
  | ((run: Run, example?: Example) => EvaluationResult | EvaluationResults)
  | ((args: {
      run?: Run;
      example?: Example;
      inputs?: Record<string, any>;
      outputs?: Record<string, any>;
      referenceOutputs?: Record<string, any>;
    }) => EvaluationResult | EvaluationResults)
  | ((args: {
      run?: Run;
      example?: Example;
      inputs?: Record<string, any>;
      outputs?: Record<string, any>;
      referenceOutputs?: Record<string, any>;
    }) => Promise<EvaluationResult | EvaluationResults>);
```

js/src/evaluation/evaluate_comparative.ts
```diff
@@ -306,16 +319,20 @@ export async function evaluateComparative(
...
     const expectedRunIds = new Set(runs.map((r) => r.id));
-    const result = await evaluator(
-      options.randomizeOrder ? shuffle(runs) : runs,
-      example
-    );
+    // Check if evaluator expects an object parameter
+    const result =
+      evaluator.length === 1
+        ? await (evaluator as _ComparativeEvaluator)({
+            runs: options.randomizeOrder ? shuffle(runs) : runs,
```
dqbd, line 329: Worth supporting evaluator(runs)?

PR #1269, opened 2024-11-27
python[patch]: one evaluate fn

baskaryan:


python/langsmith/evaluation/_runner.py
```diff
@@ -103,10 +105,52 @@ def evaluate(
...
+    client: Optional[langsmith.Client] = None,
+    blocking: bool = True,
+    experiment: Optional[EXPERIMENT_T] = None,
+    upload_results: bool = True,
+    **kwargs: Any,
+) -> ComparativeExperimentResults: ...
+
+
+def evaluate(
+    target: Union[TARGET_T, Runnable, EXPERIMENT_T, Tuple[EXPERIMENT_T, EXPERIMENT_T]],
```
jakerachleff, line 135: If we're doing this, we should have _really_ good validation on what evaluators go with what targets. And we should fail early and loudly...

jakerachleff: Also debug logs saying which type of evaluation you're running up front

PR #1277, opened 2024-12-02
js[patch]: simple evaluateComparative evaluator args

baskaryan:


js/src/evaluation/evaluate_comparative.ts
```diff
@@ -69,16 +69,29 @@ async function loadTraces(
   return results;
 }
 
+/** @deprecated Use ComparativeEvaluatorNew instead: (args: { runs, example, inputs, outputs, referenceOutputs }) => ... */
+type ComparativeEvaluatorOld = (
```
jacoblee93, line 73: I think you may still need to export this for TypeScript reasons

jacoblee93: And how about `ComparativeEvaluatorLegacy` for this and `_ComparativeEvaluator` for this new type?

PR #1278, opened 2024-12-03
Clone dataset schema

hinthornw:


python/langsmith/client.py
```diff
@@ -3283,6 +3284,8 @@ def clone_public_dataset(
                 dataset_name=dataset_name,
                 description=ds.description,
                 data_type=ds.data_type or ls_schemas.DataType.kv,
+                inputs_schema=ds.inputs_schema,
```
jakerachleff, line 3287: can we set transformations as well here?

jakerachleff (changes requested):

python/langsmith/client.py
```diff
@@ -2806,22 +2807,34 @@ def create_dataset(
         ----------
         dataset_name : str
             The name of the dataset.
-        description : str or None, default=None
+        description : Optional[str], default=None
             The description of the dataset.
-        data_type : DataType or None, default=DataType.kv
+        data_type : ls_schemas.DataType, default=ls_schemas.DataType.kv
```
jakerachleff, line 2812: note to self to mark this as deprecated

python/langsmith/schemas.py
```diff
@@ -171,6 +171,22 @@ class Config:
         frozen = True
 
 
+DatasetTransformationType = Literal[
```
jakerachleff, line 174: Any thoughts on not making this a literal so it's easier to add new ones server side?

jakerachleff (approved):

python/langsmith/schemas.py
hinthornw: I think since I union with str it at least won't be blocked if you try to say a new one

PR #1292, opened 2024-12-05
Angus/multipart compression

angus-langchain:
=== Experiment 5: Multipart Form Streaming Compression ===

ZSTD:
Original Size: 41.58 MB
Compressed Size: 1.77 MB
Ratio: 23.471
Time: 26.78ms
Serialize Time: 119.19ms
Files Processed: 11

### DIFF
--- a/python/bench/io_compression.py
+++ b/python/bench/io_compression.py
@@ -259,7 +259,69 @@ class CompressionExperiment:
         }
 
         return results
+    
+    def run_multifile_streaming_experiment(self, input_files: List[List[dict]], 
+                                        output_files: List[List[dict]]) -> Dict:
+        """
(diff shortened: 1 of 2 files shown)


python/bench/io_compression.py
```diff
@@ -1,8 +1,29 @@
...
+    def __init__(self, parts: List[Tuple[str, Tuple[None, bytes, str, Dict[str, str]]]], 
+                 context: str):
+        self.parts = parts
+        self.context = context
+
+class SerializedRunOperation:
+    def __init__(self, data: Dict[str, Any]):
+        self.operation = data["operation"]
+        self.id = data["id"]
+        self._none = orjson.dumps(data)
```
agola11, line 21: you need to pop all of the other stuff before serializing this otherwise there is no point

PR #1305, opened 2024-12-07
add update examples multipart

isahers1:


python/langsmith/client.py
```diff
@@ -3577,6 +3592,39 @@ def _prepate_multipart_data(
 
         return encoder, data
 
+    def update_examples_multipart(
+        self,
+        *,
+        updates: List[ls_schemas.ExampleUpdateWithAttachments] = [],
```
agola11, line 3598: do not use `= []` (remember earlier comments)

python/langsmith/client.py
```diff
@@ -3577,6 +3592,39 @@ def _prepate_multipart_data(
 
         return encoder, data
 
+    def update_examples_multipart(
```
agola11, line 3595: `ls_schemas.ExampleUpdateWithAttachments` and `ls_schemas.ExampleUploadWithAttachments` should not have a `dataset_id`. Instead `dataset_id` should be passed as a param to the methods itself. This prevents you from having to check to see if every update/upload has the same `dataset_id`

python/tests/integration_tests/test_client.py
```diff
@@ -1415,3 +1417,292 @@ def test_examples_length_validation(langchain_client: Client) -> None:
...
+        outputs={"answer": "A test image 2"},
+        attachments={
+            "image2": ("image/png", b"fake image data 2"),
+            "extra": ("text/plain", b"extra data"),
+        },
+    )
+    
+    created_examples = langchain_client.upload_examples_multipart(
+        uploads=[example1, example2],
+    )
```
agola11, line 1501: I don't get how these are associated with the dataset

PR #1311, opened 2024-12-09
add repro script

agola11:


python/langsmith/_internal/_operations.py
```diff
@@ -141,6 +141,27 @@ def serialize_run_dict(
         attachments=attachments if attachments is not None else None,
     )
 
+def serialize_run_dict_for_compressed_ingest(
```
agola11, line 144: @angus-langchain ignore this

PR #1314, opened 2024-12-09
Add Rust implementation for Python client's `update_run()` method.

obi1kenobi:
Implement the `update_run()` client method in the Rust PyO3 bindings, and call it from the Python client if the Rust bindings are enabled.

### DIFF
--- a/rust/crates/langsmith-pyo3/src/blocking_tracing_client.rs
+++ b/rust/crates/langsmith-pyo3/src/blocking_tracing_client.rs
@@ -47,8 +47,7 @@ impl BlockingTracingClient {
         Ok(Self { client: Arc::from(client) })
     }
 
-    // N.B.: We use `Py<Self>` so that we don't hold the GIL while running this method.
-    //       `slf.get()` below is only valid if the `Self` type is `Sync` and `pyclass(frozen)`,
+    // N.B.: `slf.get()` below is only valid if the `Self` type is `Sync` and `pyclass(frozen)`,
     //       which is enforced at compile-time.
     pub fn create_run(
         slf: &Bound<'_, Self>,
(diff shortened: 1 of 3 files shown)


rust/crates/langsmith-pyo3/src/blocking_tracing_client.rs
```diff
@@ -47,8 +47,7 @@ impl BlockingTracingClient {
         Ok(Self { client: Arc::from(client) })
     }
 
-    // N.B.: We use `Py<Self>` so that we don't hold the GIL while running this method.
```
obi1kenobi, line 50: The GIL info here was outdated, so I removed it. The current implementation drops the GIL explicitly by calling `allow_threads()`.

rust/crates/langsmith-pyo3/src/py_run.rs
```diff
@@ -46,43 +46,40 @@ impl FromPyObject<'_> for RunCreateExtended {
...
-        });
+        Ok(Self(langsmith_tracing_client::client::RunUpdateExtended {
+            run_update,
+            io,
+            attachments,
+        }))
     }
-
-    Ok(Some(attachments))
 }
```
obi1kenobi, line 83: I just moved this function down, to keep all the type definitions next to each other in the file. Otherwise it remained unchanged.

rust/crates/langsmith-pyo3/src/py_run.rs
```diff
@@ -196,6 +213,45 @@ impl FromPyObject<'_> for RunCommon {
     }
 }
 
+fn extract_attachments(value: &Bound<'_, PyAny>) -> PyResult<Option<Vec<Attachment>>> {
```
obi1kenobi, line 216: This function is unchanged, just moved farther down in the same file.

rust/crates/langsmith-pyo3/src/py_run.rs
```diff
@@ -196,6 +213,45 @@ impl FromPyObject<'_> for RunCommon {
...
+        let value = key_value_pair.get_item(1)?;
+        let value_tuple = value.downcast_exact::<PyTuple>()?;
+        let mime_type_value = value_tuple.get_item(0)?;
+        let bytes_value = value_tuple.get_item(1)?;
+
+        attachments.push(Attachment {
+            // TODO: It's unclear whether the key in the attachments dict is
+            //       the `filename`` or the `ref_name`, and where the other one is coming from.
+            ref_name: key.to_string(),
+            filename: key.to_string(),
```
agola11, line 246: it's the ref_name

agola11 (approved):
added a comment but otherwise lgtm

rust/crates/langsmith-pyo3/src/py_run.rs
obi1kenobi: The attachments dict is `name -> (mime_type, data)`, so is using the key for both `ref_name` and `filename` the right thing to do?

PR #1316, opened 2024-12-10
perf: [LS-2561] Stream compress multipart runs 

angus-langchain:
### Purpose
Compresses multipart runs into a buffer and streams the compressed data to the API. Compatible with https://github.com/langchain-ai/langchainplus/pull/7317


python/langsmith/_internal/_background_thread.py
```diff
@@ -88,6 +92,34 @@ def _tracing_thread_drain_queue(
...
+        if not (client._run_count >= runs_limit or current_size >= max_buffer_size):
+            return None
+
+        # Write final boundary and close compression stream
+        client.compressor_writer.write(f'--{client.boundary}--\r\n'.encode())
+        client.compressor_writer.flush()
+        client.compressor_writer.close()
+
+        client.tracing_queue.seek(0)
+        data = client.tracing_queue.getvalue()
```
angus-langchain, line 113: Could be beneficial to stream the data to the api here vs loading the buffer value into memory

python/langsmith/_internal/_background_thread.py
```diff
@@ -199,6 +231,19 @@ def keep_thread_active() -> bool:
...
+    client = client_ref()
+    if client is None:
+        return
+
+    while True:
+        result = _tracing_thread_drain_compressed_buffer(client)
+        if result is not None:
+            time.sleep(0.150)  # Simulate call to backend
+        else:
+            time.sleep(0.1)  # Avoid busy-waiting if no data ready
```
angus-langchain, line 244: Definitely need to improve this to use timeouts similar to the other thread function, will start implementing today

python/langsmith/_internal/_operations.py
```diff
@@ -271,3 +273,36 @@ def serialized_run_operation_to_multipart_parts_and_context(
         acc_parts,
         f"trace={op.trace_id},id={op.id}",
     )
+
+
+def compress_multipart_parts_and_context(
```
angus-langchain, line 278: I'm encoding the multipart form manually here instead of using the request toolkit so I can add a final boundary when I drain the queue vs adding a final boundary in between each run.

angus-langchain: open to other suggestions as wlel

python/langsmith/_internal/_background_thread.py
```diff
@@ -88,6 +92,34 @@ def _tracing_thread_drain_queue(
     return next_batch
 
 
+def _tracing_thread_drain_compressed_buffer(
+    client: Client,
+    runs_limit: int = 100,
+    max_buffer_size: int = 50 * 1024 * 1024
```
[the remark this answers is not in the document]
angus-langchain: done

python/langsmith/_internal/_background_thread.py
angus-langchain: done

python/langsmith/_internal/_background_thread.py
angus-langchain: done!

python/langsmith/client.py
```diff
@@ -489,6 +501,16 @@ def __init__(
         # Create a session and register a finalizer to close it
         session_ = session if session else requests.Session()
         self.session = session_
+        self.compress_traces = os.getenv("LANGSMITH_COMPRESS_TRACES") == "true"
+        if self.compress_traces:
+            self.boundary = BOUNDARY
+            self.compressor: zstandard.ZstdCompressor = zstandard.ZstdCompressor()
```
agola11, line 507: do we actually use `self.compressor` outside of instantiating `self.compressor_writer`?

python/langsmith/_internal/_operations.py
```diff
@@ -271,3 +273,36 @@ def serialized_run_operation_to_multipart_parts_and_context(
...
+        compressor_writer.write(part_header.encode())
+        
+        if isinstance(data, (bytes, bytearray)):
+            with memoryview(data) as view:
+                chunk_size = 1024 * 1024  # 1MB chunks
+                for i in range(0, len(view), chunk_size):
+                    chunk = view[i:i + chunk_size]
+                    compressor_writer.write(chunk)
+        else:
+            compressor_writer.write(str(data).encode())
```
agola11, line 305: what's the reasoning for this?

python/langsmith/_internal/_operations.py
agola11: I think this makes sense for now

python/langsmith/_internal/_operations.py
```diff
@@ -271,3 +273,36 @@ def serialized_run_operation_to_multipart_parts_and_context(
...
+        
+        # Write part terminator
+        compressor_writer.write(b'\r\n')
```
agola11, line 308: per reading the docs here https://python-zstandard.readthedocs.io/en/latest/compressor.html#zstdcompressionwriter, you'll also want to call `flush` in this method after writing the part terminator, specifically `flush(zstandard.FLUSH_BLOCK)` which seems to be the standard

python/langsmith/_internal/_background_thread.py
```diff
@@ -88,6 +94,42 @@ def _tracing_thread_drain_queue(
...
+        current_size = client.compressed_runs_buffer.tell()
+
+        # Check if we should send now
+        if not (client._run_count >= size_limit or current_size >= size_limit_bytes):
+            return None
+
+        # Write final boundary and close compression stream
+        client.compressor_writer.write(f'--{client.boundary}--\r\n'.encode())
+        client.compressor_writer.flush()
+        client.compressor_writer.close()
```
agola11, line 112: I think you just need to call `close()` here

python/langsmith/_internal/_background_thread.py
```diff
@@ -88,6 +94,42 @@ def _tracing_thread_drain_queue(
...
+                yield chunk
+
+        # Reinitialize for next batch
+        client.compressed_runs_buffer = io.BytesIO()
+        client.compressor = zstd.ZstdCompressor()
+        client.compressor_writer = client.compressor.stream_writer(
+            client.compressed_runs_buffer, closefd=False)
+        client._run_count = 0
+
+        return data_stream()
```
agola11, line 131: this doesn't look right to me. You are returning `data_stream` which operates on the buffer you just reset?

I'd move the existing buffer to a separate variable as soon you finish closing, then reset the buffer, then release the lock

python/langsmith/client.py
angus-langchain: no we don't, removing

python/langsmith/_internal/_operations.py
angus-langchain: my thinking was to stream into the compressor instead of writing all the data at once, can remove this if you think unnecessary

python/langsmith/_internal/_background_thread.py
angus-langchain: yeah good catch, missed this when implementing streaming

python/langsmith/_internal/_background_thread.py
angus-langchain: see above comment

python/langsmith/client.py
```diff
@@ -489,6 +501,17 @@ def __init__(
         # Create a session and register a finalizer to close it
         session_ = session if session else requests.Session()
         self.session = session_
+        self.compress_traces = ls_utils.get_env_var("USE_RUN_COMPRESSION")
+        if self.compress_traces:
+            self.boundary = BOUNDARY
+            self.compressed_runs_buffer: io.BytesIO = io.BytesIO()
+            self.compressor_writer: zstd.ZstdCompressionWriter = zstd.ZstdCompressor(level=3).stream_writer(
```
nfcampos, line 508: I'd probably use the lowest level, we want to minimize cpu usage

python/langsmith/_internal/_background_thread.py
```diff
@@ -88,6 +94,39 @@ def _tracing_thread_drain_queue(
...
+        client.compressor_writer.close()
+
+        filled_buffer = client.compressed_runs_buffer
+
+        client.compressed_runs_buffer = io.BytesIO()
+        client.compressor_writer = zstd.ZstdCompressor(level=3).stream_writer(
+            client.compressed_runs_buffer, closefd=False)
+        client._run_count = 0
+
+    filled_buffer.seek(0)
```
nfcampos, line 119: if we have a buffer we can just pass it directly to requests no? no need for iterator

python/langsmith/client.py
angus-langchain: With `level=1` we're only getting a compression ratio of 6.7 vs 18+, `level=2` puts us around 12. I'm open to experimenting with this to see if we still OOM with `level=1`

hinthornw (commented):
We will need to copy the zstandard's BSD license somewhere:
```
Copyright (c) 2016, Gregory Szorc
All rights reserved.

Redistribution and use in source and binary forms, with or without modification,
are permitted provided that the following conditions are met:

1. Redistributions of source code must retain the above copyright notice, this
list of conditions and the following disclaimer.

2. Redistributions in binary form must reproduce the above copyright notice,
this list of conditions and the following disclaimer in the documentation
and/or other materials provided with the distribution.

3. Neither the name of the copyright holder nor the names of its contributors
may be used to endorse or promote products derived from this software without
specific prior written permission.

THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR
ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```

python/langsmith/_internal/_background_thread.py
```diff
@@ -30,6 +36,8 @@
 
 logger = logging.getLogger("langsmith.client")
 
+HTTP_REQUEST_THREAD_POOL = cf.ThreadPoolExecutor(max_workers=cpu_count() * 3)
```
hinthornw, line 39: Should we delay creation of this until the first run?
Langsmith has a lot of downstream deps (through langchain)  that may not be using tracing


ALso, why cpu count * 3? Is the zstd C lib spinning up a separate process?

python/pyproject.toml
```diff
@@ -37,6 +37,7 @@ requests-toolbelt = "^1.0.0"
 
 # Enabled via `langsmith_pyo3` extra: `pip install langsmith[langsmith_pyo3]`.
 langsmith-pyo3 = { version = "^0.1.0rc2", optional = true }
+zstandard = "^0.23.0"
```
hinthornw, line 40: Looks like they have a conda project so should be good, but we may have to manually update our conda distro for the SDK
https://anaconda.org/conda-forge/langsmith
https://anaconda.org/conda-forge/zstandard

python/langsmith/client.py
```diff
@@ -491,6 +504,20 @@ def __init__(
         # Create a session and register a finalizer to close it
         session_ = session if session else requests.Session()
         self.session = session_
+        self.compress_traces = ls_utils.get_env_var("USE_RUN_COMPRESSION")
```
hinthornw, line 507: Do we need this as well? Seems redundant with there being a buffer defined (if that's what we're checking forf)

python/langsmith/client.py
```diff
@@ -491,6 +504,20 @@ def __init__(
...
+        if self.compress_traces:
+            self._futures: set[cf.Future] = set()
```
hinthornw, line 509: ooc: why can't we re-use the tracing queue property to support both for this? I'm still reading, but we just increased the number of properties by ~30%

python/langsmith/client.py
```diff
@@ -491,6 +504,20 @@ def __init__(
...
+            self._boundary = BOUNDARY
+            self.compressed_runs_buffer: Optional[io.BytesIO] = io.BytesIO()
+            self.compressor_writer: zstd.ZstdCompressionWriter = zstd.ZstdCompressor(
+                level=3, threads=-1
```
hinthornw, line 513: Should this be configurable? LIke `RUN_COMPRESSION={"level": 3", ...}` ?or no ? (i'm less familiar with this)

python/langsmith/client.py
```diff
@@ -1290,6 +1325,19 @@ def create_run(
...
             if self._pyo3_client is not None:
                 self._pyo3_client.create_run(run_create)
+            if self.compressed_runs_buffer is not None:
+                serialized_op = serialize_run_dict("post", run_create)
+                multipart_form = (
+                    serialized_run_operation_to_multipart_parts_and_context(
+                        serialized_op
+                    )
+                )
+                with self._buffer_lock:
```
hinthornw, line 1335: hm.. I'm not sure what way there is around it given that it seems like it's not thread safe, but this is acquiring a lock in the main thread which feels detrimental to application runtime

python/langsmith/client.py
```diff
@@ -1754,12 +1854,24 @@ def update_run(
...
+                    compress_multipart_parts_and_context(
+                        multipart_form, self.compressor_writer, self._boundary
+                    )
+                    self._run_count += 1
```
hinthornw, line 1869: I'd prefer to not track statefulness in this way if we can avoid it. Why isn't this tracked on the buffer?

python/langsmith/_internal/_operations.py
```diff
@@ -271,3 +273,32 @@ def serialized_run_operation_to_multipart_parts_and_context(
...
+
+
+def compress_multipart_parts_and_context(
+    parts_and_context: MultipartPartsAndContext,
+    compressor_writer: zstd.ZstdCompressionWriter,
+    boundary: str,
+) -> None:
+    for part_name, (filename, data, content_type, headers) in parts_and_context.parts:
+        part_header = f"--{boundary}\r\n"
+        part_header += f'Content-Disposition: form-data; name="{part_name}"'
```
hinthornw, line 285: We should collect these formatted parts in a list and do a single "".join() rather than copying the header string so many times

python/langsmith/_internal/_operations.py
```diff
@@ -5,6 +5,8 @@
 import uuid
 from typing import Literal, Optional, Union, cast
 
+import zstandard as zstd
```
hinthornw, line 8: Do we actually want this to be a strict dependency?

python/langsmith/_internal/_background_thread.py
```diff
@@ -88,6 +96,42 @@ def _tracing_thread_drain_queue(
...
+
+        if size_limit is not None and size_limit <= 0:
+            raise ValueError(f"size_limit must be positive; got {size_limit}")
+        if size_limit_bytes is not None and size_limit_bytes < 0:
+            raise ValueError(
+                f"size_limit_bytes must be nonnegative; got {size_limit_bytes}"
+            )
+
+        if (size_limit_bytes is None or current_size < size_limit_bytes) and (
+            size_limit is None or client._run_count < size_limit
```
hinthornw, line 115: Why 100?

python/langsmith/client.py
```diff
@@ -491,6 +504,20 @@ def __init__(
         # Create a session and register a finalizer to close it
         session_ = session if session else requests.Session()
         self.session = session_
+        self.compress_traces = ls_utils.get_env_var("USE_RUN_COMPRESSION")
+        if self.compress_traces:
+            self._futures: set[cf.Future] = set()
+            self._boundary = BOUNDARY
```
hinthornw, line 510: If this is a constant why do we need a whole slot property just for this?

python/langsmith/client.py
```diff
@@ -491,6 +504,20 @@ def __init__(
...
+            self.compressed_runs_buffer: Optional[io.BytesIO] = io.BytesIO()
+            self.compressor_writer: zstd.ZstdCompressionWriter = zstd.ZstdCompressor(
+                level=3, threads=-1
+            ).stream_writer(self.compressed_runs_buffer, closefd=False)
+            self._buffer_lock: threading.Lock = threading.Lock()
+            self._data_available_event = threading.Event()
+            self._run_count: int = 0
```
hinthornw, line 517: I'd prefer if we did not track statefulness as a separate property if we can help it

python/langsmith/_internal/_background_thread.py
```diff
@@ -88,6 +96,42 @@ def _tracing_thread_drain_queue(
...
+        client.compressor_writer.write(f"--{client._boundary}--\r\n".encode())
+        client.compressor_writer.close()
+
+        filled_buffer = client.compressed_runs_buffer
+
+        client.compressed_runs_buffer = io.BytesIO()
+        client.compressor_writer = zstd.ZstdCompressor(
+            level=3, threads=-1
+        ).stream_writer(client.compressed_runs_buffer, closefd=False)
+        client._run_count = 0
```
hinthornw, line 129: Can this just be a preoprty of the buffer object?

python/langsmith/_internal/_background_thread.py
angus-langchain: This is the limit that is currently used. It also performed well in flush time benchmarks.

angus-langchain: It can also be overridden by the batch ingest config

python/langsmith/_internal/_background_thread.py
angus-langchain: The max workers is separate form zstd. The workers handle the requests to the backend. Since this is I/O bound it can be greater than the cpu count.

angus-langchain: An empty ThreadPoolExecutor is relatively lightweight and no actual threads are created until work is submitted. The benefits of lazy creating one seems minimal

python/pyproject.toml
angus-langchain: Thanks for the heads up. Would this come up when creating a new version of the sdk?

python/langsmith/client.py
angus-langchain: fixed

python/langsmith/client.py
angus-langchain: I think it's also better to keep them separate since they are different types and I'd be confused if I started using compression and a property completely changed. Additionally, the different typing caused mypy errors within the client when I tried to keep them in the same property. I've removed a significant amount of other properties that were unnecessary.

python/langsmith/client.py
angus-langchain: I can't see a way around this. If you view the benchmarks in the notion doc the compression minimally affects user perceived time, whereas flush time becomes significantly faster

python/langsmith/client.py
angus-langchain: agreed, fixed

python/langsmith/_internal/_background_thread.py
angus-langchain: fixed

python/langsmith/client.py
angus-langchain: fixed

python/langsmith/client.py
angus-langchain: good idea

python/langsmith/_internal/_operations.py
angus-langchain: made optional

python/langsmith/client.py
angus-langchain: fixed

python/langsmith/client.py
angus-langchain: made optional

python/langsmith/_internal/_operations.py
angus-langchain: fixed

python/pyproject.toml
hinthornw: Ya upon release. It should be straightforward to support just noting!

python/langsmith/_internal/_constants.py
```diff
@@ -1,5 +1,8 @@
+import uuid
+
 _SIZE_LIMIT_BYTES = 20_971_520  # 20MB by default
 _AUTO_SCALE_UP_QSIZE_TRIGGER = 200
 _AUTO_SCALE_UP_NTHREADS_LIMIT = 32
 _AUTO_SCALE_DOWN_NEMPTY_TRIGGER = 4
 _BLOCKSIZE_BYTES = 1024 * 1024  # 1MB
+_BOUNDARY = uuid.uuid4().hex
```
hinthornw, line 8: Does this need to be unique every time? or should we hardcode it here?

hinthornw (approved):

PR #1325, opened 2024-12-11
JS updates to attachment type

isahers1:


js/src/client.ts
```diff
@@ -1139,12 +1140,20 @@ export class Client implements LangSmithTracingClientInterface {
...
                 continue;
               }
-              accumulatedParts.push({
-                name: `attachment.${payload.id}.${name}`,
-                payload: new Blob([content], {
-                  type: `${contentType}; length=${content.byteLength}`,
-                }),
-              });
+              // eslint-disable-next-line no-instanceof/no-instanceof
+              if (content instanceof Blob) {
```
jacoblee93, line 1144: Wait sorry why do we need this?

jacoblee93: Also - can this piggyback off the new method you created below?

isahers1: because attachment could be a Blob now (from you: https://langchain.slack.com/archives/D076HL2E77B/p1733946087301879)

js/src/tests/client.int.test.ts
```diff
@@ -1251,7 +1251,7 @@ test("annotationqueue crud", async () => {
   }
 });
 
-test("upload examples multipart", async () => {
+test.only("upload examples multipart", async () => {
```
jacoblee93, line 1254: Remove focused test

PR #1331, opened 2024-12-12
add path as an option for uploading attachments

isahers1:


python/langsmith/run_trees.py
```diff
@@ -318,7 +318,9 @@ def patch(self) -> None:
         """Patch the run tree to the API in a background thread."""
         if not self.end_time:
             self.end()
-        attachments = self.attachments
+        attachments = {
+            a: v for a, v in self.attachments.items() if isinstance(v, tuple)
+        }
```
isahers1, line 323: This was done strictly for typing

python/langsmith/schemas.py
```diff
@@ -390,6 +392,11 @@ def __repr__(self):
         """Return a string representation of the RunBase object."""
         return f"{self.__class__}(id={self.id}, name='{self.name}', run_type='{self.run_type}')"
 
+    class Config:
+        """Configuration class for the schema."""
+
+        arbitrary_types_allowed = True
+
```
isahers1, line 399: Needed to allow the AttachmentInfo

python/langsmith/_internal/_operations.py
```diff
@@ -256,17 +257,32 @@ def serialized_run_operation_to_multipart_parts_and_context(
...
+                )
+            else:
+                file_path = data
+                file_size = os.path.getsize(file_path)
+                acc_parts.append(
+                    (
+                        f"attachment.{op.id}.{n}",
+                        (
+                            None,
+                            open(file_path, "rb"),  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
isahers1: I don't think so, I believe the multipart endpoint allows a buffer to be passed (I also added a test that should cover this)

python/langsmith/run_trees.py
baskaryan: is v ever not a tuple?

python/langsmith/schemas.py
```diff
@@ -370,7 +368,9 @@ class RunBase(BaseModel):
     tags: Optional[List[str]] = None
     """Tags for categorizing or annotating the run."""
 
-    attachments: Attachments = Field(default_factory=dict)
+    attachments: Union[Attachments, Dict[str, AttachmentInfo]] = Field(
```
baskaryan, line 371: is this related to this pr?

python/langsmith/_internal/_operations.py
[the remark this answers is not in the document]
efriis: This is pretty cve worthy so we should be careful about cases where this is allowed/encouraged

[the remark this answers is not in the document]
efriis: My suggestion would be to require either an environment variable of LANGSMITH_DANGEROUSLY_ALLOW_FILESYSTEM or some client init variable saying this. Otherwise throw an error.

python/langsmith/run_trees.py
efriis: If so might be worth doing in a for loop and raising an error if not. I think would be unexpected. But mypy probably knows better than me, so there might be a case we're not handling here

efriis (commented):
Left comments. I think main issue is the arbitrary file read, which I could be convinced isn't a cve if we feel data exfiltration to langsmith is by design. But it feels moderately easy to construct a payload as a user of an app that would make the server stream /etc/passwd to whichever langsmith url you have, which is unideal.

python/langsmith/schemas.py
isahers1: yeah need this for the changes in read_run

python/langsmith/run_trees.py
isahers1: v is never not a tuple here, this is just because of the union types I believe and it thinks that something could be passed here that is never passed

isahers1 (commented):
Need to add the flag for runs (creation and updating)

python/langsmith/_internal/_operations.py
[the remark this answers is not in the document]
agola11: We need to explicitly `close` the files the are opened here, using some `try, finally` pattern. I would structure the code such that the multipart form is assembled and sent within the same method we are opening the file handles:
```
try:
    # Prepare the files for the multipart request
    # Make the POST request
finally:
    # Explicitly close the files
```

python/tests/integration_tests/test_client.py
```diff
@@ -377,6 +378,41 @@ def test_persist_update_run(langchain_client: Client) -> None:
         langchain_client.delete_project(project_name=project_name)
 
 
+def test_update_run_attachments(langchain_client: Client) -> None:
```
agola11, line 381: would be good to add a test case for `traceable` as well

python/langsmith/schemas.py
```diff
@@ -63,7 +65,7 @@ def my_function(bar: int, my_val: Attachment):
     data: bytes
 
 
-Attachments = Dict[str, Union[Tuple[str, bytes], Attachment]]
+Attachments = Dict[str, Union[Tuple[str, bytes], Attachment, Tuple[str, Path]]]
```
agola11, line 68: It's probably worth making `data` within `Attachment` `data: Union[bytes, Path]`.

python/tests/integration_tests/test_client.py
isahers1: added one in `test_runs` because that is where the previous traceable tests were

python/langsmith/client.py
```diff
@@ -6711,3 +6718,12 @@ def _convert_stored_attachments_to_attachments_dict(data, attachments_key):
...
     return attachments_dict
+
+
+def _close_files(files: List[io.BufferedReader]) -> None:
+    """Close all opened files used in multipart requests."""
+    for file in files:
+        try:
+            file.close()
+        except Exception:
+            pass
```
agola11, line 6729: probably good to log something here

python/langsmith/_internal/_operations.py
```diff
@@ -247,7 +249,7 @@ def serialized_run_operation_to_multipart_parts_and_context(
             ),
         )
     if op.attachments:
-        for n, (content_type, valb) in op.attachments.items():
+        for n, (content_type, data) in op.attachments.items():
```
agola11, line 252: nit: rename variable to `data_or_path`

python/langsmith/_internal/_operations.py
```diff
@@ -257,20 +259,39 @@ def serialized_run_operation_to_multipart_parts_and_context(
...
+                        ),
+                    )
                 )
-            )
-    return MultipartPartsAndContext(
-        acc_parts,
-        f"trace={op.trace_id},id={op.id}",
+            else:
+                file_size = os.path.getsize(data)
+                file = open(data, "rb")
```
agola11, line 276: shouldn't we check `dangerously_use_filesystem` here?

python/langsmith/client.py
```diff
@@ -1958,6 +1978,15 @@ def update_run(
...
         }
         if attachments:
+            for _, attachment in attachments.items():
+                if (
+                    isinstance(attachment, tuple)
+                    and isinstance(attachment[1], Path)
+                    and not dangerously_allow_filesystem
+                ):
+                    raise ValueError(
+                        "Must set dangerously_allow_filesystem=True to allow filesystem attachments."
```
agola11, line 1988: `to allow passing in Paths for attachments`

python/langsmith/client.py
```diff
@@ -7207,3 +7254,27 @@ def convert_prompt_to_anthropic_format(
         return anthropic._get_request_payload(messages, stop=stop)
     except Exception as e:
         raise ls_utils.LangSmithError(f"Error converting to Anthropic format: {e}")
+
+
+def _convert_stored_attachments_to_attachments_dict(data, attachments_key):
+    attachments_dict = {}
```
agola11, line 7260: add a comment for the method. Would make `attachments_key` a kwarg to make downstream code more readable

python/langsmith/_internal/_operations.py
isahers1: We check it before hand (in client.py). Everywhere this function gets called, `dangerously_use_filesystem` has already been checked (you should verify this though, I have double checked and am 99% certain but another pass is a good idea)

agola11 (approved):

python/langsmith/run_helpers.py
```diff
@@ -310,6 +311,8 @@ def traceable(
             Defaults to None.
         process_outputs: Custom serialization / processing function for outputs.
             Defaults to None.
+        dangerously_allow_filesystem: Whether to allow filesystem access for attachments.
+            Defaults to False.
```
efriis, line 315:
```suggestion
        dangerously_allow_filesystem: Whether to allow filesystem access for attachments.
            Defaults to False.
            
            Traces that reference local filepaths will be uploaded to LangSmith.
            In general, network-hosted applications should not be using this because
            referenced files are usually on the user's machine, not the host machine.
```

Some clarification like this probably good

efriis (approved):

PR #1334, opened 2024-12-12
Drain the Rust client when its last copy in Python is dropped.

obi1kenobi:
If the Python application wants to exit (or the client just gets garbage collected), we want to finish sending any batched requests before proceeding with the GC or app shutdown.

Normally, Python will make sure to shut down any Python-side threads before allowing the program to exit. But it's completely unaware of Rust's own threads, so it can't shut them down or wait for them. This is why we need to explicitly make sure they complete their work and shut down on their own. The `Drop` trait in Rust runs when the value is dropped, and we ensure we shut down the background threads *only* when we're certain the only `Arc` copy of the Rust client is the one we're holding (and which is about to go out of scope).

### DIFF
--- a/python/bench/tracing_client_via_pyo3.py
+++ b/python/bench/tracing_client_via_pyo3.py
@@ -75,7 +75,7 @@ def benchmark_run_creation(json_size, num_runs) -> None:
 
     if client._pyo3_client:
         # Wait for the queue to drain.
-        client._pyo3_client.drain()
+        del client
     else:
         client.tracing_queue.join()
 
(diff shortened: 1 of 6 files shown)


obi1kenobi:
In an ideal world, we'd rework the `langsmith-tracing-client::client::blocking::TracingClient` to take care of its own draining when it's getting dropped.

But that felt like a much bigger piece of work that will have to touch a lot more code. This PR will fix the immediate problem to get Rust-in-Python unblocked, while we work on a cleaner long-term fix.

agola11 (approved):

PR #1336, opened 2024-12-13
only pop graph if 'serialized' is not none

isahers1:


python/langsmith/client.py
```diff
@@ -1179,7 +1179,7 @@ def _run_transform(
             ):
                 # Drop completely
                 run_create.pop("serialized", None)
-            else:
+            elif run_create["serialized"] is not None:
```
hinthornw, line 1182:
```suggestion
            elif nog run_create.get("serialized")
```

python/langsmith/client.py
```diff
@@ -1179,7 +1179,7 @@ def _run_transform(
             ):
                 # Drop completely
                 run_create.pop("serialized", None)
-            else:
+            elif not run_create.get("serialized"):
```
hinthornw, line 1182: Oops i inverted it - should remove the not - my fault

PR #1340, opened 2024-12-16
[DRAFT] update evaluate to be concurrent

isahers1:


python/langsmith/evaluation/_arunner.py
```diff
@@ -642,6 +651,56 @@ async def astart(self) -> _AsyncExperimentManager:
...
+    ) -> _AsyncExperimentManager:
+        """Run predictions and evaluations in a single pipeline.
+        
+        This allows evaluators to process results as soon as they're available from
+        the target function, rather than waiting for all predictions to complete first.
+        """
+        evaluators = _resolve_evaluators(evaluators)
+        
+        if not hasattr(self, '_evaluator_executor'):
+            self._evaluator_executor = cf.ThreadPoolExecutor(max_workers=4)
```
isahers1, line 669: I don't like doing this, but I couldn't figure out another way to pass an executor that doesn't trigger a `raise RuntimeError('cannot schedule new futures after shutdown')` from `_log_evaluation_feedback`. I also don't think I ever delete this executor, but I am not 100% sure where to do so

python/langsmith/evaluation/_arunner.py
```diff
@@ -839,15 +900,8 @@ async def _arun_evaluators(
...
                     )
-                    logger.error(
-                        f"Error running evaluator {repr(evaluator)} on"
-                        f" run {run.id}: {repr(e)}",
-                        exc_info=True,
-                    )
-                if example.attachments is not None:
-                    for attachment in example.attachments:
-                        reader = example.attachments[attachment]["reader"]
-                        reader.seek(0)
```
isahers1, line 850: This no longer works because the evaluators run in parallel. I do not know the ideal solution to this, am open to any and all ideas.

python/langsmith/evaluation/_arunner.py
```diff
@@ -796,15 +855,17 @@ async def _arun_evaluators(
             run = current_results["run"]
             example = current_results["example"]
             eval_results = current_results["evaluation_results"]
-            for evaluator in evaluators:
+            lock = asyncio.Lock()
```
isahers1, line 858: I don't think it's necessary, can remove if wanted

python/langsmith/evaluation/_arunner.py
```diff
@@ -491,15 +491,24 @@ async def _aevaluate(
...
-            manager = await manager.awith_evaluators(
-                evaluators, max_concurrency=max_concurrency
-            )
-        if summary_evaluators:
-            manager = await manager.awith_summary_evaluators(summary_evaluators)
+            if evaluators:
+                # Run predictions and evaluations in a single pipeline
+                manager = await manager.awith_predictions_and_evaluators(
+                    cast(ATARGET_T, target), evaluators, max_concurrency=max_concurrency
+                )
```
isahers1, line 498: summary evaluators are still evaluated after all the predictions and evaluations have been made, can change in the future but I think much less of a bottle neck.

python/langsmith/evaluation/_arunner.py
eyurtsev: if not necessary remove -- is this used only for .extend()? (I believe .extend is atomic)

https://docs.python.org/3/faq/library.html#what-kinds-of-global-value-mutation-are-thread-safe

python/tests/integration_tests/test_client.py
```diff
@@ -1482,21 +1482,27 @@ def evaluator(run: Run, example: Example) -> Dict[str, Any]:
 async def test_aevaluate_with_attachments(langchain_client: Client) -> None:
     """Test evaluating examples with attachments."""
     dataset_name = "__test_aevaluate_attachments" + uuid4().hex[:4]
+    langchain_client = Client(
+        api_key="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"
```
eyurtsev, line 1486: api key!!

PR #1345, opened 2024-12-19
python[minor] update evaluate to be concurrent

isahers1:


python/langsmith/evaluation/_arunner.py
```diff
@@ -642,6 +651,61 @@ async def astart(self) -> _AsyncExperimentManager:
             upload_results=self._upload_results,
         )
 
+    async def awith_predictions_and_evaluators(
```
hinthornw, line 654: We could probably do something similar to what we do in the sync version to avoid having to duplicate logic here (basically share a semaphor)

python/langsmith/evaluation/_arunner.py
```diff
@@ -642,6 +651,61 @@ async def astart(self) -> _AsyncExperimentManager:
...
+    ) -> _AsyncExperimentManager:
+        """Run predictions and evaluations in a single pipeline.
+
+        This allows evaluators to process results as soon as they're available from
+        the target function, rather than waiting for all predictions to complete first.
+        """
+        evaluators = _resolve_evaluators(evaluators)
+
+        if not hasattr(self, "_evaluator_executor"):
+            self._evaluator_executor = cf.ThreadPoolExecutor(max_workers=4)
```
baskaryan, line 669: ooc where's the 4 come from?

python/langsmith/evaluation/_arunner.py
```diff
@@ -796,15 +860,20 @@ async def _arun_evaluators(
...
                     evaluator_response = await evaluator.aevaluate_run(
                         run=run,
                         example=example,
                     )
-                    eval_results["results"].extend(
-                        self.client._select_eval_results(evaluator_response)
+                    selected_results = self.client._select_eval_results(
+                        evaluator_response
                     )
+                    async with lock:
```
baskaryan, line 874: could we just return the selected_results in _run_single_evaluator and construct the eval_results after the asycio.gather? to avoid needing to lock?

python/langsmith/evaluation/evaluator.py
```diff
@@ -666,7 +667,17 @@ async def awrapper(
                     "example": example,
                     "inputs": example.inputs if example else {},
                     "outputs": run.outputs or {},
-                    "attachments": example.attachments or {} if example else {},
+                    "attachments": (
+                        {
+                            name: {
+                                "presigned_url": value["presigned_url"],
+                                "reader": io.BytesIO(value["reader"].getvalue()),
```
baskaryan, line 674: would love @agola11's input on this bit

python/langsmith/evaluation/_arunner.py
isahers1: I copied the value from `_ascore` - not really sure beyond that

python/langsmith/evaluation/_arunner.py
isahers1: should be fixed, but someone should check I did it correctly

python/langsmith/evaluation/_arunner.py
```diff
@@ -560,7 +577,52 @@ def __init__(
...
+            metadata=example.metadata,
+            modified_at=example.modified_at,
+            runs=example.runs,
+            source_run_id=example.source_run_id,
+            attachments=new_attachments,
+            _host_url=example._host_url,
+            _tenant_id=example._tenant_id,
+        )
+
+    def _make_fresh_examples(
```
agola11, line 620: This method seems unnecessary, plus the name is confusing

python/langsmith/evaluation/_arunner.py
```diff
@@ -560,7 +577,52 @@ def __init__(
...
+        new_attachments: dict[str, schemas.AttachmentInfo] = {}
+        for name, attachment in example.attachments.items():
+            if (
+                self._attachment_raw_data_dict is not None
+                and str(example.id) + name in self._attachment_raw_data_dict
+            ):
+                new_attachments[name] = {
+                    "presigned_url": attachment["presigned_url"],
+                    "reader": io.BytesIO(
+                        self._attachment_raw_data_dict[str(example.id) + name]
```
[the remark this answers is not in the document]
isahers1: No, you are correct. io.BytesIO copies the underlying bytes. This is wrong, I am working on a fix rn.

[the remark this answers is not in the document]
isahers1: ehh actually I am going to walk back my statement. based on testing I don't think bytesIO copies the data.

agola11 (approved):

python/tests/unit_tests/evaluation/test_runner.py
```diff
@@ -617,7 +617,6 @@ def summary_eval_outputs_reference(outputs, reference_outputs):
         tolerance = 3
         assert total_slow < tolerance
         assert total_quick > (SPLIT_SIZE * NUM_REPETITIONS - 1) - tolerance
-        assert any([d > 1 for d in deltas])
```
isahers1, line 620: @hinthornw I made this change to pass CI, but I would appreciate your review.

hinthornw: It's basically meant to test that we aren't iterating as two phases predict -> evaluate but instead doing generate -> evaluate as a continuous, eager stream.

python/langsmith/evaluation/_runner.py
```diff
@@ -1332,7 +1333,56 @@ def __init__(
...
+                    "mime_type": attachment["mime_type"],
+                }
+            else:
+                new_attachments[name] = attachment
+
+        # Create a new Example instance with the updated attachments
+        return schemas.Example(
+            id=example.id,
+            created_at=example.created_at,
+            dataset_id=example.dataset_id,
```
hinthornw, line 1375: Seems likely that we'll forget to update this when we add a new field - if we haven't added a test for this in the previous version, would like one

python/langsmith/evaluation/_arunner.py
```diff
@@ -491,15 +497,24 @@ async def _aevaluate(
...
-            )
-        if evaluators:
-            manager = await manager.awith_evaluators(
-                evaluators, max_concurrency=max_concurrency
-            )
-        if summary_evaluators:
-            manager = await manager.awith_summary_evaluators(summary_evaluators)
+            if evaluators:
+                # Run predictions and evaluations in a single pipeline
+                manager = await manager.awith_predictions_and_evaluators(
```
hinthornw, line 502: If predictions are streamed out do we need a separate method?

python/tests/unit_tests/evaluation/test_runner.py
hinthornw: I don't think we should remove this test

PR #1347, opened 2024-12-20
Add inline evaluation results for summary evaluators

isahers1:


python/langsmith/evaluation/_runner.py
```diff
@@ -1439,6 +1433,7 @@ def with_evaluators(
         # Split the generator into three so the manager
         # can consume each value individually.
         r1, r2, r3 = itertools.tee(experiment_results, 3)
+        # print("FOOOO", [result["evaluation_results"] for result in r3])
```
baskaryan, line 1436: delete

python/langsmith/evaluation/_runner.py
```diff
@@ -1793,16 +1794,20 @@ def _wrap(evaluator: SUMMARY_EVALUATOR_T) -> SUMMARY_EVALUATOR_T:
 
         @functools.wraps(evaluator)
         def _wrapper_inner(
-            runs: Sequence[schemas.Run], examples: Sequence[schemas.Example]
+            runs: Sequence[schemas.Run],
+            examples: Sequence[schemas.Example],
+            evaluation_results: Sequence[schemas.ExperimentResultRow],
```
baskaryan, line 1799: should we just pass in the EvaluationResults not the whole row?

python/langsmith/evaluation/evaluator.py
```diff
@@ -839,18 +768,35 @@ def _normalize_summary_evaluator(func: Callable) -> SUMMARY_EVALUATOR_T:
...
+
+        def wrapper(
+            runs: Sequence[schemas.Run],
+            examples: Sequence[schemas.Example],
+            evaluation_results: Sequence[ExperimentResultRow],
+        ) -> Union[EvaluationResult, EvaluationResults]:
+            result = func(runs, examples)
+            if isinstance(result, EvaluationResult):
+                return result
+            return _format_evaluator_result(result)  # type: ignore[return-value]
```
baskaryan, line 780: good catch!

python/langsmith/evaluation/_runner.py
isahers1: I think it's easier to associate with which example the result came from - which I imagine might be useful? Not 100% sure, up to you just lmk what to do

baskaryan (changes requested):
i think current implementation won't work and would require a bigger refactor, currently we don't await all row-level results before running summary evaluators and if we want to implement this feature we'd need to change that

PR #1348, opened 2024-12-20
Add inline evaluation results for summary evaluators

isahers1:


js/src/evaluation/_runner.ts
```diff
@@ -64,6 +62,18 @@ type DeprecatedAsyncSummaryEvaluator = (
   examples: Array<Example>
 ) => Promise<EvaluationResult | EvaluationResults>;
 
+type SyncSummaryEvaluator = (
+  runs: Array<Run>,
+  examples: Array<Example>,
+  evaluationResults: Array<ExperimentResultRow>
+) => EvaluationResult | EvaluationResults;
+
+type AsyncSummaryEvaluator = (
```
jacoblee93, line 71: Why is this different from `SummaryEvaluatorT`?

js/src/evaluation/_runner.ts
```diff
@@ -133,13 +145,17 @@ interface _ExperimentManagerArgs {
   runs?: AsyncGenerator<Run>;
   evaluationResults?: AsyncGenerator<EvaluationResults>;
   summaryResults?: AsyncGenerator<
-    (runsArray: Run[]) => AsyncGenerator<EvaluationResults, any, unknown>,
+    (
```
jacoblee93, line 148: This is a weird typing but I guess fine

jacoblee93 (commented):
This is allowing people to pass results into an evaluator then?

I don't quite understand the use case

js/src/evaluation/_runner.ts
isahers1: This is what we wrap the evaluator to internally, I do not believe it is user facing.

js/src/evaluation/_runner.ts
isahers1: Ah yeah - can I just make evaluationResults an optional param? Or is there a better solution to make this non breaking?

PR #1354, opened 2024-12-27
Add python annotation queue async methods

MarkRx:
* Adds annotation queue methods to the AsyncClient
* Adds delete_feedback async method to AsyncClient
* Fixes 422 due to httpx not filtering out params with None values (the requests library used by the normal Client does)


MarkRx:
Merged under PR https://github.com/langchain-ai/langsmith-sdk/pull/1396

PR #1358, opened 2024-12-29
feat: Make size_limit (max run count per batch) globally configurable via env var

angus-langchain:
### Purpose
Compression improves run throughput and memory usage, but performance is still slower than expected for power users due to excessive threads and small payloads. SDK users can improve the throughput for larger payloads and higher trace counts by making the run count batch ingest config globally configurable.

### Changes
* Created an environment variable `LANGSMITH_BATCH_INGEST_SIZE_LIMIT` to configure run count.

* This takes precedence over the `/info` api response and any value passed directly to the client batch ingest config.

### Tests / Benchmarks
Setting a higher run count increased compressed runs throughput by **~80-100%** depending on run count and payload size.

### DIFF
--- a/python/langsmith/_internal/_background_thread.py
+++ b/python/langsmith/_internal/_background_thread.py
@@ -18,6 +18,8 @@ from typing import (
 )
 
 from langsmith import schemas as ls_schemas
+from langsmith import utils as ls_utils
+
 from langsmith._internal._constants import (
     _AUTO_SCALE_DOWN_NEMPTY_TRIGGER,
     _AUTO_SCALE_UP_NTHREADS_LIMIT,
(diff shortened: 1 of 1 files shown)


hinthornw:
Should we prefix with "LANGSMITH_" so it gets logged automatically and so it's in our naespace?

angus-langchain:
> Should we prefix with "LANGSMITH_" so it gets logged automatically and so it's in our naespace?

@hinthornw ls_utils.get_env_var() automatically adds the prefix as well as “LANGCHAIN”.

hinthornw:
I'm saying mainly so it doesn't have collisions with any other package

hinthornw (approved):
Oops i should always just read the code

PR #1373, opened 2025-01-03
feat: Instrument wrapper for OpenAI structured outputs

angus-langchain:
### Purpose
Instrument tracing for OpenAI Structured Outputs
### Changes
1. Created openai parser wrapper for structured outputs beta method `client.beta.chat.completions.parse`
1. Added support for both sync and async tracing
1. Maintains backwards compatibility for clients that don't have beta parse methods
### Tests
Added unit tests for parsing methods, and token counts. Tested locally to confirm parsing gets traced
Screenshot 2024-12-02 at 9 45 16 PM

> [!NOTE]  
> Parsing does not support streaming, so the `reduce_fn` is always set to None

### DIFF
--- a/python/langsmith/wrappers/_openai.py
+++ b/python/langsmith/wrappers/_openai.py
@@ -254,6 +254,44 @@ def _get_wrapper(
     return acreate if run_helpers.is_async(original_create) else create
 
 
+def _get_parse_wrapper(
+    original_parse: Callable,
+    name: str,
+    tracing_extra: Optional[TracingExtra] = None,
+    invocation_params_fn: Optional[Callable] = None,
+) -> Callable:
+    textra = tracing_extra or {}
+
+    @functools.wraps(original_parse)
+    def parse(*args, **kwargs):
+        decorator = run_helpers.traceable(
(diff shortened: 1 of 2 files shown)


python/langsmith/wrappers/_openai.py
```diff
@@ -254,6 +254,44 @@ async def acreate(*args, stream: bool = False, **kwargs):
...
+        )
+        return decorator(original_parse)(*args, **kwargs)
+
+    @functools.wraps(original_parse)
+    async def aparse(*args, **kwargs):
+        kwargs = _strip_not_given(kwargs)
+        decorator = run_helpers.traceable(
+            name=name,
+            run_type="llm",
+            reduce_fn=None,
```
hinthornw, line 284: What happens if they try to stream the output?

hinthornw (approved):
Should we share more code btwn this and the other function?

angus-langchain:
> Should we share more code btwn this and the other function?

I prefer to have them separate, keeps the code more readable and each wrapper had a specific purpose. But I'm open to hot takes here

PR #1376, opened 2025-01-06
feat(js): Add method to delete a run from an annotation queue

tendev-liam:


jacoblee93 (approved):
Thank you for this!

Looks good to me! CC @hinthornw

PR #1383, opened 2025-01-07
fix(vercel): nested runs being unflattened, incorrect inheritance behavior of user runId / traceable

dqbd:
Fixes https://github.com/vercel/ai/issues/4223


js/src/vercel.ts
```diff
@@ -183,11 +183,17 @@ function stripNonAlphanumeric(input: string) {
   return input.replace(/[-:.]/g, "");
 }
 
-function convertToDottedOrderFormat(
-  [seconds, nanoseconds]: [seconds: number, nanoseconds: number],
-  runId: string,
-  executionOrder: number
-) {
+function getDotOrder(item: {
```
jacoblee93, line 186: Super nit: why change the name?

js/src/vercel.ts
```diff
@@ -202,6 +208,59 @@ function convertToDottedOrderFormat(
   );
 }
 
+function joinDotOrder(...segments: (string | undefined | null)[]): string {
+  return segments.filter(Boolean).join(".");
```
jacoblee93, line 212: Assuming this is equivalent to `s != null`

js/src/vercel.ts
```diff
@@ -797,83 +836,74 @@ export class AISDKExporter {
...
-          let override: OverrideRunCreate = {
-            id: task.item.id,
-            parent_run_id: task.item.parentId,
-            dotted_order: task.dottedOrder,
-            trace_id: task.traceId,
-          };
+          if (task.item.run != null) {
+            sampled.push({ dotOrder: task.dotOrder, run: task.item.run });
+
+            if (task.item.interop?.type === "user") {
```
jacoblee93, line 842: What does `type: "user"` represent again?

js/src/vercel.ts
```diff
@@ -797,83 +836,74 @@ export class AISDKExporter {
...
+            dotOrder: joinDotOrder(task.dotOrder, getDotOrder(item)),
+          }))
         );
       }
     }
 
-    this.logDebug(`sampled runs to be sent to LangSmith`, sampled);
+    let processed = sampled;
+
+    for (const action of actions) {
```
jacoblee93, line 876: Why do this in a separate loop instead of inline above?

jacoblee93 (approved):
LGTM if tests pass, have a few questions

js/src/vercel.ts
dqbd: This is set when `runId: string` is passed in `AISDKExporter.getSettings`

js/src/vercel.ts
dqbd: Wanted to directly pass the `RunTask` without the need and changed the name as well to make it shorter 🙂

js/src/vercel.ts
dqbd: Also catches empty segment string, which feels correct to me

js/src/vercel.ts
dqbd: Think you're right, all of these actions are affecting the consequent nodes only, will inline

PR #1384, opened 2025-01-07
feat(python): Enable zstd compression by default

angus-langchain:


python/langsmith/_internal/_background_thread.py
```diff
@@ -214,6 +216,23 @@ def tracing_control_thread_func(client_ref: weakref.ref[Client]) -> None:
...
 
+    if ls_utils.get_env_var("DISABLE_RUN_COMPRESSION") is None and use_multipart:
+        if not (client.info.instance_flags or {}).get(
+            "zstd_compression_enabled", False
+        ):
+            raise ValueError(
+                "Zstd compression is not enabled. Please update to the latest "
+                "version of LangSmith or set the environment variable "
+                "DISABLE_RUN_COMPRESSION=true."
+            )
```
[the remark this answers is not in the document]
agola11: If we throw an error here, it would break a lot of self-hosted deployments

python/langsmith/_internal/_background_thread.py
```diff
@@ -264,12 +284,16 @@ def keep_thread_active() -> bool:
...
 
 def tracing_control_thread_func_compress_parallel(
-    client_ref: weakref.ref[Client],
+    client_ref: weakref.ref[Client], flush_interval: float = 0.5
 ) -> None:
     client = client_ref()
     if client is None:
         return
 
+    assert client.compressed_runs is not None
```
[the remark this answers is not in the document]
angus-langchain: done

python/langsmith/_internal/_background_thread.py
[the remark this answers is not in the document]
angus-langchain: Yeah agreed, changed the slow to fall back to regular tracing

python/langsmith/_internal/_background_thread.py
```diff
@@ -214,6 +216,23 @@ def tracing_control_thread_func(client_ref: weakref.ref[Client]) -> None:
...
+                "version of LangSmith or set the environment variable "
+                "DISABLE_RUN_COMPRESSION=true."
+            )
+        client._futures = set()
+        client.compressed_runs = CompressedRuns()
+        client._data_available_event = threading.Event()
+        threading.Thread(
+            target=tracing_control_thread_func_compress_parallel,
+            args=(weakref.ref(client),),
+        ).start()
```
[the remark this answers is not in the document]
angus-langchain: I think the thread needs to stay active at least until we can drain all the jobs from the tracing queue before we switch to compression. Thinking about the best way to do this

python/langsmith/client.py
```diff
@@ -1293,6 +1278,7 @@ def create_run(
             if self._pyo3_client is not None:
                 self._pyo3_client.create_run(run_create)
             elif self.compressed_runs is not None:
+                assert self._data_available_event is not None
```
hinthornw, line 1281: Could we have useful error messages if these trip?

agola11: agree ^, we should avoid asserts in user-facing code

python/langsmith/_internal/_background_thread.py
```diff
@@ -214,6 +217,23 @@ def tracing_control_thread_func(client_ref: weakref.ref[Client]) -> None:
     scale_up_qsize_trigger: int = batch_ingest_config["scale_up_qsize_trigger"]
     use_multipart = batch_ingest_config.get("use_multipart_endpoint", False)
 
+    if ls_utils.get_env_var("DISABLE_RUN_COMPRESSION") is None and use_multipart:
```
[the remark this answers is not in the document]
agola11: Agree, we should check truthiness of env var, as is conventional in our SDK for other env vars

python/langsmith/_internal/_background_thread.py
```diff
@@ -214,6 +217,23 @@ def tracing_control_thread_func(client_ref: weakref.ref[Client]) -> None:
...
+        if not (client.info.instance_flags or {}).get(
+            "zstd_compression_enabled", False
+        ):
+            logger.warning(
+                "Zstd compression is not enabled. Please update to the latest "
```
agola11, line 225: nit: "Run compression" instead of "Zstd compression" -- no need to share impl details

agola11: Would also state in this message that they are falling back to regular multipart

python/tests/unit_tests/test_client.py
```diff
@@ -2069,3 +2069,74 @@ def test_evaluate_methods() -> None:
     eval_args = set(inspect.signature(aevaluate).parameters).difference({"client"})
     extra_args = client_args - eval_args
     assert not extra_args
+
+
+@patch("langsmith.client.requests.Session")
+def test_create_run_with_zstd_compression(mock_session_cls: mock.Mock) -> None:
```
agola11, line 2075: would also add a test to ensure that the case where the info endpoint does not support `zstd_compression_enabled` uses regular multipart

agola11: Additoinally, need a test case for `DISABLE_RUN_COMPRESSION`

python/langsmith/utils.py
```diff
@@ -795,3 +795,15 @@ def _get_function_name(fn: Callable, depth: int = 0) -> str:
         return _get_function_name(fn.__call__, depth + 1)
 
     return str(fn)
+
+
+def is_truish(val: Any) -> bool:
```
agola11, line 800: or `val == "1"`

agola11 (approved):

python/langsmith/_internal/_background_thread.py
```diff
@@ -214,6 +217,27 @@ def tracing_control_thread_func(client_ref: weakref.ref[Client]) -> None:
...
+        else:
+            client._futures = set()
+            client.compressed_runs = CompressedRuns()
+            client._data_available_event = threading.Event()
+            threading.Thread(
+                target=tracing_control_thread_func_compress_parallel,
+                args=(weakref.ref(client),),
+            ).start()
+
+            # client.tracing_queue.join()
```
agola11, line 238: what was the purpose of this? should clean this up

PR #1398, opened 2025-01-08
Draft of zstd compression in the PyO3 client implementation.

obi1kenobi:
Not yet tested end-to-end. Will continue on this tomorrow -- this is just for initial "vibe check" review purposes.

Will likely need to have its Rust and Python parts merged separately, since the Python client library is pinned to a specific Rust PyO3 client version and we're making a major breaking change to the PyO3 client's interface by adding the compression level argument.


obi1kenobi:
Superseded by #1401 and #1402.

PR #1401, opened 2025-01-09
Implement zstd compression for multipart payloads on the Rust side.

obi1kenobi:
I've tested this with both the Python and Rust changes applied, and I've successfully sent zstd-compressed traces to LangSmith servers.

This introduces breaking changes in the APIs for both `langsmith-tracing-client` and `langsmith-pyo3`, so I'm bumping the left-most non-zero version number in each of their manifests. In Rust, the left-most non-zero number is considered a "major" version -- in other words, leading zeroes are ignored for SemVer purposes.

This will require us to publish a new `langsmith-pyo3` Python package version with the new changes. I'd like to trigger the publish workflow after this is merged, so I can make the corresponding Python changes to enable zstd end-to-end. Lmk if you have concerns or want me to hold off on publishing a new version.

### DIFF
--- a/rust/crates/langsmith-tracing-client/src/client/blocking/processor.rs
+++ b/rust/crates/langsmith-tracing-client/src/client/blocking/processor.rs
@@ -1,6 +1,8 @@
 use std::collections::HashMap;
+use std::io::Write;
 use std::sync::mpsc::{Receiver, Sender};
 use std::sync::{mpsc, Arc, Mutex};
+use std::thread::available_parallelism;
 use std::time::{Duration, Instant};
 
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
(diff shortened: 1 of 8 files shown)


obi1kenobi:
Something to consider in the code review: from the server's perspective, this approach is equivalent to the current pure-Python zstd implementation, but has an internal caveat that may prove to be a problem down the line.

Due to how the `reqwest` APIs work, it's forced to fully materialize the entire multipart payload *in memory* before compressing it. This can lead to unnecessarily high memory use.

In an ideal world, we'd do the compression in a streaming fashion. This would either require some changes to `reqwest` APIs, or would require us to DIY the logic for how multipart requests are constructed instead of using `reqwest` here. We can do that in subsequent PRs if it proves important.

rust/crates/langsmith-tracing-client/src/client/blocking/processor.rs
```diff
@@ -315,3 +321,63 @@ impl RunProcessor {
...
+        .and_then(|mut encoder| {
+            encoder.multithread(n_workers)?;
+            Ok(encoder)
+        })
+        .map_err(|e| TracingClientError::IoError(e.to_string()))?;
+
+    encoder
+        .write_all(multipart_form_bytes)
+        .map_err(|e| TracingClientError::IoError(e.to_string()))?;
+    encoder.finish().map_err(|e| TracingClientError::IoError(e.to_string()))?;
```
obi1kenobi, line 380: This workaround will not be required anymore once the next `reqwest` version is released. This code will be replaced in #1403 which makes use of new `reqwest` functionality to do this properly.

Issue #1399, opened 2025-01-09
Issue: Bug -> Querying Experiment by Dataset Version Error


tendev-eoghan:
### Issue you'd like to raise.

When working on [https://github.com/langchain-ai/langsmith-sdk/pull/1390/](url) I came across the following bug.

`listDatasetVersions` returns versions with 6 digits to represent a fraction of a second. 
e.g. `2025-01-09T14:42:18.364000+00:00`

When calling `listProjects` passing in a datasetVersion [https://github.com/langchain-ai/langsmith-sdk/pull/1391](url) sometimes the datasetVersion returned from `listDatasetVersions` will not match the one tagged to a value from `listProjects`.

This is because `listProjects` sometimes has the datasetVersion with only 3 digits to represent a fraction of a second.
e.g. `2025-01-09T14:42:18.364+00:00`

This issue also presents itself in the langsmith UI portal when filtering experiments by dataset version.

 

### Suggestion:

standardise dataset versions to 6 digits to represent a fraction of a second across all references to dataset version.

tendev-eoghan:
With some further digging I found this function which I think might be the root cause.

[https://github.com/langchain-ai/langsmith-sdk/blob/843d55dd584c307fa9ba338d6135884aed31e201/js/src/evaluation/_runner.ts#L816](https://github.com/langchain-ai/langsmith-sdk/blob/843d55dd584c307fa9ba338d6135884aed31e201/js/src/evaluation/_runner.ts#L816)

Sometimes this function returns 3 digits of precision instead of 6.

tendev-eoghan:
Issue fixed with the above PR by @tendev-liam 

[https://github.com/langchain-ai/langsmith-sdk/pull/1406](https://github.com/langchain-ai/langsmith-sdk/pull/1406)

PR #1406, opened 2025-01-10
fix(js): remove default created_at from createExample method.

tendev-liam:
Fixes issue: [#1399 ](https://github.com/langchain-ai/langsmith-sdk/issues/1399)  

**Issue:**
Examples created from the createExample method of the JS client are created with 3 digits of fractional seconds.

**How Issue Presents:**
The endpoint to get dataset versions ([https://api.smith.langchain.com/api/v1/datasets/{dataset_id}/version](https://api.smith.langchain.com/api/v1/datasets/%7Bdataset_id%7D/version)) always responds with 6 digits of fractional seconds.
When filtering expirements by dataset ID the langsmith frontend will use one of the resuls from the get dataset verions i.e the timestamp with 6 digits.
When a expirement is being tagged with a dataset version it will use the modified_at key of an example which if created from the JS client will only contain 3 digits of precison.

**Fix:**
Hand over the creation of the createExample created_at timestamp to langsmith server.

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -2643,16 +2643,21 @@ export class Client implements LangSmithTracingClientInterface {
       datasetId_ = dataset.id;
     }
 
-    const createdAt_ = createdAt || new Date();
+    // Note: the created at time will be added by the langsmith server when not provided.
+    // This will ensure the timestamp will be of the standard langsmith datetime format.
+    // 2025-01-10T14:43:55.017298+00:00 i.e with 6 digits for fractional seconds.
+    const createdAt_ = createdAt
+      ? new Date(createdAt).toISOString()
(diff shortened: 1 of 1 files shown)


dqbd (approved):

tendev-liam:
I made this PR as an attempted fix for [this issue](https://github.com/langchain-ai/langsmith-sdk/issues/1399) and it appeared to work initially but later as mentioned [in this comment](https://github.com/langchain-ai/langsmith-sdk/issues/1399#issuecomment-2587468412) I think this is just a partial solve.
Can we merge now or do we need to investigate further?

jacoblee93:
Hmm yeah let's try to make a real fix

PR #1408, opened 2025-01-10
feat(js): add dataset version to evaluate response

lliammul:
Add the dataset version to the ExperimentResults so that we can add a tag to the dataset version on completion of the experiment.


tendev-liam:
I've moved the assignment of `dataset_version` from the `_end` function to the `start` in `_ExperimentManager`. 
This ensures that `dataset_version` is set earlier and can now be accessed via a new getter in the ExperimentResults class. Let me know if this approach is okay.

PR #1409, opened 2025-01-11
surface mime type when passing attachments to user

isahers1:


js/src/client.ts
```diff
@@ -2770,6 +2770,7 @@ export class Client implements LangSmithTracingClientInterface {
         (acc, [key, value]) => {
           acc[key.slice("attachment.".length)] = {
             presigned_url: value.presigned_url,
+            mime_type: value.mime_type || undefined,
```
jacoblee93, line 2773: You don't need `|| undefined`, just `value.mime_type`

js/src/tests/evaluate_attachments.int.test.ts
```diff
@@ -40,6 +40,10 @@ test("evaluate can handle examples with attachments", async () => {
     const expectedData = new Uint8Array(
       Buffer.from("fake image data for testing")
     );
+    const attachmentMimeType = config?.attachments?.["image"].mime_type;
```
jacoblee93, line 43: `config?.attachments?.image.mime_type;` is more conventional

For JS you only use box notation like that for things that have characters that can be represented inline e.g. if you were to do:

`config?.attachments?.["foo-bar"].mime_type;`

agola11 (approved):
approving to unblock, but please address Jacob's comments

js/src/tests/evaluate_attachments.int.test.ts
```diff
@@ -34,22 +34,22 @@ test("evaluate can handle examples with attachments", async () => {
...
     }
     const attachmentData: Uint8Array | undefined = config?.attachments?.[
       "image"
     ].presigned_url
       ? new Uint8Array(
-          (await fetch(config?.attachments?.["image"].presigned_url).then(
-            (res) => res.arrayBuffer()
+          (await fetch(config?.attachments?.image.presigned_url).then((res) =>
+            res.arrayBuffer()
           )) as ArrayBuffer
```
jacoblee93, line 53: nit: do you need the cast?

jacoblee93 (approved):
Thanks!

js/src/tests/evaluate_attachments.int.test.ts
isahers1: yes, cast is necessary:

No overload matches this call.
  The last overload gave the following error.
    Argument of type 'Iterable<number> | ArrayBuffer' is not assignable to parameter of type 'ArrayBufferLike'.
      Type 'Iterable<number>' is not assignable to type 'ArrayBufferLike'.ts(2769)
lib.es5.d.ts(2405, 5): The last overload is declared here.

PR #1410, opened 2025-01-12
feat(js): add readExamplesWithRuns method for fetching examples w/ runs

tendev-eoghan:
Also made dataKey param optional in `_getCursorPaginatedList` so that it can be used in multiple functions including new function `readExamplesWithRuns`.

Previously it was only callable from `listRuns`.

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -697,7 +697,7 @@ export class Client implements LangSmithTracingClientInterface {
     path: string,
     body: RecordStringAny | null = null,
     requestMethod = "POST",
-    dataKey = "runs"
+    dataKey?: string
   ): AsyncIterable<T[]> {
     const bodyParams = body ? { ...body } : {};
     while (true) {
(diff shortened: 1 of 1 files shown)


js/src/client.ts
```diff
@@ -2779,6 +2784,72 @@ export class Client implements LangSmithTracingClientInterface {
...
+    datasetId: string,
+    {
+      sessionIds,
+      comparativeExperimentId,
+      filters,
+      preview = false,
+      offset = 0,
+      limit = 20,
+    }: {
+      sessionIds: string[];
```
jacoblee93, line 2811: Should we make the required parameter part of the signature?

js/src/client.ts
```diff
@@ -2779,6 +2784,72 @@ export class Client implements LangSmithTracingClientInterface {
...
+      limit,
+    };
+
+    let examplesYielded = 0;
+    for await (const exampleWithRuns of this._getCursorPaginatedList<Example>(
+      endpoint,
+      body
+    )) {
+      if (limit) {
+        if (examplesYielded >= limit) {
```
jacoblee93, line 2837: This check could happen after incrementing `examplesYielded` before another call occurs?

tendev-eoghan: Yes thanks for the fix - moved check to beneath the increment

PR #1413, opened 2025-01-14
Implement streaming multipart zstd compression.

obi1kenobi:
Uses a hand-written streaming multipart writer that can wrap a zstd compressor, so that we try to not keep `QueuedRun` values in memory any longer than strictly necessary.

A "TODO" for tomorrow's revamp of the error types is to switch from unbounded to bounded channels and either use the `QueueFull` error variant or just block until the channel frees up. @agola11 do you have a preference one way or the other? This governs what happens in the rare instance if e.g. the server is much too slow at ingesting data and traces are piling up on the client. We can either drop data, or force the client to slow down, and this is a product-level decision.

It's a good idea to use bounded channels over unbounded ones, even if the bound is huge and shouldn't ever be hit. Bounded channels are both more efficient *and* can more gracefully handle excessive load. Unbounded channels grow until OOM and then kill the program, which is a worse experience.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -153,6 +153,8 @@ version = "1.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f34d93e62b03caf570cccc334cbc6c2fceca82f39211051345108adcba3eebdc"
 dependencies = [
+ "jobserver",
+ "libc",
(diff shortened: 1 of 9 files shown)


obi1kenobi:
Supersedes #1401, #1402, #1403.

agola11 (approved):

PR #1417, opened 2025-01-15
fix: feedback creates as runs complete in evaluate

isahers1:


agola11:
what's the context on this one?

isahers1:
> what's the context on this one?

We used to create feedback after all of the evaluators for all examples have run, but this change makes it so feedback is created as soon as the evaluator is done running. This doesn't really impact users unless they are running super long experiments, but in that case it would improve their experience I think.

PR #1423, opened 2025-01-15
feat(js): Add API refs build

jacoblee93:


js/scripts/typedoc-plugin.cjs
```diff
@@ -0,0 +1,113 @@
...
+    reflection.parent &&
+    reflection.name !== "constructor"
+  ) {
+    if (kind === ReflectionKind.Property) {
+      return true;
+    }
+  }
+
+  if (REFLECTION_KINDS_TO_HIDE.find((kindToHide) => kindToHide === kind)) {
+    if (reflection.name.startsWith("_") || reflection.name.startsWith("lc_")) {
```
bracesproul, line 63: maybe `ls_` for langsmith? Not sure if there's either/or in the sdk

js/typedoc.json
```diff
@@ -0,0 +1,27 @@
...
+    "typedoc-plugin-expand-object-like-types",
+    "./scripts/typedoc-plugin.cjs"
+  ],
+  "tsconfig": "./tsconfig.json",
+  "excludePrivate": true,
+  "excludeInternal": true,
+  "excludeExternals": false,
+  "excludeNotDocumented": false,
+  "includeVersion": true,
+  "sourceLinkTemplate": "https://github.com/langchain-ai/langsmith-sdk/blob/{gitRevision}/{path}#L{line}",
```
bracesproul, line 21: i would double check this actually results in the correct links. i think since it goes off of `tsconfig` for the path it might not realize that they're nested inside `/js`

js/typedoc.json
```diff
@@ -0,0 +1,27 @@
...
+  "logLevel": "Error",
+  "name": "LangChain.js",
```
bracesproul, line 23: LangSmith?

bracesproul (approved):
couple questions, overall lgtm

js/typedoc.json
jacoblee93: Seems to work, am actually surprised

PR #1430, opened 2025-01-17
fix: send error instead of extra

samnoyes:


python/langsmith/evaluation/_runner.py
```diff
@@ -1591,9 +1591,8 @@ def _run_evaluators(
                             results=[
                                 EvaluationResult(
                                     key=key,
-                                    source_run_id=run.id,
```
samnoyes, line 1594: This was just setting the evaluator trace to be the run itself (the one the feedback is associated with). is there a way to get the evaluator trace here @hinthornw ?

hinthornw: lol

PR #1439, opened 2025-01-21
python[minor]: Release 0.3.0

baskaryan:


python/langsmith/_internal/_operations.py
```diff
@@ -323,7 +320,10 @@ def compress_multipart_parts_and_context(
             compressed_runs.uncompressed_size += len(data)
             compressed_runs.compressor_writer.write(data)
         else:
-            encoded_data = str(data).encode()
+            if isinstance(data, BufferedReader):
+                encoded_data = data.read()
+            else:
+                encoded_data = str(data).encode()
```
baskaryan, line 326: does this make sense @efriis @isahers1

efriis: Looks good

I think you can optimize this more if you create a case that streams the reader's content into the compressor_writer, but I didn't do the compressor_writer work

nfcampos (approved):

PR #1452, opened 2025-01-22
Compress create feedback

angus-langchain:
### Purpose
Compress feedback along with runs using the same compression buffer and background thread

### Changes
* Renamed compressed runs variables to compressed traces
* Compress feedback along with runs
* Added additional client ref to `num_refs` to account for compression bg thread.

### Tests
Tested locally to ensure feedback ingest works with and without compression
Added unit tests for feedback ingest with compression


hinthornw (approved):
changes lgtm if we can fix up the int tests

PR #1464, opened 2025-01-24
add wrap_anthropic

isahers1:


baskaryan (approved):

baskaryan:
think we're not handling system messages correctly, they're specified as a sep argument. we should check for it and prepend it to the traced messages

PR #1465, opened 2025-01-25
Pass multiple run ids to listFeedbacks correctly

airhorns:
Comma separated runIds give a:

```
[422]: Unprocessable Entity. Server response: {"detail":["query.run.0: Input should be a valid UUID, invalid character: expected an optional prefix of `urn:uuid:` followed by [0-9a-fA-F-], found `,` at 37"]}
```

They need to be added as separate, individual params


jacoblee93:
Thank you, sorry for the delay

jacoblee93 (approved):

Issue #1469, opened 2025-01-28
Versions `>=0.3.0` have argparse conflict with package `pytest-playwright`


rkechols:
## Problem Description

Langsmith Python SDK versions `>=0.3.0` (latest is currently `0.3.2`) seem to add command-line argument `--output` to `pytest` entrypoint (perhaps all CLI entrypoints?), which conflicts with pytest plugin `pytest-playwright` (for example) which adds the same command-line argument ([see docs](https://playwright.dev/python/docs/test-runners#cli-arguments)). When both packages are installed in the same python environment, this renders `pytest` completely unusable, giving the following error:
```
argparse.ArgumentError: argument --output: conflicting option string: --output
```

I must apologize that this may seem simply as a conflict between two packages, but I would say the core issue is here in the `langsmith` package because:
- only direct entrypoints should interact with command-line arguments
- only pytest plugins should directly modify the behavior of the `pytest` entrypoint

Additionally, even if we suppose the `langsmith` package is justified in interacting with command-line arguments:
- the `--output` command-line argument is not documented by `langsmith` (at least not anywhere that I could find; correct me if I'm wrong)
- `pytest-playwright` got to it first; they published releases using `--output` before `langsmith` did

## Potential Solutions

- Offer a separate python package such as `pytest-langsmith` which is a proper pytest plugin, and only modify pytest behavior if this plugin is installed; this allows the environment to have regular `langsmith` installed without it interfering with command-line arguments
- Use an environment variable in place of a command-line argument
- Rename the command-line argument from plain `--output` to `--langsmith-output` to reduce chance of collisions

Until an actual solution is implemented, I will be tied to using `langsmith<0.3.0`

## Full Error Stack Trace
```
$ pytest
Traceback (most recent call last):
  File "/Users/rechols/temp/.venv/bin/pytest", line 10, in <module>
    sys.exit(console_main())
             ~~~~~~~~~~~~^^
  File "/Users/rechols/temp/.venv/lib/python3.13/site-packages/_pytest/config/__init__.py", line 201, in console_main
    code = main()
  File "/Users/rechols/temp/.venv/lib/python3.13/site-packages/_pytest/config/__init__.py", line 156, in main
    config = _prepareconfig(args, plugins)
  File "/Users/rechols/temp/.venv/lib/python3.13/site-packages/_pytest/config/__init__.py", line 341, in _prepareconfig
    config = pluginmanager.hook.pytest_cmdline_parse(
[...]

PR #1472, opened 2025-01-29
fix(py): Rename CLI arg for pytest plugin from `--output` to `--langsmith-output` to avoid conflict

rkechols:
## Description

Including the pytest plugin in the base package and having it enabled by default creates conflicts with other pytest plugins (see issue https://github.com/langchain-ai/langsmith-sdk/issues/1469).

This PR:
- Renames the command-line argument for pytest plugin from `--output` to `--langsmith-output`
- Updates in-repo documentation regarding the pytest plugin

## TODO
- Update online docs describing the CLI arg: https://docs.smith.langchain.com/evaluation/how_to_guides/pytest

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -422,14 +422,7 @@ my_function("Jason is 25 years old")
 The LangSmith pytest plugin lets Python developers define their datasets and evaluations as pytest test cases.
 See [online docs](https://docs.smith.langchain.com/evaluation/how_to_guides/pytest) for more information.
 
-This plugin is installed as part of the LangSmith SDK, but is not enabled by default.  
-To enable the plugin for your project, add `"langsmith.pytest_plugin"` to the `pytest_plugins` list in your root `conftest.py` file:
-
-```python
-# root `conftest.py` file
-pytest_plugins = ["langsmith.pytest_plugin"]
-```
-
+This plugin is installed as part of the LangSmith SDK, and is enabled by default.
 See also official pytest docs: [How to install and use plugins](https://docs.pytest.org/en/stable/how-to/plugins.html)
 
 ## Additional Documentation
--- a/python/README.md
+++ b/python/README.md
@@ -345,7 +345,7 @@ def my_function(text: str):
 my_function("hello world")
 ```
 
-# Instructor
+## Instructor
 
 We provide a convenient integration with [Instructor](https://jxnl.github.io/instructor/), largely by virtue of it essentially just using the OpenAI SDK.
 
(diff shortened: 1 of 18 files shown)


rkechols:
@jacoblee93 I've just tested locally, and renaming to `--langsmith-output` does indeed resolve the conflict

python/README.md
```diff
@@ -417,6 +417,21 @@ def my_function(text: str) -> UserDetail:
...
+
+The LangSmith pytest plugin lets Python developers define their datasets and evaluations as pytest test cases.
+See [online docs](https://docs.smith.langchain.com/evaluation/how_to_guides/pytest) for more information.
+
+This plugin is installed as part of the LangSmith SDK, but is not enabled by default.  
+To enable the plugin for your project, add `"langsmith.pytest_plugin"` to the `pytest_plugins` list in your root `conftest.py` file:
+
+```python
+# root `conftest.py` file
+pytest_plugins = ["langsmith.pytest_plugin"]
```
baskaryan, line 430: delete

python/langsmith/pytest_plugin.py
```diff
@@ -301,23 +309,24 @@ def pytest_configure(config):
     config.addinivalue_line(
         "markers", "langsmith: mark test to be tracked in LangSmith"
     )
-    if config.getoption("--output") in ("langsmith", "ls"):
+    if config.getoption("--langsmith-output") in ("langsmith", "ls"):
```
rkechols, line 312: @baskaryan is this line still correct? It looked to me like `--output` was given a string value such as `langsmith` or `ls`, but when switching to `--langsmith-output` it was changed to storing a boolean.

python/README.md
rkechols: README has been updated again

python/langsmith/pytest_plugin.py
jacoblee93: Yes thank you

jacoblee93:
Integration test failure seems unrelated @baskaryan

rkechols:
Looks good to me!
Thanks for your responsiveness and cooperation!

python/langsmith/pytest_plugin.py
baskaryan: yea think given that it has "langsmith" in the arg name think its fine to make it a boolean flag now, not much need for it to take a string

jacoblee93:
Closing this in favor of #1482 in order to get this out

PR #1487, opened 2025-02-04
patch(js): utf-8 encode multipart parts

angus-langchain:


js/src/client.ts
```diff
@@ -1144,10 +1145,11 @@ export class Client implements LangSmithTracingClientInterface {
         const fields = { inputs, outputs, events };
         // encode the main run payload
         const stringifiedPayload = stringifyForTracing(payload);
+        const valueBytes = encoder.encode(stringifiedPayload).length;
```
jacoblee93, line 1148: Is this necessary? Shouldn't `stringifyForTracing` already return proper encoding?

angus-langchain: it is necessary. The stringify function itself doesn't actually encode anything. It converts the js object into a JSON string.
JSON.stringify() produces a JavaScript string, which is always UTF-16.

jacoblee93: Right, but *could/should* our `stringifyForTracing` wrapper do the conversion?

angus-langchain: If we don’t need to parse a length=... parameter (just application/json), then JSON.stringify(...) should be sufficient. The browser will treat it as a UTF-8 text blob, and the server sees it as JSON with no mismatch in length.

angus-langchain: But I can add an explicit encoding

jacoblee93 (approved):
Thanks!

hinthornw (approved):

PR #1488, opened 2025-02-04
fix: use go backend instead of python backend for example functions

isahers1:


js/src/client.ts
```diff
@@ -10,7 +10,7 @@ import {
   Example,
   ExampleCreate,
   ExampleUpdate,
-  ExampleUpdateWithId,
```
jacoblee93, line 13: Just checking - was this exported publicly?

Also good to preserve naming scheme:

`ExampleUpdateWithoutId`

js/src/client.ts
```diff
@@ -1060,24 +1064,40 @@ export class Client implements LangSmithTracingClientInterface {
...
     const allAttachments: Record<string, Attachments> = {};
-    let preparedCreateParams = [];
-    for (const create of runCreates ?? []) {
-      const preparedCreate = this.prepareRunCreateOrUpdateInputs(create);
-      if (
-        preparedCreate.id !== undefined &&
-        preparedCreate.attachments !== undefined
-      ) {
-        allAttachments[preparedCreate.id] = preparedCreate.attachments;
+    let preparedCreateParams =
```
jacoblee93, line 1067: Is this equivalent to what was there before?

If so, why change?

js/src/client.ts
```diff
@@ -2676,113 +2697,164 @@ export class Client implements LangSmithTracingClientInterface {
     return result["examples"] as ExampleSearch[];
   }
 
+  public async createExample(update: ExampleCreate): Promise<Example>;
```
jacoblee93, line 2700: Why overload?

js/src/client.ts
```diff
@@ -2676,113 +2697,164 @@ export class Client implements LangSmithTracingClientInterface {
...
+      response.example_ids?.[0] ?? uuid.v4()
     );
-
-    await raiseForStatus(response, "create example");
-    const result = await response.json();
-    return result as Example;
+    return example;
   }
 
   public async createExamples(props: {
```
jacoblee93, line 2761: This one should be an overload:

```
public async createExamples(props: Example[]);

public async createExamples(props: ...<current signature>)
```

isahers1:
deprecate the old signatures

js/src/client.ts
isahers1: yes it was exported before and it is still being exported for backwards compatibility

js/src/client.ts
```diff
@@ -2404,7 +2410,7 @@ export class Client implements LangSmithTracingClientInterface {
     } else if (datasetName !== undefined) {
       datasetId = (await this.readDataset({ datasetName })).id;
     } else {
-      throw new Error("Must provide datasetName or datasetId");
+      throw new Error("Must provide dataset name or ID");
```
jacoblee93, line 2413: Old error message is better

js/src/client.ts
```diff
@@ -2676,113 +2682,182 @@ export class Client implements LangSmithTracingClientInterface {
...
+  ): Promise<Example>;
+
+  public async createExample(
+    inputsOrUpdate: KVMap | ExampleCreate,
+    outputs?: KVMap,
+    options?: CreateExampleOptions
   ): Promise<Example> {
-    let datasetId_ = datasetId;
-    if (datasetId_ === undefined && datasetName === undefined) {
+    if ("dataset_id" in inputsOrUpdate || "dataset_name" in inputsOrUpdate) {
```
jacoblee93, line 2701: Simplify all this with a user-defined type guard:

https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

You will be able to avoid the cast below

jacoblee93 (commented):
Interfaces look good, implementation could be cleaned up with type guards

PR #1489, opened 2025-02-04
feat(py): Increase concurrency for pytest evals, fix output rendering nits

jacoblee93:


python/langsmith/testing/_internal.py
```diff
@@ -684,23 +653,24 @@ def end_run(
...
+        pytest_nodeid,
     ) -> None:
-        # TODO: remove this hack so that run durations are correct
         # Ensure example is fully updated
-        self.wait_example_updates(example_id)
-        # Ensure that run end time is after example modified at.
-        example_modified_at = self.get_version(example_id=example_id) or end_time
-        end_time = max(example_modified_at, end_time)
-        run_tree.end(outputs=outputs, end_time=end_time)
+        self.sync_example(example_id, inputs=run_tree.inputs, outputs=reference_outputs)
```
baskaryan, line 671: this is big brain

python/langsmith/testing/_internal.py
```diff
@@ -684,23 +653,24 @@ def end_run(
...
-        self, run_tree, example_id, outputs, end_time, pytest_plugin, pytest_nodeid
+        self,
+        run_tree,
+        example_id,
+        outputs,
+        reference_outputs,
+        pytest_plugin,
+        pytest_nodeid,
     ) -> None:
-        # TODO: remove this hack so that run durations are correct
```
baskaryan, line 695: note we are still delaying run durations so think we should leave the todo

python/langsmith/testing/_internal.py
```diff
@@ -684,23 +653,24 @@ def end_run(
...
         # Ensure example is fully updated
-        self.wait_example_updates(example_id)
-        # Ensure that run end time is after example modified at.
-        example_modified_at = self.get_version(example_id=example_id) or end_time
-        end_time = max(example_modified_at, end_time)
```
baskaryan, line 700: i think we still want to keep this part. just because sync_example returns doesn't mean the example has actually been updated in the backend, stuff is queued and can take a bit

python/langsmith/testing/_internal.py
```diff
@@ -837,12 +820,6 @@ def _create_test_case(
     )
     example_id, example_name = _get_example_id(func, inputs, test_suite.id)
     example_id = langtest_extra["id"] or example_id
-    test_suite.sync_example(
-        example_id,
-        inputs=inputs,
-        outputs=outputs,
-        metadata={"signature": _get_test_repr(func, signature), "name": example_name},
-    )
```
baskaryan, line 845: are we sure the example doesn't need to exist before its referenced in a run? also i think if we dont do this here we lose pytest fixture inputs/outputs

python/langsmith/testing/_internal.py
jacoblee93: Ah ok interesting

python/langsmith/testing/_internal.py
jacoblee93: I think it will work but will test with a fresh dataset too

python/langsmith/testing/_internal.py
baskaryan: when i didn't have this i was pretty frequently running into issue that example modified time was after run end time. maybe something else that you've changed will alleviate that, but worth really testing

baskaryan: note to test it make sure you're running on a new dataset and/or that your tests are updating the examples in some way. otherwise you won't be able to tell

jacoblee93: Seems to be working ok for me

python/langsmith/testing/_internal.py
jacoblee93: Added dedicated logging for fixture inputs and outputs, and @angus-langchain confirmed this will work

python/langsmith/pytest_plugin.py
```diff
@@ -41,8 +41,8 @@ def _handle_output_args(args):
     """Handle output arguments."""
     if any(opt in args for opt in ["--langsmith-output"]):
         # Only add --quiet if it's not already there
-        if not any(a in args for a in ["-q", "--quiet"]):
```
baskaryan, line 44: i think we still want to check for presence of -q or --quiet?

python/langsmith/testing/_internal.py
```diff
@@ -591,23 +576,7 @@ def sync_example(
...
-            self._example_futures[example_id].append(future)
-
-    def _sync_example(
-        self,
-        example_id: uuid.UUID,
-        inputs: Optional[dict],
-        outputs: Optional[dict],
-        metadata: Optional[dict],
-    ) -> None:
+        metadata = metadata.copy() if metadata else metadata
```
baskaryan, line 579: whats the copy for again

jacoblee93: Dunno, was there already

python/langsmith/pytest_plugin.py
jacoblee93: Why?

jacoblee93: Seems to work fine: https://www.loom.com/share/764ecd32c495459db8d7095a3f8b4a80

baskaryan (approved):

jacoblee93:
Thank you!

PR #1525, opened 2025-02-19
allow keyword args for evaluators

isahers1:


python/langsmith/evaluation/evaluator.py
```diff
@@ -744,7 +752,7 @@ async def awrapper(
                     "outputs": [run.outputs or {} for run in runs],
                     "reference_outputs": example.outputs or {} if example else {},
                 }
-                args = (arg_map[arg] for arg in positional_args)
+                args = (arg_map[arg] for arg in all_args)
                 return await func(*args)
```
baskaryan, line 756: dont the kwargs need to be passed in as kwargs

python/langsmith/evaluation/evaluator.py
```diff
@@ -633,28 +633,72 @@ def _normalize_evaluator_func(
...
-        pname in supported_args for pname in positional_args
-    ) or positional_args == ["run", "example"]:
-        # For backwards compatibility we assume custom arg names are Run and Example
-        # types, respectively.
-        return func
+        pname in supported_args or pname in args_with_defaults for pname in all_args
+    ) or all_args == [
+        "run",
+        "example",
+    ]:
```
baskaryan, line 662: i dont think we need to update this case, its for old eval functions that take two positional args with random names. this change shouldn't affect this code path?

isahers1: The case I was thinking about was `def evaluator(foo, *, bar)`. Which I believe failed in the current case.

baskaryan: but that was never supported, so why do we need to support now? in general we want ppl to name things correctly, we just have to support `def evaluator(foo, bar)` for backwards compat

baskaryan (approved):

PR #1536, opened 2025-02-24
feat(js): Allow evaluators to return arrays of results

jacoblee93:


jacoblee93:
Python changes were unnecessary, PTAL @hinthornw

hinthornw (approved):

PR #1538, opened 2025-02-26
Fix evaluator errors

isahers1:


python/langsmith/evaluation/_arunner.py
```diff
@@ -1005,7 +1007,7 @@ async def _run_single_evaluator(evaluator):
                             results=[
                                 EvaluationResult(
                                     key=key,
-                                    source_run_id=run.id,
```
baskaryan, line 1008: not sure i follow this bit, does source run not refer to the run being evaluated?

isahers1: yeah this is confusing. this is the run for the evaluator. basically the issue was that by setting it to the parent run, when you were in langsmith in the experiment view and clicked on a failed evaluator run (or what should have been the evaluator run) it just opened up the target run. This fixes that.

isahers1:
Tests failing are 429 errors from OpenAI, not actual errors

python/langsmith/evaluation/evaluator.py
```diff
@@ -307,7 +307,10 @@ def is_async(self) -> bool:
         return hasattr(self, "afunc")
 
     def evaluate_run(
-        self, run: Run, example: Optional[Example] = None
+        self,
+        run: Run,
+        example: Optional[Example] = None,
+        source_run_id: Optional[uuid.UUID] = None,
```
baskaryan, line 313: can we give this a diff name, like evaluator_run_id?

baskaryan (approved):

Issue #642, opened 2024-04-30
`BaseRun.start_time` and `BaseRun.end_time` should be strings


davidfant:
Today I used [patch-package](https://github.com/ds300/patch-package) to patch `langsmith@0.1.20` for the project I'm working on. start/end time from the API is an ISO string, not a number

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/langsmith/dist/schemas.d.ts b/node_modules/langsmith/dist/schemas.d.ts
index 189a1cd..80d53ae 100644
--- a/node_modules/langsmith/dist/schemas.d.ts
+++ b/node_modules/langsmith/dist/schemas.d.ts
@@ -39,11 +39,11 @@ export interface BaseRun {
     /** A human-readable name for the run. */
     name: string;
     /** The epoch time at which the run started, if available. */
-    start_time?: number;
+    start_time?: string;
     /** Specifies the type of run (tool, chain, llm, etc.). */
     run_type: string;
     /** The epoch time at which the run ended, if applicable. */
-    end_time?: number;
+    end_time?: string;
     /** Any additional metadata or settings for the run. */
     extra?: KVMap;
     /** Error message, captured if the run faces any issues. */
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

PR #1540, opened 2025-02-27
change start/end time from number to string

isahers1:
Resolves https://github.com/langchain-ai/langsmith-sdk/issues/642

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -238,8 +238,8 @@ interface CreateRunParams {
   inputs: KVMap;
   run_type: string;
   id?: string;
-  start_time?: number;
-  end_time?: number;
+  start_time?: string;
+  end_time?: string;
   extra?: KVMap;
   error?: string;
   serialized?: object;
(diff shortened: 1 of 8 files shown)


js/src/client.ts
```diff
@@ -922,7 +922,7 @@ export class Client implements LangSmithTracingClientInterface {
     const runCreate: RunCreate = this.prepareRunCreateOrUpdateInputs({
       session_name,
       ...run,
-      start_time: run.start_time ?? Date.now(),
+      start_time: run.start_time ?? String(Date.now()),
```
dqbd, line 925: new Date().toISOString()

js/src/run_trees.ts
```diff
@@ -22,7 +22,7 @@ function stripNonAlphanumeric(input: string) {
 }
 
 export function convertToDottedOrderFormat(
-  epoch: number,
+  epoch: string,
```
dqbd, line 25: _nit_: why the epoch here is not a number?

PR #1548, opened 2025-03-05
fix(python,js): Round feedback scores with more than 4 decimal places

jacoblee93:


python/langsmith/client.py
```diff
@@ -256,6 +256,21 @@ def _validate_api_key_if_hosted(api_url: str, api_key: Optional[str]) -> None:
...
+def _format_feedback_score(score: Union[float, int, bool, None]):
+    """Format a feedback score by truncating numerical values to 4 decimal places.
+    
+    Args:
+        score: The score to format, can be a number or any other type
+        
+    Returns:
+        The formatted score
+    """
+    if isinstance(score, (int, float)):
```
baskaryan, line 268: not needed for ints technically

baskaryan (approved):

python/langsmith/client.py
jacoblee93: Ah yeah true

PR #1552, opened 2025-03-06
fix, don't append v1 if it is already in the api url

isahers1:


python/langsmith/client.py
```diff
@@ -4159,7 +4159,11 @@ def _update_examples_multipart(
         try:
             response = self.request_with_retries(
                 "PATCH",
-                f"/v1/platform/datasets/{dataset_id}/examples",
+                (
+                    f"/v1/platform/datasets/{dataset_id}/examples"
+                    if "v1" not in self.api_url
```
langchain-infra, line 4164: slight nit, can we do an ends with /v1? e.g they might have langsmithv1 as their dns name or something and this would still trigger.

isahers1: yup, will fix

angus-langchain (approved):

PR #1567, opened 2025-03-11
Wrap Openai Responses Pt 1

hinthornw:


python/langsmith/wrappers/_openai.py
```diff
@@ -382,5 +391,41 @@ def wrap_openai(
...
+            client.responses.parse = _get_parse_wrapper(  # type: ignore[method-assign]
+                client.responses.parse,
+                "openai.responses.parse",
+                tracing_extra=tracing_extra,
+                invocation_params_fn=functools.partial(
+                    _infer_invocation_params, "chat"
+                ),
+            )
+
+        # TODO: Handle stream() correctly. It's a context manager.
```
baskaryan, line 429: hopefully can copy a lot of wrap_anthropic

hinthornw: Ya both would use the same primitive i think

baskaryan (approved):

baskaryan (commented):
probs should convert input to a message and call trace ChatOpenAI? https://dev.smith.langchain.com/public/c098a92d-bada-43ae-b7dd-5731a136a538/r

PR #1568, opened 2025-03-11
Create OTel Exporter for LangChain tracing

angus-langchain:
### Purpose
Folks want to natively trace LangChain and LangGraph through our sdk. The implementation follows [OpenTelemetry GenAI Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/)

### Usage
The OTEL integration can be enabled by:

Installing the OTEL dependencies: `pip install langsmith[otel]`
Setting environment variable: `LANGSMITH_OTEL_ENABLED=true`

By default, the exporter will send data to the LangSmith API OTEL endpoint, but this can be customized by setting standard OTEL environment variables:
```
OTEL_EXPORTER_OTLP_ENDPOINT: Override the endpoint URL
OTEL_EXPORTER_OTLP_HEADERS: Add custom headers (API keys are added automatically)
OTEL_SERVICE_NAME: Set a custom service name (defaults to "langsmith")
```

Depends on https://github.com/langchain-ai/langchainplus/pull/8582 (including integration test passing)


python/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "langsmith"
-version = "0.3.13"
+version = "0.3.14rc1"
```
angus-langchain, line 3: will rebase to current version before merging

python/tests/integration_tests/test_client.py
```diff
@@ -3012,3 +3012,47 @@ def _get_run(run_id: ID_TYPE) -> bool:
...
+    retrieved_run = client.read_run(run_id)
+    assert retrieved_run.id == run_id
+    assert retrieved_run.start_time == start_time
+    assert retrieved_run.end_time == end_time
+    assert retrieved_run.name == "OTEL Export Test"
+    assert retrieved_run.run_type == "llm"
+    assert retrieved_run.inputs == {"prompt": "Hello, OTEL!"}
+    assert retrieved_run.outputs == {"answer": "Hello, User!"}
+    assert retrieved_run.extra == {"metadata": {"foo": "bar"}}
+    assert retrieved_run.tags == ["otel", "test"]
```
agola11, line 3058: might be good to handle more cases here

python/langsmith/_internal/_background_thread.py
```diff
@@ -168,6 +168,30 @@ def _tracing_thread_handle_batch(
...
+    batch: List[TracingQueueItem],
+) -> None:
+    """Handle a batch of tracing queue items by exporting them to OTEL."""
+    try:
+        ops = combine_serialized_queue_operations([item.item for item in batch])
+
+        run_ops = [op for op in ops if isinstance(op, SerializedRunOperation)]
+
+        if run_ops and client.otel_exporter is not None:
+            client.otel_exporter.export_batch(run_ops)
```
agola11, line 183: maybe log something here because `client.otel_exporter` should never be `None` here?

python/langsmith/_internal/_background_thread.py
```diff
@@ -278,15 +306,23 @@ def keep_thread_active() -> bool:
             sub_threads.append(new_thread)
             new_thread.start()
         if next_batch := _tracing_thread_drain_queue(tracing_queue, limit=size_limit):
-            _tracing_thread_handle_batch(
-                client, tracing_queue, next_batch, use_multipart
-            )
+            if client.otel_exporter is not None:
+                _otel_tracing_thread_handle_batch(client, tracing_queue, next_batch)
+            else:
+                _tracing_thread_handle_batch(
```
akira, line 312: is it possible to have both OTEL and regular LangSmith tracing or would only go to one or the other?

python/langsmith/_internal/otel/_otel_exporter.py
```diff
@@ -0,0 +1,651 @@
...
+            span.set_attribute(LANGSMITH_SESSION_ID, str(run_info.get("session_id")))
+        if run_info.get("session_name"):
+            span.set_attribute(
+                LANGSMITH_SESSION_NAME, str(run_info.get("session_name"))
+            )
+
+        # Set GenAI attributes according to OTEL semantic conventions
+        # Set gen_ai.operation.name
+        operation_name = run_info.get("run_type", "chain")
+        span.set_attribute(GEN_AI_OPERATION_NAME, operation_name)
```
akira, line 308: what are the mappings for this, would it come through as "chat", etc?

python/langsmith/_internal/otel/_otel_exporter.py
```diff
@@ -0,0 +1,651 @@
...
+GEN_AI_REQUEST_TOP_P = "gen_ai.request.top_p"
+GEN_AI_REQUEST_FREQUENCY_PENALTY = "gen_ai.request.frequency_penalty"
+GEN_AI_REQUEST_PRESENCE_PENALTY = "gen_ai.request.presence_penalty"
+GEN_AI_RESPONSE_FINISH_REASONS = "gen_ai.response.finish_reasons"
+GENAI_PROMPT = "gen_ai.prompt"
+GENAI_COMPLETION = "gen_ai.completion"
+
+# Additional GenAI attributes
+GEN_AI_REQUEST_STREAMING = "gen_ai.request.streaming"
+GEN_AI_REQUEST_HEADERS = "gen_ai.request.headers"
```
akira, line 49: are these general semconv or should they be langsmith prefixed?

python/langsmith/_internal/_background_thread.py
angus-langchain: one or the other, but ls accepts the otel messages

python/langsmith/_internal/otel/_otel_exporter.py
angus-langchain: mapping in _get_operation_name

python/langsmith/_internal/otel/_otel_exporter.py
angus-langchain: langsmith prefixed

akira (approved):

python/langsmith/_internal/_background_thread.py
codefromthecrypt: ack on "one or the other", but the latter point "accepts the otel messages" is probably not that useful unless I missed something. 

If you run it with `opentelemetry-instrument` to bootstrap the otel exporters, both langsmith and other instrumentation will export to the same collector. That's true.

However, they will be in different traces as the tracer implementation here is an export only-concern. Meaning it cannot propagate trace context to normal otel instrumentation. For that reason, they'd be in different traces, and likely less useful as a result.

angus-langchain: @codefromthecrypt I was referring to LangSmith if a user wanted to fan out on receiver to multiple OTel backends. ack that this implementation is export only, but it should still be possible to inject headers into LangSmith OTel tracing. Would be happy to work with you as a design partner to discuss future features for distributed tracing with you though!

Issue #1566, opened 2025-03-11
Add OpenAI Agents tracing support


baskaryan:
### Feature request

https://openai.github.io/openai-agents-python/tracing/#custom-tracing-processors

### Motivation

Similar to wrap_openai, add automatic agent tracing

PR #1572, opened 2025-03-12
Support Tracing OpenAI Agents SDK

angus-langchain:
https://openai.github.io/openai-agents-python/tracing/#custom-tracing-processors

Resolves https://github.com/langchain-ai/langsmith-sdk/issues/1566


python/langsmith/wrappers/_openai_agents.py
```diff
@@ -0,0 +1,114 @@
+import logging
+import uuid
+from typing import Dict, Optional
+
+try:
+    from agents import tracing  # type: ignore[import]
```
hinthornw, line 6: could maybe cache this attempt?

angus-langchain: what benefit would this have

python/langsmith/wrappers/_openai_agents.py
```diff
@@ -0,0 +1,116 @@
...
+                )
+                self.client.create_run(**run_data)
+            except Exception as e:
+                logger.exception(f"Error creating trace run: {e}")
+
+        def on_trace_end(self, trace: tracing.Trace) -> None:
+            run_id = self._runs.pop(trace.trace_id, None)
+            if run_id:
+                try:
+                    self.client.update_run(
```
baskaryan, line 67: ooc why dont we trace any inputs/outputs for the root run?

python/langsmith/wrappers/_openai_agents.py
```diff
@@ -0,0 +1,116 @@
...
+                    run_id=run_id,
+                    error=str(span.error) if span.error else None,
+                    inputs=extracted.get("inputs", {}),
+                    outputs=extracted.get("outputs", {}),
+                    extra={"metadata": extracted.get("metadata", {})},
+                )
+                self.client.update_run(**run_data)
+
+        def shutdown(self) -> None:
+            if self.client is not None:
```
baskaryan, line 107: can this be none? whats pt of tracer if it is?

python/langsmith/wrappers/_agent_utils.py
```diff
@@ -0,0 +1,183 @@
...
+    ) -> Dict[str, Any]:
+        data = {
+            "inputs": parse_io(span_data.input, "input"),
+            "outputs": parse_io(span_data.output, "output"),
+            "metadata": {
+                "model": span_data.model,
+                "model_config": span_data.model_config,
+            },
+        }
+        if span_data.usage:
```
baskaryan, line 101: ooc do they provide input/output_token details

python/langsmith/wrappers/_agent_utils.py
```diff
@@ -0,0 +1,183 @@
...
+            data["usage_metadata"] = {
+                "total_tokens": span_data.usage.get("total_tokens"),
+                "input_tokens": span_data.usage.get("prompt_tokens"),
+                "output_tokens": span_data.usage.get("completion_tokens"),
+            }
+        return data
+
+    def _extract_response_span_data(
```
baskaryan, line 109: i think with the new rendering updates we can just dump input/output as is and FE should be able to display nicely. thats what im planning to do in wrap_openai

python/langsmith/wrappers/_agent_utils.py
baskaryan: also whats diff between generations and responses? couldn't find explanation in their docs

baskaryan:
think we can simplify responses tracing, otherwise lgtm

python/langsmith/wrappers/_openai_agents.py
angus-langchain: The OpenAI SDK doesn't provide inputs/outputs in the trace start or end methods

python/langsmith/wrappers/_agent_utils.py
angus-langchain: The backend expects a dict object, we'd need to make changes in the backend as well

python/langsmith/wrappers/_agent_utils.py
angus-langchain: They do yes, via `prompt_tokens`, `total_tokens` and `completion_tokens`

baskaryan: as in do that have a input_token_details and output_token_details (contains specifics about eg reasoning, audio tokens)? probably more relevant for Response api, which i know has this info for model calls

baskaryan: seems they dont https://openai.github.io/openai-agents-python/ref/usage/#agents.usage.Usage.input_tokens

python/langsmith/wrappers/_openai_agents.py
```diff
@@ -0,0 +1,110 @@
...
+            )
+
+
+from langsmith import client as ls_client
+
+logger = logging.getLogger(__name__)
+
+if HAVE_AGENTS:
+
+    class LangsmithTracingProcessor(tracing.TracingProcessor):  # type: ignore[no-redef]
```
baskaryan, line 32: should we call this OpenAIAgentsTracingProcessor? end user knows its langsmith if they're importing from langsmith, and LangSmithTracingProcessor sounds like something much more generic than an openai agents tracer

baskaryan: updated

baskaryan (approved):

PR #1573, opened 2025-03-13
fix(js): Avoid sampling non-root child runs that were previously filtered

jacoblee93:


js/src/client.ts
```diff
@@ -778,16 +786,39 @@ export class Client implements LangSmithTracingClientInterface {
...
+
+        // If we've already made a decision about this trace, follow it
+        if (this.filteredPostUuids.has(traceId)) {
+          continue;
+        }
+
+        // For new traces, apply sampling
+        if (run.id === traceId) {
+          if (this._shouldSample()) {
+            console.log("Sampling run", run.inputs);
```
angus-langchain, line 814: debugging print?

angus-langchain (approved):

js/src/client.ts
jacoblee93: Yeah my bad

js/src/client.ts
```diff
@@ -1012,8 +1042,8 @@ export class Client implements LangSmithTracingClientInterface {
       preparedUpdateParams = standaloneUpdates;
     }
     const rawBatch = {
-      post: this._filterForSampling(preparedCreateParams),
-      patch: this._filterForSampling(preparedUpdateParams, true),
+      post: preparedCreateParams,
```
angus-langchain, line 1033: where is sampling being applied now?

jacoblee93: All comes through `create/updateRun`

angus-langchain: thanks!

PR #1598, opened 2025-03-20
fix(js): specify "Content-Type" for getRunStats to avoid parsing errors

szedelenyi:
I always get the following `422` error when I try to use the `getRunStats` method:
```
{
  detail: [
    'body: Input should be a valid dictionary or object to extract fields from'
  ]
}
```

I figured out, it can be solved by specifying the "Content-Type"


jacoblee93:
Sorry for the delay - thanks so much!

PR #1609, opened 2025-03-27
fix(py): serialize langchain objects in testing.log_outputs, fix rich display table [LS-3237]

baskaryan:


python/langsmith/testing/_internal.py
```diff
@@ -1302,3 +1303,25 @@ def _stringify(x: Any) -> str:
         return dumps_json(x).decode("utf-8", errors="surrogateescape")
     except Exception:
         return str(x)
+
+
+def _dumpd(x: Any) -> Any:
+    """Serialize LangChain Serializable objects."""
```
jacoblee93, line 1309: I think we should try to at least replace other non-serializable objects rather than just failing

baskaryan: like pydantic?

jacoblee93: Yeah or other things

jacoblee93: It also doesn't handle arrays well, error message is confusing:

`dictionary update sequence element #0 has length 7; 2 is required`

But would be down to just ship this first if it'll fix messages in general

jacoblee93 (approved):
Thanks!

python/langsmith/pytest_plugin.py
```diff
@@ -73,7 +74,7 @@ def pytest_runtest_call(item):
         original_func = item.obj
         item.obj = ls_test(**kwargs)(original_func)
         request_obj = getattr(item, "_request", None)
-        if request_obj is not None and "request" not in item.funcargs:
+        if request_obj is not None and "request" in item.funcargs:
```
jacoblee93, line 77: @baskaryan someone flipped a boolean 😬

python/langsmith/pytest_plugin.py
```diff
@@ -75,6 +76,7 @@ def pytest_runtest_call(item):
         request_obj = getattr(item, "_request", None)
         if request_obj is not None and "request" not in item.funcargs:
             item.funcargs["request"] = request_obj
+        if request_obj is not None and "request" not in item._fixtureinfo.argnames:
```
jacoblee93, line 79: @baskaryan checking this separately seems to have fixed the problem

I am not sure why we are modifying internals like this

python/pyproject.toml
```diff
@@ -36,7 +36,7 @@ requests-toolbelt = "^1.0.0"
 langsmith-pyo3 = { version = "^0.1.0rc2", optional = true }
 zstandard = "^0.23.0"
 rich = {version = "^13.9.4", optional = true}
-pytest = {version = ">=7.0.0", optional = true}
+pytest = {version = ">=8.0.0", optional = true}
```
baskaryan, line 39: is this necessary? does it not work with v7 after the fix?

jacoblee93: This is to fix the VCR issue from Friday

jacoblee93: Never mind, I'm dumb

PR #1617, opened 2025-03-30
fix(py): Relax integration tests

jacoblee93:
Some extra field got added, currently broken here:

https://github.com/langchain-ai/langsmith-sdk/actions/runs/14135241769/job/39620024048?pr=1613


hinthornw:
Diff fix in other PR. tl;dr we need to be more exclusive in parsing s3_urls -> attachments in the read path now that we offload more data to s3

PR #1623, opened 2025-04-01
fix(py): Always upload multipart inputs and outputs parts even if falsey

jacoblee93:


python/langsmith/testing/_internal.py
```diff
@@ -566,7 +566,7 @@ def sync_example(
         self,
         example_id: uuid.UUID,
         *,
-        inputs: Optional[dict] = None,
+        inputs: Optional[dict] = {},
```
jacoblee93, line 569: Opting to fallback to empty object in the test runner because `create_examples` currently throws if `inputs` is `None`

baskaryan: gotta be careful with mutable defaults in py, should leave as None and overwrite in the function https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

jacoblee93: O no

python/langsmith/client.py
```diff
@@ -4085,34 +4085,32 @@ def _prepare_multipart_data(
                 )
             )
 
-            if example.inputs:
-                inputsb = _dumps_json(example.inputs)
+            inputsb = _dumps_json(example.inputs)
```
baskaryan, line 4088: this we should probably do
```suggestion
            inputsb = _dumps_json(example.inputs or {})
```
same with outputs. to upload empty dict not null

jacoblee93: I think maybe ok to have `null` in example rows but don't feel strongly

jacoblee93:
Done PTAL

python/tests/integration_tests/test_client.py
```diff
@@ -177,6 +177,15 @@ def test_datasets(langchain_client: Client) -> None:
         langchain_client.list_examples(dataset_id=new_dataset.id)  # type: ignore
     )
     assert len(examples2) == 2
+    langchain_client.create_example(
+        inputs={},
+        outputs=None,
+        dataset_id=new_dataset.id,
+    )
+    examples3 = list(
```
baskaryan, line 185: can we check the value of the last created example

jacoblee93: Done

baskaryan (approved):

PR #1633, opened 2025-04-03
add update_dataset method

isahers1:


isahers1:
test errors are 429 quota errors from openai

python/langsmith/client.py
```diff
@@ -3599,6 +3594,98 @@ def delete_dataset(
...
+    def update_dataset(
+        self,
+        *,
+        dataset_id: Optional[ID_TYPE] = None,
+        dataset_name: Optional[str] = None,
+        name: Optional[str] = None,
+        description: Optional[str] = None,
+        inputs_schema: Optional[Dict[str, Any]] = None,
+        outputs_schema: Optional[Dict[str, Any]] = None,
+        patch_examples: Optional[Dict[ID_TYPE, ls_schemas.ExampleUpdate]] = None,
```
baskaryan, line 3629: feels like this doesn't belong here, can just use update_examples? or do we need this bc when you change schemas you have to simultaneously update the examples

PR #1635, opened 2025-04-04
patch[python]: Add debug logging for batch requests to LangSmith

angus-langchain:
### Purpose

Enabling this logging is useful for pinpointing where traces are being lost. Right now it's difficult to tell if they are making it from the sdk into LangSmith or not

### DIFF
--- a/python/langsmith/_internal/_background_thread.py
+++ b/python/langsmith/_internal/_background_thread.py
@@ -328,6 +328,7 @@ def tracing_control_thread_func(client_ref: weakref.ref[Client]) -> None:
             _tracing_thread_handle_batch(
                 client, tracing_queue, next_batch, use_multipart
             )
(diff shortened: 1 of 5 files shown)


python/langsmith/_internal/_background_thread.py
```diff
@@ -129,6 +129,7 @@ def _tracing_thread_drain_compressed_buffer(
         client.compressed_traces.compressor_writer.close()
 
         filled_buffer = client.compressed_traces.buffer
+        filled_buffer.context = client.compressed_traces._context
```
hinthornw, line 132: this is threadsafe?

hinthornw (approved):

python/langsmith/_internal/_background_thread.py
angus-langchain: Yes, we grab a lock prior to this call

Issue #1640, opened 2025-04-07
Issue: Invalid type NoneType when in otel mode


codefromthecrypt:
### Issue you'd like to raise.

I'm trying the langsmith otel integration and get the following when I run:

```
Invalid type NoneType for attribute 'langsmith.metadata.ls_temperature' value.
```

### Suggestion:

I would suggest to not add an attribute when the value is none.

angus-langchain:
Hi @codefromthecrypt, thanks for reporting. I'll take a look this morning and see if I can repro/fix.

PR #1641, opened 2025-04-07
patch[python]: Don't set metadata values if the value is None

angus-langchain:
This change fixes an issue in the LangSmith OTEL integration where metadata attributes with a value of None were being set on spans, causing errors like:

Invalid type NoneType for attribute 'langsmith.metadata.ls_temperature' value.

Changes:

Added a check to ensure that metadata values are only set if they are not None.

Bumped the package version from 0.3.24 to 0.3.25.

Resolves: https://github.com/langchain-ai/langsmith-sdk/issues/1640


hinthornw (approved):
It's ok if it's falsey, right?

Issue #1630, opened 2025-04-03
Issue: Client silently stops posting traces to the API (`auto_batch_tracing=True`, `DISABLE_RUN_COMPRESSION=False`)


menezesandre:
### Issue you'd like to raise.

I have encountered this issue a few times while running an application with LangSmith tracing, with auto batch and compression enabled. At some point it stops posting traces to the LangSmith API, while the application keeps running ok and there is no log stating that something went wrong. I have to restart the app to get it posting again.

I couldn't figure out exactly why this happens. I'm assuming that the thread running `tracing_control_thread_func_compress_parallel` dies for some reason, and so there is nothing reading from the `compressed_traces`.

### Suggestion:

- Improve error handling and logging for `tracing_control_thread_func_compress_parallel` (while `_tracing_thread_drain_queue` and `_tracing_thread_handle_batch` are wrapped in a try-except, the corresponding actions for compression enabled are not)
- Add thread management to the main thread. I.e. keep a reference to the tracing control threads in `Client`, and check on them whenever putting something in `compressed_traces`/`tracing_queue`, potentially re-launching the threads (similar to what is done in [concurrent.futures.ThreadPoolExecutor.submit](https://github.com/python/cpython/blob/main/Lib/concurrent/futures/thread.py#L215))

menezesandre:
Hello @angus-langchain, thanks for the quick reply!

> What version of the SDK are you on?

Last occurrence was with version 0.3.19 of the python SDK. Now I updated to 0.3.24, will let you know if it happens again. But looking at https://github.com/langchain-ai/langsmith-sdk/compare/v0.3.19...v0.3.24, I can't see any change that would affect this issue.

> Does tracing work without compression?

To clarify, it's not that tracing with compression _doesn't work_ - it does most of the time. It just sometimes stops working and I need to restart the app (happened 3 times in a 15 days period). So all I can say is that I have never experienced this issue with compression disabled.

angus-langchain:
@menezesandre, thanks for the info. I'll look into this and report back. In the meantime, you can use `client.flush()` or `wait_for_all_tracers()` before exiting your process. This will ensure all traces are sent
* https://docs.smith.langchain.com/observability/how_to_guides/trace_with_langchain#ensure-all-traces-are-submitted-before-exiting
* https://docs.smith.langchain.com/observability/how_to_guides/annotate_code#using-the-langsmith-sdk

angus-langchain:
[...]

PR #1649, opened 2025-04-08
chore[python]: Add thread management to main thread

angus-langchain:
If our background threads get terminated during execution, we should ensure the threads are running then when more runs are added to the queue.

Issue reported in https://github.com/langchain-ai/langsmith-sdk/issues/1630

### DIFF
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -77,6 +77,9 @@ from langsmith._internal._background_thread import (
 from langsmith._internal._background_thread import (
     tracing_control_thread_func as _tracing_control_thread_func,
 )
+from langsmith._internal._background_thread import (
+    tracing_control_thread_func_compress_parallel as _tracing_control_thread_func_compress_parallel,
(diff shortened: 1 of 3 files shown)


python/langsmith/client.py
```diff
@@ -1393,6 +1423,11 @@ def create_run(
                     "Adding compressed multipart to queue with context: %s",
                     multipart_form.context,
                 )
+                self._ensure_thread_running(
+                    "_compress_control_thread",
+                    _tracing_control_thread_func_compress_parallel,
+                    weakref.ref(self),
```
menezesandre, line 1429: The client reference is already passed in https://github.com/langchain-ai/langsmith-sdk/blob/d4ddcbc9bb082ca1e02519ea9ed7c7c58bdc7dd7/python/langsmith/client.py#L685
so it shouldn't be here, right?
```suggestion
```
(same for all calls to `_ensure_thread_running`)

PR #1650, opened 2025-04-09
patch[python]: Don't shut down compression thread if drain fails

angus-langchain:
### Purpose
If we fail to drain the compression buffer, we shouldn't shut down the background thread. We should also add logging if draining the buffer fails.


python/langsmith/_internal/_background_thread.py
```diff
@@ -102,41 +102,47 @@ def _tracing_thread_drain_queue(
...
 
-        client.compressed_traces.reset()
+            client.compressed_traces.reset()
 
-    filled_buffer.seek(0)
-    return (filled_buffer, compressed_traces_info)
+        filled_buffer.seek(0)
+        return (filled_buffer, compressed_traces_info)
+    except Exception:
+        logger.error("Error in tracing queue", exc_info=True)
```
hinthornw, line 142: Could we make  this more explicit so the user knows it isn't impacting their runtime, but rather indicates a langsmith trace may have failed to submit?

Want to avoid having people remove langsmith because they think it's causing errors

hinthornw (approved):

PR #1653, opened 2025-04-10
add fallbacks for create/update examples

isahers1:


python/langsmith/client.py
```diff
@@ -4535,11 +4535,32 @@ def create_examples(
                 )
             ]
 
-        return self._upload_examples_multipart(
-            dataset_id=cast(uuid.UUID, dataset_id),
-            uploads=uploads,
-            dangerously_allow_filesystem=dangerously_allow_filesystem,
-        )
+        try:
```
baskaryan, line 4538: instead of try excepting can we just do whatever check we're doing in _upload_examples_multipart?

python/langsmith/client.py
```diff
@@ -4535,11 +4535,32 @@ def create_examples(
...
+            return self._upload_examples_multipart(
+                dataset_id=cast(uuid.UUID, dataset_id),
+                uploads=uploads,
+                dangerously_allow_filesystem=dangerously_allow_filesystem,
+            )
+        except Exception:
```
agola11, line 4544: this is too broad, catch ValueError

python/langsmith/client.py
isahers1: this has the same effect no? the check is the first line of _upload_examples_multipart, so that check gets made and if it fails we immediately go to the except case

python/langsmith/client.py
```diff
@@ -4535,11 +4535,32 @@ def create_examples(
...
+            for upload in uploads:
+                if dump_model(upload).get("attachments", None):
+                    raise ValueError(
+                        "Must upgrade your LangSmith version to use attachments"
+                    )
```
agola11, line 4549: why do you need to do a `dump_model` here? Also if attachments aren't present just discard the attachments and log a warning

python/langsmith/client.py
baskaryan: yea but currently this woudl catch any exception, include unrelated ones. also just cleaner to check for a specific thing than to try/except

python/langsmith/client.py
```diff
@@ -4620,8 +4641,34 @@ def create_example(
             if example_id
             else uuid.uuid4()
         )
-        self._upload_examples_multipart(dataset_id=dataset_id, uploads=[data])
-        return self.read_example(example_id=data.id)
+
+        try:
+            self._upload_examples_multipart(dataset_id=dataset_id, uploads=[data])
+            return self.read_example(example_id=data.id)
+        except Exception:
```
agola11, line 4648: again, excep[tion handling too broad, catch ValueError

python/tests/integration_tests/test_client.py
```diff
@@ -73,7 +73,21 @@ def langchain_client() -> Client:
     )
 
 
-def test_datasets(langchain_client: Client) -> None:
+@pytest.fixture(params=[True, False])
+def parameterized_multipart_client(request) -> Client:
```
baskaryan, line 77: nifty

python/tests/integration_tests/test_client.py
```diff
@@ -73,7 +73,21 @@ def langchain_client() -> Client:
...
+    """Client fixture with parameterized dataset_examples_multipart_enabled setting."""
+    get_env_var.cache_clear()
+    return Client(
+        info={
+            "instance_flags": {
+                "dataset_examples_multipart_enabled": request.param,
+                "examples_multipart_enabled": True,
```
baskaryan, line 84: whats the diff b/n the two flags?

isahers1: the first one is all that is used today, there was a brief period of time where we used the second one and it just exists for backwards compatibility

isahers1:
integration tests errors are just 429s from openai (we should fix this at some point)

python/langsmith/client.py
```diff
@@ -4535,11 +4536,35 @@ def create_examples(
...
+                headers={**self._headers, "Content-Type": "application/json"},
+                data=_dumps_json(
+                    [
+                        {**dump_model(upload), "dataset_id": str(dataset_id)}
+                        for upload in uploads
+                    ]
+                ),
+            )
+            ls_utils.raise_for_status_with_text(response)
+            return None
```
baskaryan, line 4567: should we return response.json()?

python/langsmith/client.py
```diff
@@ -4976,19 +5027,36 @@ def update_example(
             **{k: v for k, v in example_dict.items() if v is not None}
         )
 
-        if dataset_id is not None:
+        if dataset_id is None:
```
baskaryan, line 5030: unrelated but ooc is there a reason update_example doesn't just call update_examples?

baskaryan (approved):

python/langsmith/client.py
isahers1: no, it probably should

PR #1655, opened 2025-04-10
patch[python]: Use multipart EP for feedback creation

angus-langchain:
### Purpose
All feedback ingest should go through multipart, in order to do this, we need to set the`trace_id`.

### DIFF
--- a/python/langsmith/testing/_internal.py
+++ b/python/langsmith/testing/_internal.py
@@ -464,13 +464,13 @@ def _end_tests(test_suite: _LangSmithTestSuite):
         test_suite.client.update_dataset_tag(
             dataset_id=dataset_id,
             as_of=dataset_version,
-            tag=f'git:commit:{git_info["commit"]}',
+            tag=f"git:commit:{git_info['commit']}",
         )
     if dataset_version and git_info["branch"] is not None:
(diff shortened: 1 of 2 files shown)


python/langsmith/testing/_internal.py
```diff
@@ -560,7 +560,7 @@ def submit_result(
         self._executor.submit(self._submit_result, run_id, score)
 
     def _submit_result(self, run_id: uuid.UUID, score: Optional[int]) -> None:
-        self.client.create_feedback(run_id, key="pass", score=score)
+        self.client.create_feedback(run_id, key="pass", score=score, trace_id=run_id)
```
angus-langchain, line 563: This is the only changed line, the rest is just formatting

jacoblee93 (approved):
Big w but why so many formatting changes?

python/langsmith/testing/_internal.py
jacoblee93, line 563: @akira asks you to add a comment explaining that `trace_id` will always be `run_id` here because we're leaving feedback on the root test run

angus-langchain: done

PR #1659, opened 2025-04-14
patch[python, js]: Update runs with project name

angus-langchain:
Currently, patch requests don't carry information about the project they are associated with. This causes upstream issues on ingestion when the patch is sent separately and we don't know the trace tier of the project, and thus, assign it the default tracing project tier. The result of this is that we upgrade the s3 urls post ingestion API: resulting in higher storage costs and an increase in class A operations.

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "langsmith",
-  "version": "0.3.15",
+  "version": "0.3.16",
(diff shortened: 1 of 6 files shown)


jacoblee93 (approved):
Don't forget to run `yarn format` on JS

hinthornw (approved):

PR #1668, opened 2025-04-16
feat(py,js): Allow logging feedback from nested traceables within pytest and jestlike tests

jacoblee93:


python/langsmith/testing/_internal.py
```diff
@@ -1284,7 +1284,7 @@ def test_openai_says_hello():
     metadata = {
         "experiment": test_case.test_suite.experiment.name,
         "reference_example_id": test_case.example_id,
-        "reference_run_id": parent_run.id,
+        "reference_run_id": test_case.run_id,
```
baskaryan, line 1287: does line 1291 also need to be updated

jacoblee93: Nice catch, yes

Although now that I look at it - why is this even a run to begin with?

baskaryan (approved):

jacoblee93:
Thanks! Failing tests are OpenAI rate limits so won't block on those

PR #1673, opened 2025-04-18
feat(js): Adds JS debug mode that logs output request URLs and responses

jacoblee93:
Enabled by passing `debug: true` to a client or by setting `LANGSMITH_DEBUG="true"`

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "langsmith",
-  "version": "0.3.18",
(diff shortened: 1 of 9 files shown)


jacoblee93:
Example (ignore the Jest formatting on multiple lines, outside of test cases logs will show up as consecutive):

Screenshot 2025-04-18 at 9 30 12 AM

angus-langchain (approved):

PR #1675, opened 2025-04-21
feat(js) Add function to get Group Runs

tendev-eoghan:
Add GroupRunsParams interface and implement groupRuns method for run grouping

Function implemented for [Group Runs](https://api.smith.langchain.com/redoc#tag/run/operation/group_runs_api_v1_runs_group_post)

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -205,7 +205,12 @@ interface GroupRunsParams {
   /**
    * The ID or IDs of the project(s) to filter by.
    */
-  projectId: string;
+  projectId?: string;
+
+  /**
+   * The ID or IDs of the project(s) to filter by.
+   */
+  projectName?: string;
 
   /**
    * @example "conversation"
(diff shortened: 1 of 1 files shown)


js/src/client.ts
```diff
@@ -309,6 +372,22 @@ type MultipartPart = {
   payload: Blob;
 };
 
+type Thread = {
```
jacoblee93, line 375: Think this should be exported?

js/src/client.ts
```diff
@@ -1613,6 +1694,73 @@ export class Client implements LangSmithTracingClientInterface {
...
+    const url = `${this.apiUrl}${path}`;
+
+    while (true) {
+      const currentBody = {
+        ...baseBody,
+        offset: currentOffset,
+      };
+
+      // Remove undefined values from the payload
+      const filteredPayload = Object.fromEntries(
```
jacoblee93, line 1732: Is this necessary?

js/src/client.ts
```diff
@@ -1613,6 +1694,73 @@ export class Client implements LangSmithTracingClientInterface {
     }
   }
 
+  public async *groupRuns(props: GroupRunsParams): AsyncIterable<Thread> {
```
jacoblee93, line 1697: Could we call this `listGroupRuns` instead?

js/src/client.ts
tendev-eoghan: Maybe not, I copied it from `getRunStats`.

https://github.com/langchain-ai/langsmith-sdk/blob/6368413746770abb7b171437c778791b6d111e12/js/src/client.ts#L1832

jacoblee93: Think it's Pythonic but shouldn't cause problems

jacoblee93 (approved):
Thanks!

PR #1676, opened 2025-04-21
chore[python]: Set dotted order for openai agents sdk integration

angus-langchain:
Currently our openai agents sdk calls create/update run methods without `trace_id` and `dotted_order`, which means tracing will happen in the main thread.

This fix sets those variables in the agents integration

Issue discovered in: https://github.com/openai/openai-agents-python/issues/529

### DIFF
--- a/python/langsmith/wrappers/_agent_utils.py
+++ b/python/langsmith/wrappers/_agent_utils.py
@@ -223,18 +223,15 @@ if HAVE_AGENTS:
 
         return data
 
-    def _create_current_dotted_order(
-        start_time: Optional[datetime], run_id: Optional[str]
+    def ensure_dotted_order(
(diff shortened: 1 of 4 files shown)


python/langsmith/wrappers/_agent_utils.py
```diff
@@ -220,3 +222,19 @@ def extract_span_data(span: tracing.Span) -> Dict[str, Any]:
             return {}
 
         return data
+
+    def _create_current_dotted_order(
```
baskaryan, line 226: ooc do we not have an existing util for this anywhere

angus-langchain: we do on the runtree Object.

baskaryan (approved):

PR #1683, opened 2025-04-23
fix: don't allow example creation with incorrect kwargs

isahers1:


python/langsmith/schemas.py
```diff
@@ -127,10 +129,17 @@ class ExampleCreate(BaseModel):
     use_source_run_attachments: Optional[list[str]] = None
     source_run_id: Optional[UUID] = None
 
+    model_config = ConfigDict(extra="forbid")
```
baskaryan, line 132: should we also make `inputs` required?

isahers1: yeah good point

PR #1684, opened 2025-04-23
fix(py): try/except writing to multiple endpoints

baskaryan:
prevent one run creation failure from block all run creations, in case of multiple write endpoints


python/tests/integration_tests/test_async_client.py
```diff
@@ -16,6 +16,7 @@ class InputsSchema(BaseModel):
         name: str  # type: ignore[annotation-unchecked]
         age: int  # type: ignore[annotation-unchecked]
 
+    dataset = None
```
baskaryan, line 19: unrelated to this pr but exposed by current issues with staging

angus-langchain (approved):

PR #1705, opened 2025-04-29
feat: support rubric_instructions for annotation queues in the SDK

samnoyes:


js/src/client.ts
```diff
@@ -3832,17 +3835,24 @@ export class Client implements LangSmithTracingClientInterface {
...
-    const queueIteratorResult = await this.listAnnotationQueues({
-      queueIds: [queueId],
-    }).next();
-    if (queueIteratorResult.done) {
-      throw new Error(`Annotation queue with ID ${queueId} not found`);
-    }
-    return queueIteratorResult.value;
+  public async readAnnotationQueue(
+    queueId: string
+  ): Promise<AnnotationQueueDetails> {
```
jacoblee93, line 3842: Don't think I like this name, sounds like it's just returning metadata about an annotation queue

Why not just add it as optional to the `AnnotationQueue` schema? If it's not present on list, then how about `AnnotationQueueWithDetails`

js/src/client.ts
```diff
@@ -3857,16 +3867,21 @@ export class Client implements LangSmithTracingClientInterface {
...
       _getFetchImplementation(this.debug),
       `${this.apiUrl}/annotation-queues/${assertUuid(queueId, "queueId")}`,
       {
         method: "PATCH",
         headers: { ...this.headers, "Content-Type": "application/json" },
-        body: JSON.stringify({ name, description }),
+        body: JSON.stringify({
+          name,
+          description,
+          rubric_instructions: rubricInstructions,
```
jacoblee93, line 3883: I would double check this doesn't unset previous instructions if not passed, have had a few multipart bugs where that happens recently

js/src/schemas.ts
```diff
@@ -567,6 +567,11 @@ export interface AnnotationQueue {
   tenant_id: string;
 }
 
+export interface AnnotationQueueDetails extends AnnotationQueue {
```
jacoblee93, line 570: See above

python/langsmith/client.py
```diff
@@ -6259,7 +6259,8 @@ def create_annotation_queue(
         name: str,
         description: Optional[str] = None,
         queue_id: Optional[ID_TYPE] = None,
-    ) -> ls_schemas.AnnotationQueue:
+        rubric_instructions: Optional[str] = None,
+    ) -> ls_schemas.AnnotationQueueDetails:
```
jacoblee93, line 6263: See above

js/src/client.ts
samnoyes: yeah i didn't do that because not present on list so didnt want it to be confusing - will update to `AnnotationQueueWithDetails`

js/src/client.ts
samnoyes: synced offline: our backend actually doesnt support this anyway (so if you don't pass rubric_instructions, it sets it to None)

jacoblee93 (approved):

PR #1708, opened 2025-05-01
fix(js): Fix AI SDK tracing when nested in AI SDK tool calls

jacoblee93:


jeffbarg (approved):
LGTM!

js/src/tests/vercel_node_otel.int.test.ts
```diff
@@ -0,0 +1,108 @@
...
+  );
+
+  await wrapper();
+
+  await sdk.shutdown();
+
+  await waitUntilRunFound(client, runId, true);
+
+  const storedRun = await client.readRun(runId);
+  expect(storedRun.id).toEqual(runId);
```
dqbd, line 107: Worth adding a test with `getAssumedTreeFromCalls(callSpy.mock.calls)`? Here we're only checking that the run has been created.

dqbd (approved):
Approving to unblock

PR #1709, opened 2025-05-01
fix(js): Store dotted order for runs separately in case OTEL export happens in multiple batches

jacoblee93:


angus-langchain (approved):

jacoblee93:
Sorry @angus-langchain sadly this doesn't seem to work properly yet :/ will probably ask for your review again later

js/src/tests/vercel.test.ts
```diff
@@ -941,7 +948,7 @@ test("traceable", async () => {
   });
 });
 
-test("unrelated spans around", async () => {
+test.skip("unrelated spans around", async () => {
```
jacoblee93, line 951: @dqbd this one is now failing but I think the new behavior is correct - would love to chat with you about it

dqbd (approved):

PR #1712, opened 2025-05-02
fix(js): Trace AI SDK OTEL exported images as base64

jacoblee93:


js/src/vercel.ts
```diff
@@ -124,9 +124,39 @@ function convertCoreToSmith(
...
+            if (
+              typeof imageUrl === "object" &&
+              "type" in imageUrl &&
+              "data" in imageUrl
+            ) {
+              // Typing is wrong here if a buffer is passed in
+              uint8Array = new Uint8Array(imageUrl.data as Uint8Array);
+            } else if (
+              typeof imageUrl === "object" &&
+              Object.keys(imageUrl).every((key) => !isNaN(Number(key)))
```
hinthornw, line 141: Didn't realize this was a popular format

js/src/vercel.ts
```diff
@@ -124,9 +124,39 @@ function convertCoreToSmith(
...
+              uint8Array = new Uint8Array(
+                Array.from({
+                  ...imageUrl,
+                  length: Object.keys(imageUrl).length,
+                })
+              );
+            }
+            if (uint8Array) {
+              let binary = "";
+              for (let i = 0; i < uint8Array.length; i++) {
```
hinthornw, line 153: Is this better/equivalent to like ` String.fromCharCode.apply(`?

js/src/vercel.ts
jacoblee93: It's not, something is getting badly converted/serialized somewhere in the OTEL bowels

js/src/vercel.ts
jacoblee93: Don't think there's a difference, I don't like to get cute when I can help it

PR #1730, opened 2025-05-13
Remove default for fetching child_run_ids from the SDK

akira:
* Make `child_run_ids` not a default field in select.
* Also fix the `load_child_runs` path to use query by trace.


jacoblee93 (commented):
Looks fine, would suggest an integration test or two as well to verify but I can take that on if you'd like

jacoblee93 (approved):

jacoblee93:
Looks like Python async client doesn't support child run fetching yet so we should fix that separately

PR #1746, opened 2025-05-19
feat: Add process_usage method for traceable

jacoblee93:


hinthornw:
Would be kinda nice to add to the run_tree api.

```python
def foo():
    if (rt := ls.get_current_run_tree()):
        rt.add_usage(...)
```
or
```python
def foo():
    if (rt := ls.get_current_run_tree()):
        rt.set(inputs=..., outputs=..., usage=...)
```

jacoblee93:
> Would be kinda nice to add to the run_tree api.
> 
> ```python
> def foo():
>     if (rt := ls.get_current_run_tree()):
>         rt.add_usage(...)
> ```
> 
> or
> 
> ```python
> def foo():
>     if (rt := ls.get_current_run_tree()):
>         rt.set(inputs=..., outputs=..., usage=...)
> ```

Basically just as shorthand?

I suppose could do that yeah

python/langsmith/schemas.py
```diff
@@ -1109,6 +1109,45 @@ class UsageMetadata(TypedDict):
...
+    """The cost of the output tokens."""
+    total_cost: NotRequired[Decimal]
+    """The total cost of the tokens."""
+    input_cost_details: NotRequired[dict[str, Any]]
+    """The cost details of the input tokens."""
+    output_cost_details: NotRequired[dict[str, Any]]
+    """The cost details of the output tokens."""
+
+
+class ExtractedUsageMetadata(TypedDict, total=False):
```
baskaryan, line 1124: not sure i follow, doesn't UsageMetadata already not require all keys to be present? is the issue that we want to make input_tokens/output_tokens/etc also not required?

python/langsmith/run_trees.py
```diff
@@ -210,6 +211,10 @@ def set(
                 self.outputs = {}
             else:
                 self.outputs = dict(outputs)
+        if usage is not NOT_PROVIDED:
+            self.extra.setdefault("metadata", {}).setdefault(
+                "usage_metadata", {}
+            ).update(usage or {})
```
baskaryan, line 217: nit, should we set this instead of updating? think id expect a `set` method to fully override any existing values

python/langsmith/run_trees.py
```diff
@@ -173,6 +173,7 @@ def set(
         outputs: Optional[Mapping[str, Any]] = NOT_PROVIDED,
         tags: Optional[Sequence[str]] = NOT_PROVIDED,
         metadata: Optional[Mapping[str, Any]] = NOT_PROVIDED,
+        usage: Optional[ls_schemas.ExtractedUsageMetadata] = NOT_PROVIDED,
```
baskaryan, line 176: thoughts on calling it usage_metadata? to match the actual metadata key

python/langsmith/run_helpers.py
```diff
@@ -288,6 +311,9 @@ def traceable(
     process_inputs: Optional[Callable[[dict], dict]] = None,
     process_outputs: Optional[Callable[..., dict]] = None,
     process_chunk: Optional[Callable] = None,
+    extract_usage: Optional[
```
baskaryan, line 314: think we drop this for now and think some more about the ideal devx here

python/langsmith/run_trees.py
jacoblee93: You are probably right yeah

python/langsmith/run_trees.py
jacoblee93: Yeah probably a better idea - was hoping we could transition to just `usage` in the future but probably better to do it all at once

python/langsmith/schemas.py
```diff
@@ -1109,6 +1109,45 @@ class UsageMetadata(TypedDict):
...
+    """Usage metadata dictionary extracted from a run.
+
+    Should be the same as UsageMetadata, but does not require all
+    keys to be present.
+    """
+
+    prompt_tokens: int
+    """The number of tokens used for the prompt."""
+    completion_tokens: int
```
baskaryan, line 1133: these should be input_tokens and output_tokens

baskaryan (approved):

PR #1754, opened 2025-05-28
Add include_extra to let get_run_from_annotation_queue return metadata

EugeneJinXin:
Issue: RunWithAnnotationQueueInfo had no way to return any run metadata.  Part of Goalie [ticket](https://linear.app/langchain/issue/LS-3144/runwithannotationqueueinfo-should-return-run-metadata) asked by customer. 

Fix: With this change, the SDK side will have flexibility to get extra (metadata) if they want to, by setting include_extra in their client. get_run_from_annotation_queue, for example:
`run = client.get_run_from_annotation_queue(queue_id="xxx", index=0, **include_extra=True**)` 

Test: adding UnitTests to corresponding ls.client  and async client


EugeneJinXin:
Failed Integration Test seems expected, because it's hitting real langsmith prod service so i need the[ backend PR](https://github.com/langchain-ai/langchainplus/pull/9913) to land first

PR #1759, opened 2025-05-30
fix(py): Fix URL construction in attachment downloading

anceschia:
### Problem
The `_convert_stored_attachments_to_attachments_dict` function was attempting to download attachments using `requests.get(value["presigned_url"])`, but `value["presigned_url"]` often contains only an API path (e.g., `/api/public/download?jwt=...`) rather than a complete URL. This caused `requests.get()` to fail with connection errors when trying to download attachment data.

### Solution
- **Modified function signature**: Updated `_convert_stored_attachments_to_attachments_dict` to accept an optional `api_url` parameter
- **Added URL construction logic**: When `api_url` is provided, the function now uses the existing `_construct_url` helper to properly construct the complete URL by combining the API base URL with the attachment path
- **Backward compatibility**: Made `api_url` optional (defaults to `None`) so the function can handle cases where `presigned_url` is already a complete URL
- **Updated call sites**: Modified `list_examples()` and `read_example()` call sites within the `Client` class to pass `self.api_url` as the `api_url` parameter

### DIFF
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -4757,7 +4757,7 @@ class Client:
 
         example = response.json()
         attachments = _convert_stored_attachments_to_attachments_dict(
-            example, attachments_key="attachment_urls"
+            example, attachments_key="attachment_urls", api_url=self._api_url
         )
 
         return ls_schemas.Example(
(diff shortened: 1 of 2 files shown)


anceschia:
> lgtm, could you add a test before I approve and merge?

Done 👍🏻

PR #1762, opened 2025-05-30
feat:  OTEL_ENABLED mode sends to both otel and ls, while allowing otel only through OTEL_ONLY var

EugeneJinXin:
# Hybrid Tracing: Overview

### Before: Choose One
```
LANGSMITH_OTEL_ENABLED = "true"
traceable_function() 
    │
    ▼
┌─────────────┐
│ Background  │ 
│   Thread    │
└─────┬───────┘
      │
      ▼
   Either:
 ┌──────────┐  OR  ┌─────────────┐
 │   OTEL   │      │ LangSmith   │
 │ Endpoint │      │  Endpoint   │
 └──────────┘      └─────────────┘
```

### After: 
Default behavior of LANGSMITH_OTEL_ENABLED: Get Both
```

traceable_function() 
    │
    ▼
LANGSMITH_OTEL_ENABLED = "true"
┌─────────────┐
│ Background  
│   Thread    │
└─────┬───────┘
      │
      ▼
   Both:
 ┌──────────┐ AND ┌─────────────┐
 │   OTEL   │     │ LangSmith   │
 │ Endpoint │     │  Endpoint   │
 └──────────┘     └─────────────┘
```

User can do Otel-only if insist to, through additional env var LANGSMITH_OTEL_ONLY
```

traceable_function() 
    │
    ▼
LANGSMITH_OTEL_ENABLED = "true" and LANGSMITH_OTEL_ONLY = "true"
┌─────────────┐
│ Background  
│   Thread    │
└─────┬───────┘
      │
      ▼
   ONLY OTEL:
 ┌──────────┐ 
 │   OTEL   │     
 │ Endpoint │   
 └──────────┘    
```

###  ** Increase Tracing Coverage**
- Get the best of both worlds
- No more choosing between tools
- Future-proof observability

###  **Risk Mitigation** 
- If one service is down, you still have traces
- Backup observability data

## Technical detail for reviewer
- A new `_hybrid_tracing_thread_handle_batch()` to make sure call task_done() once per item, regardless of how many exporters succeeded
- Sub-thread spawning logic also adopted this. 
- Future improvement can work on the special case of feedback handling, and whether needs to tune up performance of these two parallel exports by using two worker threads etc.

## Test plan
- 11 test cases covering three main scenarios: hybrid mode (OTEL_ENABLED=true) that exports to both OTEL and LangSmith simultaneously, OTEL-only mode (OTEL_ONLY=true) that exports exclusively to OTEL, and LangSmith-only mode (default) that exports only to LangSmith. 
- Also i did a local E2E test, setting up a jaegar and tested the (three mode) * 2 way (from https://docs.smith.langchain.com/observability/how_to_guides/trace_langchain_with_otel#configuring-alternate-otlp-endpoints)  and confirmed results through manually inspection , i have https://docs.google.com/document/d/1hgS-YPAqvavbkYN0DyrFkjd91lD-n0n6kAedAfAiCFM/edit?tab=t.0 which contains all the scenarios i tested


python/langsmith/_internal/_background_thread.py
```diff
@@ -180,23 +181,21 @@ def _tracing_thread_handle_batch(
             "Error details:",
             exc_info=True,
         )
-        # exceptions are logged elsewhere, but we need to make sure the
-        # background thread continues to run
-        pass
     finally:
-        for _ in batch:
-            tracing_queue.task_done()
+        if mark_task_done:
```
EugeneJinXin, line 203: FYI: this is to let `_hybrid_tracing_thread_handle_batch()` reuse `_otel_tracing_thread_handle_batch` and `_tracing_thread_handle_batch`, without double marking task done

EugeneJinXin:
Marking ready for review since the test case that is failling seems due to Integration test got rate limited,  irrelevant to the PR. They should pass once we wait a bit and re-run

python/langsmith/_internal/_background_thread.py
```diff
@@ -351,24 +454,56 @@ def keep_thread_active() -> bool:
...
+            # OTEL_ENABLED=true enables hybrid mode by default
+            if next_batch := _tracing_thread_drain_queue(
+                tracing_queue, limit=size_limit
+            ):
+                _hybrid_tracing_thread_handle_batch(
+                    client, tracing_queue, next_batch, use_multipart
+                )
+        else:
+            # Either OTEL-only mode (OTEL_ONLY=true) or LangSmith-only mode
+            if next_batch := _tracing_thread_drain_queue(
```
angus-langchain: nit: you drain for both hybrid and single tracing paths

python/langsmith/_internal/_background_thread.py
```diff
@@ -220,12 +229,106 @@ def _otel_tracing_thread_handle_batch(
...
+                    else:
+                        raise
+
+
+def _hybrid_tracing_thread_handle_batch(
+    client: Client,
+    tracing_queue: Queue,
+    batch: list[TracingQueueItem],
+    use_multipart: bool,
+    mark_task_done: bool = True,
```
angus-langchain, line 286: is this every set to false other than in the single path batch handlers?

python/langsmith/_internal/_background_thread.py
```diff
@@ -220,12 +229,106 @@ def _otel_tracing_thread_handle_batch(
...
+
+        # Get the global TracerProvider and check its resource attributes
+        tracer_provider = trace.get_tracer_provider()
+        if hasattr(tracer_provider, "resource") and hasattr(
+            tracer_provider.resource, "attributes"
+        ):
+            is_internal = tracer_provider.resource.attributes.get(
+                "langsmith.internal_provider", False
+            )
+            logger.debug(
```
angus-langchain, line 371: nice

python/langsmith/_internal/_background_thread.py
```diff
@@ -220,12 +229,106 @@ def _otel_tracing_thread_handle_batch(
...
+        )
+        _otel_tracing_thread_handle_batch(
+            client, tracing_queue, batch, mark_task_done=False
+        )
+    else:
+        logger.debug(
+            "Using user-provided TracerProvider, sending to both OTEL and LangSmith"
+        )
+        # Export to OTEL
+        _otel_tracing_thread_handle_batch(
```
angus-langchain: should we do these in parallel since they're both I/O bound?

angus-langchain: I would use a threadpoolexecutor here

python/langsmith/_internal/_background_thread.py
```diff
@@ -351,24 +454,56 @@ def keep_thread_active() -> bool:
             )
             sub_threads.append(new_thread)
             new_thread.start()
-        if next_batch := _tracing_thread_drain_queue(tracing_queue, limit=size_limit):
+        if ls_utils.is_truish(
+            ls_utils.get_env_var("OTEL_ENABLED")
```
angus-langchain: nit: create util function for this repeated check: get_tracing_mode()

python/langsmith/_internal/_background_thread.py
EugeneJinXin: thx, fixed

python/langsmith/_internal/_background_thread.py
EugeneJinXin: great idea! thought about it but didn't try at first, adopted now!

python/langsmith/_internal/_background_thread.py
EugeneJinXin: fantastic, greatly helped. adopted this in my new commit!

python/langsmith/_internal/_background_thread.py
EugeneJinXin: it's set to false in used inside _hybrid_tracing_thread_handle_batch mainly, any concern or suggestion?

python/tests/unit_tests/test_run_helpers.py
```diff
@@ -542,8 +542,8 @@ def my_function(a: int) -> int:
                 assert body["post"]
                 posts.extend(body["post"])
         assert len(posts) == 2
-        parent = next(p for p in posts if p["parent_run_id"] is None)
-        child = next(p for p in posts if p["parent_run_id"] is not None)
+        parent = next(p for p in posts if p.get("parent_run_id") is None)
```
angus-langchain, line 545: why did we need to update tests? Were these failing on main?

EugeneJinXin: i think it's a small bug irrelevant to my test (i say irrelevant because looking at bagatur's other PR today, the test case is breaking his side as well https://github.com/langchain-ai/langsmith-sdk/pull/1774) so it's breaking main branch

python/langsmith/__init__.py
```diff
@@ -20,7 +20,8 @@
     from langsmith.utils import ContextThreadPoolExecutor
 
 # Avoid calling into importlib on every call to __version__
-__version__ = "0.3.45"
+
+__version__ = "0.4.0"
```
angus-langchain, line 24: we will need to update langchain deps since this is a minor bump

EugeneJinXin: plan to do so today, and langsmith-doc

PR #1767, opened 2025-06-03
make evaluators more flexible

isahers1:


python/langsmith/evaluation/evaluator.py
```diff
@@ -263,13 +263,69 @@ def _coerce_evaluation_result(
                     "Expected an EvaluationResult object, or dict with a metric"
                     f" 'key' and optional 'score'; got empty result: {result}"
                 )
+            if (
+                sum(
```
vbarda: (nit) can probably do `len(set(k for k in ... if k in result))`

python/langsmith/evaluation/evaluator.py
```diff
@@ -263,13 +263,69 @@ def _coerce_evaluation_result(
...
+                raise ValueError(
+                    "Can only upload one of 'comment', 'justification', 'reasoning', "
+                    "'reason'"
+                )
             if "key" not in result and allow_no_key:
-                result["key"] = self._name
-            if all(k not in result for k in ("score", "value", "comment")):
+                non_accepted_keys = list(
+                    filter(
+                        lambda k: k
```
vbarda: (nit) better to avoid lambdas and filter explicitly

python/tests/evaluation/test_evaluation.py
```diff
@@ -548,3 +548,58 @@ async def mock_correctness_evaluator(outputs: dict, reference_outputs: dict):
 
     finish_time = time.time()
     assert (finish_time - start) <= 8.5
+
+
+def test_evaluate_with_new_accepted():
```
vbarda: (nit) better name?

python/langsmith/evaluation/evaluator.py
```diff
@@ -263,13 +263,63 @@ def _coerce_evaluation_result(
                     "Expected an EvaluationResult object, or dict with a metric"
                     f" 'key' and optional 'score'; got empty result: {result}"
                 )
+            if (
+                len(set(result) & {"comment", "justification", "reasoning", "reason"})
```
jacoblee93, line 267: Why so many aliases? Agree `comment` isn't the best but we should pick one alternative and stick with it no?

isahers1: just to give people flexibility? I think this isn't crazy imo. If our docs were perfect then I would be fine sticking with 1 but I feel like people could very reasonably expect to use any of these.

jacoblee93: I mean why stop at 4 then?

Really should be exactly one or one plus a deprecated name. Only reason to broaden/alias this would be if we needed to accept a format from some other source or if there were multiple conceptual things going on, which is not the case

PR #1770, opened 2025-06-05
feat[python]: enable tracing to multiple projects

angus-langchain:


lc-arjun (approved):

python/langsmith/run_trees.py
```diff
@@ -400,12 +413,66 @@ def _get_dicts_safe(self):
...
+        if self.replicas:
+            for project_name, updates in self.replicas:
+                # Avoid double posting for the current project
+                if project_name == self.session_name:
+                    run_dict = self._get_dicts_safe()
+                else:
+                    run_dict = self._remap_for_project(project_name, updates)
+                self.client.create_run(**run_dict)
+        else:
+            kwargs = self._get_dicts_safe()
```
jacoblee93, line 472: I think this `else` will prevent posting to the default project if there are replicas

May be desired behavior

lc-arjun: yes, this is the desired behavior

python/langsmith/run_trees.py
```diff
@@ -400,12 +413,66 @@ def _get_dicts_safe(self):
...
+            {
+                "id": new_id,
+                "trace_id": new_trace,
+                "parent_run_id": new_parent,
+                "dotted_order": dotted,
+                "session_name": project_name,
+            }
+        )
+        if updates:
+            dup.update(updates)
```
jacoblee93, line 458: Why not just do this once in `post`?

Also, this will not apply if you're posting a replica to the same project as `project_name`

python/langsmith/run_trees.py
jacoblee93: Ah ok, do we need the explicit project check then?

PR #1783, opened 2025-06-10
 feat[python]: Enable OpenTelemetry trace context propagation

### LINKED ISSUE #1725: Issue: [python] Trace context not propagated into OpenTelemetry spans

First off, I'm excited to see Langsmith tracing interoperating with OpenTelemetry! Thank you for working on this.

This looks similar to https://github.com/langchain-ai/langsmith-sdk/issues/1682. I'm finding that OTel spans are not being correctly parented by Langsmith spans, for example when my traced tools get executed. For example using the below tool with LangGraph's ReAct prebuilt agent

```py
@tool
@tracer.start_as_current_span("OTel get_greeting")
def get_greeting() -> str:
    """Get a greeting to give to the user"""
    return "Hello there 😊"

model = ChatVertexAI(model=model_name)
checkpointer = InMemorySaver()
agent = create_react_agent(model, tools=[get_greeting])
agent.invoke(...)
```

I get a trace like this:

Image

I haven't tested this with async functions.

### Suggestion:

I'm guessing Langsmith is either
- not setting its spans in OTel's context and/or
- not running code with the OTel context [attached](https://opentelemetry-python.readthedocs.io/en/latest/api/context.html#opentelemetry.context.attach), possibly because of some kind of queuing.

angus-langchain:
### Purpose
trace context propagation problem. Previously, when using LangSMith in otel mode. OpenTelemetry spans within LangSmith-traced functions would create disconnected traces, making it difficult to understand the complete trace.
Now:
User OpenTelemetry spans now inherit LangSmith trace context
LangSmith spans nest within existing OpenTelemetry traces
 
 Fixes #1725

  ### Changes
  - Added deterministic UUID to OpenTelemetry ID conversion
  - Added trace context propagation for both langsmith and otel spans

  Basic Usage
```
  @traceable
  def process_data(data: str) -> str:
      # This OpenTelemetry span will inherit LangSmith trace context
      with tracer.start_as_current_span("otel_processing") as span:
          span.set_attribute("data.input", data)
          result = f"Processed: {data}"
          span.set_attribute("data.output", result)
          return result
```

### Tests
Added unit tests and tested locally with langchain, traceable, and otel spans

### DIFF
--- a/python/langsmith/_internal/_otel_utils.py
+++ b/python/langsmith/_internal/_otel_utils.py
@@ -1,5 +1,3 @@
-"""Utilities for OpenTelemetry integration with LangSmith."""
-
 from __future__ import annotations
 
 from typing import Optional
@@ -15,16 +13,11 @@ def uuid_to_otel_trace_id(uuid_val: UUID) -> str:
     Returns:
         A 32-character lowercase hex string representing the trace ID.
     """
-    # UUIDs are already 128 bits, which is what OTel trace IDs need
-    # Convert to 32-character lowercase hex string
     return uuid_val.hex
 
 
 def uuid_to_otel_span_id(uuid_val: UUID) -> str:
     """Convert a UUID to an OpenTelemetry span ID.
-
-    Since span IDs are 64 bits (8 bytes) and UUIDs are 128 bits (16 bytes),
-    we need to derive a 64-bit value deterministically from the UUID.
 
     Args:
         uuid_val: The UUID to convert.
@@ -32,10 +25,8 @@ def uuid_to_otel_span_id(uuid_val: UUID) -> str:
     Returns:
         A 16-character lowercase hex string representing the span ID.
     """
-    # Use the first 8 bytes of the UUID for deterministic span ID
     uuid_bytes = uuid_val.bytes
     span_id_bytes = uuid_bytes[:8]
-    # Convert to 16-character lowercase hex string
     return span_id_bytes.hex()
 
 
@@ -87,48 +78,3 @@ def get_otel_span_id_from_uuid(uuid_val: UUID) -> int:
     """
     span_id_hex = uuid_to_otel_span_id(uuid_val)
     return otel_span_id_to_int(span_id_hex)
-
-
-def get_otel_trace_id_from_run_tree() -> Optional[int]:
-    """Get OpenTelemetry trace ID from current LangSmith run tree.
-
-    Returns:
-        Integer representation of the trace ID, or None if no current run tree.
-    """
-    from langsmith.run_helpers import get_current_run_tree
-
-    current_run = get_current_run_tree()
-    if current_run is None:
-        return None
-
-    return get_otel_trace_id_from_uuid(current_run.trace_id)
-
-
-def get_otel_span_id_from_run_tree() -> Optional[int]:
-    """Get OpenTelemetry span ID from current LangSmith run tree.
-
-    Returns:
-        Integer representation of the span ID, or None if no current run tree.
-    """
-    from langsmith.run_helpers import get_current_run_tree
-
-    current_run = get_current_run_tree()
-    if current_run is None:
-        return None
-
-    return get_otel_span_id_from_uuid(current_run.id)
-
-
-def get_otel_parent_span_id_from_run_tree() -> Optional[int]:
-    """Get OpenTelemetry parent span ID from current LangSmith run tree.
-
-    Returns:
-        Integer representation of the parent span ID, or None if no parent.
-    """
-    from langsmith.run_helpers import get_current_run_tree
-
-    current_run = get_current_run_tree()
-    if current_run is None or current_run.parent_run_id is None:
-        return None
-
-    return get_otel_span_id_from_uuid(current_run.parent_run_id)
--- a/python/langsmith/_internal/_otel_utils.py
+++ b/python/langsmith/_internal/_otel_utils.py
@@ -1,0 +1,134 @@
+"""Utilities for OpenTelemetry integration with LangSmith."""
+
+from __future__ import annotations
+
+from typing import Optional
+from uuid import UUID
+
+
+def uuid_to_otel_trace_id(uuid_val: UUID) -> str:
+    """Convert a UUID to an OpenTelemetry trace ID.
+
+    Args:
+        uuid_val: The UUID to convert.
+
+    Returns:
+        A 32-character lowercase hex string representing the trace ID.
+    """
+    # UUIDs are already 128 bits, which is what OTel trace IDs need
+    # Convert to 32-character lowercase hex string
+    return uuid_val.hex
+
+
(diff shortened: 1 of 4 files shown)


python/langsmith/_internal/otel/_otel_exporter.py
```diff
@@ -183,18 +191,41 @@ def _create_span_for_run(
             end_time = run_info.get("end_time")
             end_time_utc_nano = self._as_utc_nano(end_time)
 
-            # Start the span
+            # Create deterministic trace and span IDs to match user OpenTelemetry spans
+            trace_id_int = get_otel_trace_id_from_uuid(op.trace_id)
```
EugeneJinXin, line 195: will it benefit if we reuse _maybe_create_otel_context ?

python/langsmith/run_helpers.py
```diff
@@ -1863,3 +1916,46 @@ def _cleanup_traceback(e: BaseException):
...
+    span_context = SpanContext(
+        trace_id=trace_id,
+        span_id=span_id,
+        is_remote=False,
+        trace_flags=TraceFlags(TraceFlags.SAMPLED),
+        trace_state=TraceState(),
+    )
+
+    non_recording_span = NonRecordingSpan(span_context)
+    return use_span(non_recording_span)
```
EugeneJinXin, line 2014: just double check, we don't need to set 'end_on_exit=True' param in use_span(), right?

python/langsmith/run_helpers.py
```diff
@@ -1863,3 +1916,46 @@ def _cleanup_traceback(e: BaseException):
...
+
+    span_context = SpanContext(
+        trace_id=trace_id,
+        span_id=span_id,
+        is_remote=False,
+        trace_flags=TraceFlags(TraceFlags.SAMPLED),
+        trace_state=TraceState(),
+    )
+
+    non_recording_span = NonRecordingSpan(span_context)
```
EugeneJinXin, line 2013: i believe NonRecordingSpan won't be exported right? i'm asking because (TraceFlags.SAMPLED) made me not sure without testing mysefl

python/langsmith/run_helpers.py
angus-langchain: It won't apply as this is a non recording span

python/langsmith/run_helpers.py
angus-langchain: That flag just indicates that its part of a sampled trace, the non recoding span won't be exported

python/langsmith/_internal/otel/_otel_exporter.py
angus-langchain: no we need to set the context directly, not wrap it in a new context manager

EugeneJinXin (approved):

python/langsmith/_internal/_otel_utils.py
```diff
@@ -0,0 +1,79 @@
+from __future__ import annotations
+
+from uuid import UUID
+
+
+def uuid_to_otel_trace_id(uuid_val: UUID) -> str:
```
hinthornw: Seems a bit excessive sincve this is only used in one location?

python/langsmith/_internal/_otel_utils.py
```diff
@@ -0,0 +1,79 @@
...
+    Args:
+        uuid_val: The UUID to convert.
+
+    Returns:
+        A 32-character lowercase hex string representing the trace ID.
+    """
+    return uuid_val.hex
+
+
+def uuid_to_otel_span_id(uuid_val: UUID) -> str:
```
hinthornw: ditto here

python/tests/unit_tests/test_hybrid_tracing.py
```diff
@@ -442,3 +442,75 @@ def test_hybrid_tracing_graceful_cleanup(self, mock_session):
...
+    )
+    def test_otel_context_propagation_with_traceable(self):
+        """Test that OpenTelemetry context is properly set using NonRecordingSpan."""
+        pytest.importorskip("opentelemetry")
+
+        from langsmith import utils
+
+        utils.get_env_var.cache_clear()
+
+        from opentelemetry.sdk.trace import TracerProvider
```
hinthornw: why do we need to lazy import all these things? It's just a test

python/tests/unit_tests/test_hybrid_tracing.py
```diff
@@ -442,3 +442,75 @@ def test_hybrid_tracing_graceful_cleanup(self, mock_session):
                 )
 
             mock_session.request.assert_called()
+
+    @patch.dict(
+        os.environ, {"LANGSMITH_OTEL_ENABLED": "true", "LANGSMITH_TRACING": "true"}
+    )
+    def test_otel_context_propagation_with_traceable(self):
```
hinthornw, line 465: I think we need to test deeper. ls -> otel -> ls -> otel.

Also need an async test.

What happens for generators?  Like what happens if I have a generator function that traces with otel., i start it but don't fully consume it. Then i start a traceable  as a sibling, then i finish consuming the otel function w/ generator. What does that look like?

python/langsmith/run_helpers.py
```diff
@@ -1863,3 +1916,46 @@ def _cleanup_traceback(e: BaseException):
...
+        run_tree: The current run tree.
+
+    Returns:
+        Context manager for use_span or None if not available.
+    """
+    if not run_tree or not utils.is_truish(utils.get_env_var("OTEL_ENABLED")):
+        return None
+
+    try:
+        from opentelemetry.trace import (  # type: ignore[import]
```
hinthornw: Ithink we can speed this up by caching whether opentelemetry is installed

python/langsmith/run_helpers.py
```diff
@@ -1863,3 +1916,46 @@ def _cleanup_traceback(e: BaseException):
...
+        return None
+
+    trace_id = get_otel_trace_id_from_uuid(run_tree.trace_id)
+    span_id = get_otel_span_id_from_uuid(run_tree.id)
+
+    span_context = SpanContext(
+        trace_id=trace_id,
+        span_id=span_id,
+        is_remote=False,
+        trace_flags=TraceFlags(TraceFlags.SAMPLED),
```
hinthornw, line 2009: ooc: why sampled?

angus-langchain: This means the trace should be processed / not dropped

python/langsmith/run_helpers.py
angus-langchain: done

python/tests/unit_tests/test_hybrid_tracing.py
angus-langchain: If you create a generator function that you start, then create a sibling traceable, and then close the generator function; the sibling and generator context are separate, so the traces will not be combined

PR #1784, opened 2025-06-10
fix(js): Fix JSDom check for Node 22

jacoblee93:
Node 21 added the navigator interface, so the JSDom check in `env.ts` will always return true for recent Node versions:

https://developer.mozilla.org/en-US/docs/Web/API/Navigator#browser_compatibility

JSDom's default user agent contains the string "jsdom" so it still should be fine:

https://github.com/jsdom/jsdom?tab=readme-ov-file#advanced-configuration

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "langsmith",
-  "version": "0.3.30",
+  "version": "0.3.31",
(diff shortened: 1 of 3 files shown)


bracesproul (approved):
lgtm!

nfcampos (approved):

PR #1794, opened 2025-06-13
fix(js): Pass events through on run create

jacoblee93:
Seems like an oversight?

CC @lc-arjun


hinthornw (approved):
I think reason for not including originally is that events wouldn't usually be created until after the run was created anyway, so it's mainly used at patch time

PR #1796, opened 2025-06-16
feat(js): Add replica support for JS

jacoblee93:
CC @lc-arjun

### DIFF
--- a/js/src/run_trees.ts
+++ b/js/src/run_trees.ts
@@ -63,6 +63,7 @@ export interface RunTreeConfig {
   trace_id?: string;
   dotted_order?: string;
   attachments?: Attachments;
+  replicas?: [string, KVMap | undefined][];
 }
 
 export interface RunnableConfigLike {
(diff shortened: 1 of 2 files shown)


js/src/run_trees.ts
```diff
@@ -438,26 +523,47 @@ export class RunTree implements BaseRun {
...
+          error: runData.error,
+          parent_run_id: runData.parent_run_id,
+          session_name: runData.session_name,
+          reference_example_id: runData.reference_example_id,
+          end_time: runData.end_time,
+          dotted_order: runData.dotted_order,
+          trace_id: runData.trace_id,
+          events: runData.events,
+          tags: runData.tags,
+          extra: runData.extra,
```
lc-arjun, line 541: can you just spread `runData` here? or important to be explicit?

angus-langchain (approved):

js/src/run_trees.ts
jacoblee93: I think we could, just wanted to preserve roughly the same code path as before.

jacoblee93: Yeah let's leave it for now actually since `_remapForProject` technically returns a `RunCreate`

PR #1798, opened 2025-06-16
patch[python]: Only set replica updates on patch

angus-langchain:


python/langsmith/run_trees.py
```diff
@@ -463,11 +463,7 @@ def post(self, exclude_child_runs: bool = True) -> None:
         """Post the run tree to the API asynchronously."""
         if self.replicas:
             for project_name, updates in self.replicas:
-                # Avoid double posting for the current project
-                if project_name == self.session_name:
-                    run_dict = self._get_dicts_safe()
-                else:
-                    run_dict = self._remap_for_project(project_name, updates)
+                run_dict = self._remap_for_project(project_name)
```
jacoblee93, line 469: We could do is something like this in `_remap_for_project` to avoid unnecessary remapping:

https://github.com/langchain-ai/langsmith-sdk/pull/1796/files#diff-fa3bb02ee6fa794ab2494fe9c12ab64d8f1daac80bbffcf54ae88f50934ae20bR432

I think this is nice because if the user e.g. passes `run_id` manually for the default project it works as expected. But open to just removing that from JS instead if desired.

jacoblee93 (approved):

PR #1804, opened 2025-06-17
feat[python]: Trace to multiple projects with distributed tracing

angus-langchain:


python/langsmith/run_trees.py
```diff
@@ -703,8 +704,10 @@ def from_headers(
             init_args["extra"]["metadata"] = metadata
             tags = sorted(set(baggage.tags + init_args.get("tags", [])))
             init_args["tags"] = tags
-            if baggage.project_name:
```
angus-langchain, line 706: @hinthornw Was this meant to be nested within the `if baggage.metadata or baggage.tags:` conditional?

python/langsmith/run_trees.py
```diff
@@ -751,6 +756,7 @@ def from_header(cls, header_value: Optional[str]) -> _Baggage:
         metadata = {}
         tags = []
         project_name = None
+        replicas = []
```
hinthornw: can init to none right?

python/langsmith/run_trees.py
hinthornw: no it wasn't

PR #1814, opened 2025-06-25
feat(js): Add OTEL support

jacoblee93:
CC @dqbd @angus-langchain


js/package.json
```diff
@@ -172,7 +172,10 @@
     "zod": "^3.23.8"
   },
   "peerDependencies": {
-    "openai": "*"
+    "openai": "*",
+    "@opentelemetry/api": "^1.9.0",
```
angus-langchain: I've made these optional in python

js/src/client.ts
```diff
@@ -653,6 +669,17 @@ export class Client implements LangSmithTracingClientInterface {
     this.batchSizeBytesLimit = config.batchSizeBytesLimit;
     this.fetchOptions = config.fetchOptions || {};
     this.manualFlushMode = config.manualFlushMode ?? this.manualFlushMode;
+    if (getEnvironmentVariable("OTEL_ENABLED") === "true") {
+      const existingTracerProvider = otel_trace.getTracerProvider();
+      const langSmithTracerProvider = getOTLPTracerProvider();
+      // If user has set global tracer before, this fails and returns false
+      const globalSuccessfullyOverridden = otel_trace.setGlobalTracerProvider(langSmithTracerProvider);
```
angus-langchain: will this overwrite the existing tracer provider?

js/src/client.ts
```diff
@@ -1063,6 +1126,16 @@ export class Client implements LangSmithTracingClientInterface {
     await this.drainAutoBatchQueue(sizeLimitBytes);
   }
 
+  private _cloneCurrentOTELContext() {
```
angus-langchain, line 1121: looks good, I had to mess with traceable in order to get otel spans correctly nested in langsmith/langchain spans so I'd make sure context works both from otel to langsmith and vice versa

js/src/_internal/otel/client.ts
```diff
@@ -0,0 +1,85 @@
...
+    if (name && valueParts.length > 0) {
+      headers[name.trim()] = valueParts.join("=").trim();
+    }
+  });
+
+  return headers;
+}
+
+export function getOTLPTracerProvider(): OTELTracerProvider {
+  if (!HAS_OTEL) {
```
angus-langchain: looks like this currently just checks env vars but I saw you added dependancies to todo

js/package.json
jacoblee93: Yes will figure all that out later although there's often not a great way to make them completely optional in TS

js/src/_internal/otel/client.ts
jacoblee93: yeah cleaning this up

js/src/client.ts
jacoblee93: `If user has set global tracer before, this fails and returns false`

OTEL TS loves this pattern

js/src/utils/_uuid.ts
```diff
@@ -1,7 +1,11 @@
-import * as uuid from "uuid";
+// Relaxed UUID validation regex (allows any valid UUID format including nil UUIDs)
+const UUID_REGEX =
+  /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
 
 export function assertUuid(str: string, which?: string): string {
-  if (!uuid.validate(str)) {
+  // Use relaxed regex validation instead of strict uuid.validate()
+  // This allows edge cases like nil UUIDs or test UUIDs that might not pass strict validation
+  if (!UUID_REGEX.test(str)) {
```
angus-langchain, line 8: why did you need to change this?

js/src/experimental/otel/translator.ts
```diff
@@ -0,0 +1,595 @@
...
+  operation: T;
+  id: string;
+  trace_id: string;
+  run: T extends "post" ? RunCreate : RunUpdate;
+};
+
+export class LangSmithToOTELTranslator {
+  private tracer?: OTELTracer;
+
+  private spans: Map<string, OTELSpan> = new Map();
```
angus-langchain, line 28: This never gets cleaned up, not sure if blocking but should we add a timeout

js/src/experimental/otel/translator.ts
```diff
@@ -0,0 +1,595 @@
...
+
+  private createSpanForRun(
+    op: SerializedRunOperation,
+    runInfo: RunCreate,
+    otelContext?: OTELContext
+  ): OTELSpan | undefined {
+    if (!this.tracer) {
+      return;
+    }
+    const activeSpan = otelContext && getOTELTrace().getSpan(otelContext);
```
angus-langchain, line 63: like above, should we have a timeout to end the run and cleanup the map

js/src/traceable.ts
```diff
@@ -29,11 +29,51 @@ import {
...
+      const resolvedTracer = otel_trace.getTracer("langsmith", __version__);
+      return resolvedTracer.startActiveSpan(
+        runTree.name,
+        {
+          attributes: {
+            "langsmith.traceable": "true",
+          },
+        },
+        () => {
+          otel_trace.setSpanContext(otel_context.active(), spanContext);
```
angus-langchain, line 71: does this work on the current trace, seems like its set after startActiveSpan

js/src/traceable.ts
```diff
@@ -558,7 +598,7 @@ export function traceable<Func extends (...args: any[]) => any>(
           extractAttachmentsFn
         );
 
-        return [currentRunTree, [currentRunTree, ...restArgs] as Inputs];
+        return [currentRunTree, restArgs as Inputs];
```
angus-langchain: why did we need this change?

js/src/client.ts
```diff
@@ -5055,17 +5149,18 @@ export class Client implements LangSmithTracingClientInterface {
...
         "[WARNING]: When tracing in manual flush mode, you must call `await client.flush()` manually to submit trace batches."
       );
       return Promise.resolve();
     }
-    return Promise.all([
+    await Promise.all([
       ...this.autoBatchQueue.items.map(({ itemPromise }) => itemPromise),
       this.batchIngestCaller.queue.onIdle(),
     ]);
+    await getDefaultOTLPTracerComponents()?.spanProcessor?.forceFlush();
```
angus-langchain: this could block, do we need to flush always?

js/src/experimental/otel/exporter.ts
```diff
@@ -0,0 +1,85 @@
...
+import * as constants from "./constants.js";
+import { isTracingEnabled } from "../../env.js";
+
+export class LangSmithOTLPTraceExporter extends OTLPTraceExporter {
+  export(
+    spans: ReadableSpan[],
+    resultCallback: Parameters<OTLPTraceExporter["export"]>[1]
+  ): void {
+    if (!isTracingEnabled()) {
+      return;
```
angus-langchain: do we need to call resultCallback here?

angus-langchain (approved):

js/src/utils/_uuid.ts
jacoblee93: OTEL spans now generate 000-00-0-00 UUIDs and that fails this validate check. This is also only ever used clientside when doing things like fetching runs, and the backend seems to happily accept these "invalid" ones

js/src/experimental/otel/translator.ts
jacoblee93: It does: https://github.com/langchain-ai/langsmith-sdk/pull/1814/files#diff-86c0d5802cda40a97aff332d431e504310ff7e226862a9d11283a557572c533dR150

But we probably want to also clear it from the map if the initial `POST` has an `endTime`

js/src/experimental/otel/translator.ts
jacoblee93: See above

js/src/traceable.ts
jacoblee93: Docs are pretty unclear but it seems to work

js/src/traceable.ts
jacoblee93: Oh this looks incidental, fixing

js/src/client.ts
jacoblee93: Ah good catch thank you!

jacoblee93: How would it block improperly? We want to wait for this to flush don't we?

js/src/experimental/otel/exporter.ts
jacoblee93: Yes great call

js/src/experimental/otel/translator.ts
jacoblee93: Done

jacoblee93:
@dgieselaar neat, thanks! Will take a look.

LangSmith also has an endpoint that ingests OTEL already if interesting to you folks: https://docs.smith.langchain.com/observability/how_to_guides/trace_with_opentelemetry

Just FYI rate limits are a bit lower at the moment

dqbd (approved):

PR #1821, opened 2025-07-01
feat(js): Add transform span method to exporter, allow setting run name

jacoblee93:


js/src/experimental/otel/exporter.ts
```diff
@@ -24,6 +24,34 @@ function parseHeadersString(headersStr: string): Record<string, string> {
...
+   *     }
+   *     return span;
+   *   }
+   * });
+   * ```
+   *
+   * @param span - The span to transform.
+   * @returns A transformed version of the span.
+   */
+  transformExportedSpan?: (span: ReadableSpan) => ReadableSpan;
```
angus-langchain: or `Promise<ReadableSpan>` right?

js/src/experimental/otel/setup.ts
```diff
@@ -12,13 +12,36 @@ import {
...
 
 import {
   setDefaultOTLPTracerComponents,
   setOTELInstances,
 } from "../../singletons/otel.js";
 
+/**
+ * Configuration options for initializing OpenTelemetry with LangSmith.
+ */
+export type InitializeOTELConfig = {
```
angus-langchain, line 28: I guess this is breaking but fine since no one using yet?

js/src/experimental/otel/setup.ts
```diff
@@ -70,7 +89,9 @@ export const initializeOTEL = ({
     otel_context.setGlobalContextManager(contextManager);
   }
 
-  const DEFAULT_LANGSMITH_SPAN_EXPORTER = new LangSmithOTLPTraceExporter({});
+  const DEFAULT_LANGSMITH_SPAN_EXPORTER = new LangSmithOTLPTraceExporter(
+    exporterConfig
```
angus-langchain, line 93: can this be undefined?

js/src/experimental/otel/exporter.ts
```diff
@@ -76,84 +110,95 @@ export class LangSmithOTLPTraceExporter extends OTLPTraceExporter {
     if (!isTracingEnabled()) {
       return resultCallback({ code: 0 });
     }
-    for (const span of spans) {
-      if (!span.attributes[constants.GENAI_PROMPT]) {
-        if (span.attributes["ai.prompt"]) {
-          span.attributes[constants.GENAI_PROMPT] =
-            span.attributes["ai.prompt"];
+    const runExport = async () => {
+      for (let span of spans) {
```
angus-langchain, line 116: does this actually update the array elements or just copies of the array elements? especially if `transformExportedSpan` returns a new span?

js/src/experimental/otel/exporter.ts
```diff
@@ -76,84 +110,95 @@ export class LangSmithOTLPTraceExporter extends OTLPTraceExporter {
     if (!isTracingEnabled()) {
       return resultCallback({ code: 0 });
     }
-    for (const span of spans) {
-      if (!span.attributes[constants.GENAI_PROMPT]) {
-        if (span.attributes["ai.prompt"]) {
-          span.attributes[constants.GENAI_PROMPT] =
-            span.attributes["ai.prompt"];
+    const runExport = async () => {
```
angus-langchain, line 115: since this is async/ fire and forget, wont a shutdown after `export()` exit before the http call completes or the `resultCallback` gets called.

Can we make export() itself async?

js/src/experimental/otel/exporter.ts
jacoblee93: Yes whoops

js/src/experimental/otel/exporter.ts
jacoblee93: Think it should delay until after `resultCallback` is called with something or times out. That's generally what happens in JS if a function does async work but takes a callback

This is a subclass so can't just make it `async` since the parent class won't await it without a keyword

js/src/experimental/otel/exporter.ts
jacoblee93: Updates them in place

If we really wanted to be safe we could make a deep copy but I think probably overkill for now

js/src/experimental/otel/setup.ts
jacoblee93: Don't think it's breaking? Optional to provide a config arg at all and we're just adding props

js/src/experimental/otel/setup.ts
jacoblee93: Yes and that's okay

PR #1826, opened 2025-07-05
chore(js): Add rate limit test for new AI SDK exporter

jacoblee93:


js/src/experimental/otel/constants.ts
```diff
@@ -45,6 +45,7 @@ export const LANGSMITH_RUN_ID = "langsmith.span.id";
 export const LANGSMITH_TRACE_ID = "langsmith.trace.id";
 export const LANGSMITH_DOTTED_ORDER = "langsmith.span.dotted_order";
 export const LANGSMITH_PARENT_RUN_ID = "langsmith.span.parent_id";
+export const LANGSMITH_USAGE_METADATA = "langsmith.usage_metadata";
```
hinthornw, line 48: @angus-langchain seems we don' set this in the python side - any reason for that?

jacoblee93: Because it's brand new :)

js/src/utils/vercel.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+  }
+  return inputTokenDetails;
+}
+
+export function extractUsageMetadata(span: {
+  status?: { code: number };
+  attributes?: Record<string, unknown>;
+}) {
+  const isError = span.status?.code === 2;
```
hinthornw: Since we're doing the `foo as KVMap` when calling do we need to do `span?.status` ?

jacoblee93: Could probably type it more strongly but want to be defensive here in case `status` is missing for some reason on a span

js/src/utils/vercel.types.ts
```diff
@@ -15,6 +15,7 @@ interface BaseLLMSpanAttributes {
 
   "ai.usage.promptTokens": number;
   "ai.usage.completionTokens": number;
+  "ai.response.providerMetadata": string;
```
hinthornw: Just checking - this doesn't need `?:`?

js/src/utils/vercel.ts
hinthornw: I mean the opposite - are we defensive enough?

hinthornw (commented):
Also in `vercel.ts`: `experimental_providerMetadata` seems to be deprecated 
```
**
  @deprecated Use `providerOptions` instead.
   */
```

js/src/utils/vercel.ts
jacoblee93: I think so but can make the param optional

js/src/utils/vercel.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  if (typeof span.attributes["ai.usage.promptTokens"] === "number") {
+    usageMetadata.input_tokens = span.attributes["ai.usage.promptTokens"];
+  }
+
+  if (typeof span.attributes["ai.usage.completionTokens"] === "number") {
+    usageMetadata.output_tokens = span.attributes["ai.usage.completionTokens"];
+  }
+
+  if (typeof span.attributes["ai.response.providerMetadata"] === "string") {
```
hinthornw, line 66: This wouldn't ever be an object already would it?

js/src/utils/vercel.types.ts
jacoblee93: Yeah suppose it technically does

js/src/utils/vercel.ts
jacoblee93: OTEL spans can't be objects I think and if it is an object something weird has happened and we should drop it

jacoblee93:
> Also in vercel.ts: experimental_providerMetadata seems to be deprecated

The attribute is now on the span anyway so it shouldn't matter

js/src/experimental/otel/exporter.ts
```diff
@@ -181,6 +182,9 @@ export class LangSmithOTLPTraceExporter extends OTLPTraceExporter {
           )
         ) {
           span.attributes[constants.LANGSMITH_RUN_TYPE] = "llm";
+          const usageMetadata = extractUsageMetadata(span);
+          span.attributes[constants.LANGSMITH_USAGE_METADATA] =
```
hinthornw, line 186: Should we only run this line if usageMetadata is populated? or every time

jacoblee93: Would only happen if span attributes were missing which would be very weird, and even then passing an empty stringified object here shouldn't cause any issues

PR #1829, opened 2025-07-07
feat: [ts] LANGSMITH_RUNS_ENDPOINTS

hinthornw:
Supports writing runs to mulitiple endpoints.
Takes precedence over langsmith/langchain endpoint.
Raise error if langsmith/langchain_endpoint is also provided.
used in LGP self-hosted under certain scenarios

This has been a feat in python for a while

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -84,10 +84,15 @@ import {
 
 import { serialize as serializePayloadForTracing } from "./utils/fast-safe-stringify/index.js";
 
+export interface Destination {
+  url: string;
+  apiKey?: string;
+}
+
 export interface ClientConfig {
   apiUrl?: string;
   apiKey?: string;
-  apiUrls?: Record<string, string>;
+  destinations?: Destination[];
   callerOptions?: AsyncCallerParams;
   timeout_ms?: number;
   webUrl?: string;
@@ -591,7 +596,7 @@ export class Client implements LangSmithTracingClientInterface {
 
   private batchIngestCaller: AsyncCaller;
 
-  private _writeApiUrls: Record<string, string | undefined> = {};
+  private _destinations: Record<string, string | undefined> = {};
 
   private timeout_ms: number;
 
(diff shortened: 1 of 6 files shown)


js/src/client.ts
```diff
@@ -83,6 +87,7 @@ import { serialize as serializePayloadForTracing } from "./utils/fast-safe-strin
 export interface ClientConfig {
   apiUrl?: string;
   apiKey?: string;
+  apiUrls?: Record<string, string>;
```
jacoblee93: Would really expect this to be an array - maybe some better name or has this already been merged into Python?

I'd suggest `destinations` and making this an array of objects even - url x API key is a little weird?

hinthornw: This has been in python for a year. The env var needs to be compatible. The client config could be different if we think breaking alignment is worth the ux improvement

js/src/client.ts
```diff
@@ -1032,6 +1051,85 @@ export class Client implements LangSmithTracingClientInterface {
...
+      );
+    }
+    const envVar = getEnvironmentVariable("LANGSMITH_RUNS_ENDPOINTS");
+    if (!envVar) return {};
+    try {
+      const parsed: Record<string, string> = JSON.parse(envVar);
+      console.warn(
+        `Parsed ${parsed} | LS: ${getEnvironmentVariable(
+          "LANGSMITH_ENDPOINT"
+        )} | LC: ${getEnvironmentVariable("LANGCHAIN_ENDPOINT")}`
```
jacoblee93: Why is this a warning?

js/src/client.ts
hinthornw: but ya [{url: foo, apiKey: bar}, {url: foo, apiKey: bar}] is more extensible

js/src/client.ts
```diff
@@ -1032,6 +1051,85 @@ export class Client implements LangSmithTracingClientInterface {
...
+
+  private async _callWithWriteApiUrls(
+    caller: AsyncCaller,
+    path: string,
+    init: RequestInit & { duplex?: string },
+    errorContext: string,
+    ignoreConflict = false
+  ) {
+    const errors: unknown[] = [];
+    for (const [apiUrl, apiKey] of Object.entries(this._writeApiUrls)) {
```
jacoblee93: Probably better to do this in parallel

js/src/client.ts
hinthornw: lol i'm sloppy need to del

jacoblee93: Lol!

js/src/client.ts
```diff
@@ -1032,6 +1051,77 @@ export class Client implements LangSmithTracingClientInterface {
...
+      }
+      console.warn(
+        "Invalid LANGSMITH_RUNS_ENDPOINTS – must be valid JSON mapping of url->apiKey"
+      );
+      return [];
+    }
+  }
+
+  private async _callWithWriteApiUrls(
+    caller: AsyncCaller,
```
jacoblee93, line 1088: Why does this need to be passed?

PR #1830, opened 2025-07-08
Add runs endpoints env to replicas

hinthornw:
Alternative to https://github.com/langchain-ai/langsmith-sdk/pull/1829

### DIFF
--- a/js/src/tests/replica_endpoints.test.ts
+++ b/js/src/tests/replica_endpoints.test.ts
@@ -2,6 +2,7 @@
 /* eslint-disable no-process-env */
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 /* eslint-disable prefer-const */
+/* eslint-disable no-instanceof/no-instanceof */
 import { jest } from "@jest/globals";
 import { Client } from "../client.js";
 import { RunTree } from "../run_trees.js";


js/src/client.ts
```diff
@@ -331,6 +331,8 @@ interface CreateRunParams {
   trace_id?: string;
   dotted_order?: string;
   attachments?: Attachments;
+  apiKey?: string;
```
jacoblee93: I think this is user facing right? Do we want that?

js/src/run_trees.ts
```diff
@@ -489,12 +516,23 @@ export class RunTree implements BaseRun {
...
+              replica.projectName,
+              runtimeEnv,
+              true
+            );
+          }
+          if (replica.apiKey) {
+            runCreate.apiKey = replica.apiKey;
+          }
+          if (replica.apiUrl) {
+            runCreate.apiUrl = replica.apiUrl;
```
jacoblee93: Would prefer to pass these separately into `createRun` - or hot take maybe even give each replica its own client?

js/src/client.ts
hinthornw: Ya it is. I could hide it and say eff typescript?

js/src/run_trees.ts
hinthornw: hmmm would that be worse

jacoblee93: Yeah probably too hot of a take

jacoblee93 (approved):

PR #1843, opened 2025-07-10
feat[python]: Add distributed tracing with sliced traces for child projects

angus-langchain:
### Description
Adds support for distributed tracing where parent projects receive the full trace while child projects receive a sliced trace starting from the distributed parent ID.
Screenshot 2025-07-10 at 9 34 36 AM

## Limitations / TODOs
1. Tracing only works within 1 Workspace. Cross-workspace distributed tracing is not supported yet.
2. Tracing will only work for a parent-child relationship. It will not work for chains of 3 or more distributed graphs (e.g., A→B→C). Only the immediate parent-child pair will have proper trace slicing.
4. Requires manual configuration: The `distributed: True` flag must be explicitly set in replica configuration for each child project that should receive sliced traces.

### DIFF
--- a/python/langsmith/run_trees.py
+++ b/python/langsmith/run_trees.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import functools
 import json
 import logging
 import sys
@@ -9,6 +10,8 @@ from collections.abc import Mapping, Sequence
 from datetime import datetime, timezone
 from typing import Any, Optional, Union, cast
 from uuid import NAMESPACE_DNS, UUID, uuid4, uuid5
+
+from typing_extensions import TypedDict
 
 try:
     from pydantic.v1 import Field, root_validator  # type: ignore[import]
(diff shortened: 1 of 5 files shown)


python/langsmith/run_trees.py
```diff
@@ -415,6 +418,37 @@ def _get_dicts_safe(self):
...
         return self_dict
 
+    def _slice_parent_id(self, parent_id: str, run_dict: dict) -> None:
+        """Slice the parent id from dotted order.
+
+        Additionally check if the current run is a child of the parent. If so, update
+        the parent_run_id to None, and set the trace id to the new root id after
+        parent_id.
+        """
+        if run_dict.get("dotted_order"):
```
hinthornw:
```suggestion
        if dotted_order := run_dict.get("dotted_order"):
            segs = _parse_dotted_order(dotted_order)
```

python/langsmith/run_trees.py
```diff
@@ -415,6 +418,37 @@ def _get_dicts_safe(self):
             self_dict["outputs"] = self.outputs.copy()
         return self_dict
 
+    def _slice_parent_id(self, parent_id: str, run_dict: dict) -> None:
+        """Slice the parent id from dotted order.
+
+        Additionally check if the current run is a child of the parent. If so, update
+        the parent_run_id to None, and set the trace id to the new root id after
+        parent_id.
+        """
```
hinthornw, line 529: maybe :

parent_id = str(parent_id) atthe beginning since we keep re-casting afterwards

python/langsmith/run_trees.py
```diff
@@ -415,6 +418,37 @@ def _get_dicts_safe(self):
...
+        if run_dict.get("dotted_order"):
+            segs = _parse_dotted_order(run_dict["dotted_order"])
+            # Find the index where seg_id == parent_id
```
hinthornw:
```suggestion
```

python/langsmith/run_trees.py
```diff
@@ -415,6 +418,37 @@ def _get_dicts_safe(self):
...
+            start_idx = None
+            for idx, (_, seg_id) in enumerate(segs):
```
hinthornw, line 535: (off topic) we could probably bin search this since they're all sortable; don't think it would actually be faster at typical tree depth though

angus-langchain: can add a todo but yeah might be an over optimization atp

PR #1845, opened 2025-07-11
If stream fails, fallback to buffered body request

angus-langchain:
### Description

In the case that a stream request cannot successfully be sent to LangSmith multipart tracing, fallback to tracing via buffered body.  This is a less performant method of tracing, but adding a fallback makes requests more resilient.

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -1607,22 +1607,27 @@ export class Client implements LangSmithTracingClientInterface {
     const buildBuffered = () => this._createNodeFetchBody(parts, boundary);
     const buildStream = () => this._createMultipartStream(parts, boundary);
 
-    const send = async (body: BodyInit) =>
-      await this.batchIngestCaller.call(
+    const send = async (body: BodyInit) => {
+      const headers: Record<string, string> = {
+        ...this.headers,
+        "Content-Type": `multipart/form-data; boundary=${boundary}`,
+      };
+      if (options?.apiKey !== undefined) {
+        headers["x-api-key"] = options.apiKey;
+      }
+      return await this.batchIngestCaller.call(
         _getFetchImplementation(this.debug),
         `${options?.apiUrl ?? this.apiUrl}/runs/multipart`,
         {
           method: "POST",
-          headers: {
-            ...this.headers,
-            "Content-Type": `multipart/form-data; boundary=${boundary}`,
-          },
+          headers,
           body,
           duplex: "half",
           signal: AbortSignal.timeout(this.timeout_ms),
           ...this.fetchOptions,
         }
       );
+    };
 
     try {
       let streamedAttempt = false;
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -1617,7 +1617,7 @@ export class Client implements LangSmithTracingClientInterface {
       if (options?.apiKey !== undefined) {
         headers["x-api-key"] = options.apiKey;
       }
-      return await this.batchIngestCaller.call(
+      return this.batchIngestCaller.call(
         _getFetchImplementation(this.debug),
         `${options?.apiUrl ?? this.apiUrl}/runs/multipart`,
         {
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -634,6 +634,8 @@ export class Client implements LangSmithTracingClientInterface {
 
   private langSmithToOTELTranslator?: LangSmithToOTELTranslator;
 
+  private streamingEnabled = true;
+
   debug = getEnvironmentVariable("LANGSMITH_DEBUG") === "true";
 
   constructor(config: ClientConfig = {}) {
@@ -1626,8 +1628,8 @@ export class Client implements LangSmithTracingClientInterface {
       let streamedAttempt = false;
       let res: Response;
 
-      // attempt stream
-      if (!isNodeFetch) {
+      // attempt stream only if not disabled and not using node-fetch
+      if (!isNodeFetch && this.streamingEnabled) {
         streamedAttempt = true;
         res = await send(await buildStream());
       } else {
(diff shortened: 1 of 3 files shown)


js/src/client.ts
```diff
@@ -1597,34 +1597,55 @@ export class Client implements LangSmithTracingClientInterface {
...
+          },
           body,
           duplex: "half",
           signal: AbortSignal.timeout(this.timeout_ms),
           ...this.fetchOptions,
         }
       );
+
+    try {
+      let streamedAttempt = false;
```
jacoblee93: I think we should set this for the client so that it never tries again, since I don't really think this is retryable

js/src/client.ts
```diff
@@ -1745,9 +1766,8 @@ export class Client implements LangSmithTracingClientInterface {
         sessionId = project.id;
       }
       const tenantId = await this._getTenantId();
-      return `${this.getHostUrl()}/o/${tenantId}/projects/p/${sessionId}/r/${
-        run.id
-      }?poll=true`;
+      return `${this.getHostUrl()}/o/${tenantId}/projects/p/${sessionId}/r/${run.id
```
jacoblee93: incidental?

js/src/client.ts
```diff
@@ -3245,18 +3261,18 @@ export class Client implements LangSmithTracingClientInterface {
...
-          splits?: Array<string | Array<string>>;
-          sourceRunIds?: Array<string>;
-          useSourceRunIOs?: Array<boolean>;
-          useSourceRunAttachments?: Array<string[]>;
-          attachments?: Array<Attachments>;
-          exampleIds?: Array<string>;
-          datasetId?: string;
-          datasetName?: string;
-        }
+        inputs?: Array<KVMap>;
```
jacoblee93: Why are there so many formatting changes?

js/src/client.ts
```diff
@@ -634,6 +634,8 @@ export class Client implements LangSmithTracingClientInterface {
 
   private langSmithToOTELTranslator?: LangSmithToOTELTranslator;
 
+  private streamingEnabled = true;
```
jacoblee93: nit: prefer falsey defaults, so `streamingDisabled`

js/src/client.ts
```diff
@@ -1597,34 +1599,55 @@ export class Client implements LangSmithTracingClientInterface {
...
-        "----LangSmithFormBoundary" + Math.random().toString(36).slice(2);
-      const body = await (_globalFetchImplementationIsNodeFetch()
-        ? this._createNodeFetchBody(parts, boundary)
-        : this._createMultipartStream(parts, boundary));
-
-      const headers: Record<string, string> = {
-        ...this.headers,
-        "Content-Type": `multipart/form-data; boundary=${boundary}`,
-      };
-      if (options?.apiKey !== undefined) {
```
jacoblee93, line 1612: Let's keep this case please

js/src/client.ts
jacoblee93: Also better to call it something more specific like `multipartStreamingDisabled`

js/src/client.ts
```diff
@@ -1597,34 +1599,55 @@ export class Client implements LangSmithTracingClientInterface {
...
+      // attempt stream only if not disabled and not using node-fetch
+      if (!isNodeFetch && this.streamingEnabled) {
+        streamedAttempt = true;
+        res = await send(await buildStream());
+      } else {
+        res = await send(await buildBuffered());
+      }
+
+      // if stream fails, fallback to buffered body
+      if (streamedAttempt && res.status >= 400 && res.status < 500) {
```
jacoblee93: Wonder if worth picking out the specific error?

js/src/client.ts
```diff
@@ -1597,34 +1599,55 @@ export class Client implements LangSmithTracingClientInterface {
...
+        console.warn(
+          `Streaming multipart request failed with EOF error, falling back to buffered mode. Context: ${context}`
```
jacoblee93: Since this would trigger on like 401s etc.

js/src/client.ts
```diff
@@ -1597,23 +1599,23 @@ export class Client implements LangSmithTracingClientInterface {
...
       const headers: Record<string, string> = {
         ...this.headers,
         "Content-Type": `multipart/form-data; boundary=${boundary}`,
       };
       if (options?.apiKey !== undefined) {
         headers["x-api-key"] = options.apiKey;
       }
-
-      const res = await this.batchIngestCaller.call(
+      return await this.batchIngestCaller.call(
```
jacoblee93: nit for the future - `return await` is unnecessary here:

https://stackoverflow.com/questions/38708550/difference-between-return-await-promise-and-return-promise

js/src/client.ts
```diff
@@ -1625,6 +1627,33 @@ export class Client implements LangSmithTracingClientInterface {
...
+        res = await send(await buildBuffered());
+      }
+
+      // if stream fails, fallback to buffered body
+      if (
+        !this.multipartStreamingDisabled &&
+        res.status === 422
+      ) {
+        console.warn(
+          `Streaming multipart request failed with EOF error, falling back to buffered mode. Context: ${context}`
```
jacoblee93: Actually, can we change this to only log/trigger if the default URL is not our backend?

And maybe log something about to check that your custom host supporting streaming multipart?

js/src/client.ts
```diff
@@ -1625,6 +1627,33 @@ export class Client implements LangSmithTracingClientInterface {
           ...this.fetchOptions,
         }
       );
+    };
+
+    try {
+      let res: Response;
+
+      // attempt stream only if not disabled and not using node-fetch
+      if (!isNodeFetch && !this.multipartStreamingDisabled) {
```
jacoblee93, line 1639: As discussed offline, to be completely safe and avoid race conditions let's keep the local state of whether the request was already tried with streaming

jacoblee93 (approved):
I was going to say let's log only if they're not on default host rather than avoid the retry entirely but this should be fine

PR #1847, opened 2025-07-12
feat(js): Support feedback attribution for OTEL evaluations

jacoblee93:


js/src/evaluation/evaluator.ts
```diff
@@ -231,7 +231,11 @@ export class DynamicRunEvaluator<Func extends (...args: any[]) => any>
       {
         project_name: "evaluators",
         name: "evaluator",
-        id: sourceRunId,
+        on_end: (runTree) => {
```
dqbd, line 234: _nit_: I wish we used pascalCase here for consistency?

jacoblee93: I think it comes all the way down from run tree

Would be great to figure it out yeah

PR #1849, opened 2025-07-13
fix(js): Propagate microseconds added in dotted order conversion to run tree start time

jacoblee93:
CC @agola11 @dqbd

### DIFF
--- a/js/src/schemas.ts
+++ b/js/src/schemas.ts
@@ -91,13 +91,13 @@ export interface BaseRun {
   name: string;
 
   /** The epoch time at which the run started, if available. */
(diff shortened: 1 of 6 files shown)


js/src/schemas.ts
```diff
@@ -91,13 +91,13 @@ export interface BaseRun {
   name: string;
 
   /** The epoch time at which the run started, if available. */
-  start_time?: number;
+  start_time?: number | string;
```
jacoblee93, line 94: This was actually just straight up wrong, LangSmith returns this as a date string

PR #1850, opened 2025-07-14
fix(js): Respect project names set at runtime for OTEL exported traces

jacoblee93:
Rely on span attribute rather than sending it in the header

Very frustrating that headers are not configurable at runtime in the exporter superclass :(

### DIFF
--- a/js/src/experimental/otel/exporter.ts
+++ b/js/src/experimental/otel/exporter.ts
@@ -63,6 +63,11 @@ export type LangSmithOTLPTraceExporterConfig = ConstructorParameters<
    * The name of the project to export traces to.
    */
   projectName?: string;
+
+  /**
+   * Default headers to add to exporter requests.
+   */
+  headers?: Record<string, string>;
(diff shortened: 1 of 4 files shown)


js/src/experimental/otel/exporter.ts
```diff
@@ -93,29 +100,29 @@ export class LangSmithOTLPTraceExporter extends OTLPTraceExporter {
...
-        config?.apiKey ?? getLangSmithEnvironmentVariable("API_KEY");
-      if (apiKey) {
-        defaultHeaderString = `x-api-key=${apiKey}`;
-      }
-
-      const project =
-        config?.projectName ?? getLangSmithEnvironmentVariable("PROJECT");
-      if (project) {
-        defaultHeaderString += `,Langsmith-Project=${project}`;
+    let headers = config?.headers;
```
jacoblee93, line 103: Incidental here, just allowing users to pass this in separately if needed

PR #1854, opened 2025-07-16
fix(js): Fallback to parent run id if parent run is not present in run trees loaded from LangChain

jacoblee93:
This is difficult to test in langsmith due to circular deps but verified independently and will add a test to LangChain.

### DIFF
--- a/js/src/run_trees.ts
+++ b/js/src/run_trees.ts
@@ -45,8 +45,8 @@ export interface RunTreeConfig {
   parent_run?: RunTree;
   parent_run_id?: string;
   child_runs?: RunTree[];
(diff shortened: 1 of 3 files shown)


js/src/run_trees.ts
```diff
@@ -45,8 +45,8 @@ export interface RunTreeConfig {
   parent_run?: RunTree;
   parent_run_id?: string;
   child_runs?: RunTree[];
-  start_time?: number;
-  end_time?: number;
+  start_time?: number | string;
+  end_time?: number | string;
```
jacoblee93, line 49: Incidental, just cleaning up types

PR #1859, opened 2025-07-18
feat: Configure

hinthornw:


python/langsmith/run_trees.py
```diff
@@ -58,6 +59,35 @@ def get_cached_client(**init_kwargs: Any) -> Client:
     return _CLIENT
 
 
+def configure(
```
angus-langchain, line 62: can you add a docstring

hinthornw: 🫡

PR #1861, opened 2025-07-20
fix(python): Avoid resetting operation.name to "chain" during update_run

qnnn:
Prevents gen_ai.operation.name from being unintentionally reset to "chain" during spec conversion.
The issue stems from a missing run_type in update_run.

### DIFF
--- a/python/tests/integration_tests/test_client.py
+++ b/python/tests/integration_tests/test_client.py
@@ -2,9 +2,11 @@
 
 import asyncio
 import datetime
+import importlib
 import io
 import logging
 import os
+import queue
 import random
 import string
 import sys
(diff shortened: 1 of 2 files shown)


python/tests/integration_tests/test_client.py
```diff
@@ -3299,3 +3311,97 @@ def child():
     finally:
         if langchain_client.has_project(project_name=project_name):
             langchain_client.delete_project(project_name=project_name)
+
+
+def test_otel_trace_attributes(monkeypatch: pytest.MonkeyPatch):
+    monkeypatch.setenv("LANGSMITH_OTEL_ENABLED", "true")
+    get_env_var.cache_clear()
```
qnnn, line 3454: I'm not sure whether this unit test should be skipped due to caching.

qnnn:
Thanks! I’ve rebased onto the latest main and resubmitted.

angus-langchain (approved):

PR #1864, opened 2025-07-21
docs: add missing end_time param to list_runs

ads2280:


python/langsmith/client.py
```diff
@@ -2428,6 +2429,8 @@ def list_runs(
                 The ID of the parent run to filter by.
             start_time (Optional[datetime.datetime]):
                 The start time to filter by.
+            end_time (Optional[datetime.datetime]):
+                The end time to filter by.
```
hinthornw: could we be more explicit (here and in start_time) that it only picks runs whose end_time is earlier than the provided time. And for start_time it only picks runs that start after the provided start time (i wrote bad docs )

hinthornw (approved):

PR #1867, opened 2025-07-23
fix (js): include env var metadata in patch

hinthornw:


js/src/utils/env.ts
```diff
@@ -154,6 +158,8 @@ export function getLangChainEnvVarsMetadata(): Record<string, string> {
     }
   }
 
+  _cachedEnvVarsMetadata = envVars;
```
jacoblee93: Why do this?

jacoblee93 (approved):

js/src/utils/env.ts
hinthornw: ha good callout

PR #1877, opened 2025-07-29
feat(python): Add replica endpoints support to Python SDK

angus-langchain:
### Description
Backport the replica endpoints functionality from JavaScript to Python, enabling runs to be sent to multiple LangSmith instances with different API keys and URLs via replicas, instead of a separate slot.

### Changes
1. New replica format: Added WriteReplica type supporting api_url, api_key,  project_name, and updates
2. Environment variable support: LANGSMITH_RUNS_ENDPOINTS now parsed into replicas (replaces old _write_api_urls logic)
3. Tracing context: Baggage headers now support both legacy and new replica formats for distributed tracing
4. Client methods: create_run() and update_run() accept optional api_key/api_url parameters
5. Backward compatibility: Legacy tuple format (project_name, updates) still supported

### Breaking Changes
1. LANGSMITH_RUNS_ENDPOINTS no longer populates _write_api_urls (now handled via replicas)

### DIFF
--- a/python/langsmith/run_helpers.py
+++ b/python/langsmith/run_helpers.py
@@ -44,6 +44,7 @@ from langsmith import client as ls_client
 from langsmith import run_trees, schemas, utils
 from langsmith._internal import _aiter as aitertools
 from langsmith.env import _runtime_env
+from langsmith.run_trees import Replica
 
 if TYPE_CHECKING:
     from types import TracebackType
@@ -108,7 +109,7 @@ def tracing_context(
     parent: Optional[Union[run_trees.RunTree, Mapping, str, Literal[False]]] = None,
     enabled: Optional[Union[bool, Literal["local"]]] = None,
     client: Optional[ls_client.Client] = None,
-    replicas: Optional[Sequence[tuple[str, Optional[dict]]]] = None,
+    replicas: Optional[Sequence[Replica]] = None,
     **kwargs: Any,
 ) -> Generator[None, None, None]:
     """Set the tracing context for a block of code.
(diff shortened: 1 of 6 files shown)


python/langsmith/utils.py
```diff
@@ -85,6 +85,30 @@ def __init__(
         self.exceptions = exceptions
 
 
+ERR_CONFLICTING_ENDPOINTS = "ERR_CONFLICTING_ENDPOINTS"
+
+
+class LangSmithConflictingEndpointsError(LangSmithError):
```
jacoblee93: This feels overly specific - what's the use case for someone catching this specifically?

I would opt for less complexity and a more general error since nobody should ever encounter this more than once or in prod

python/langsmith/run_helpers.py
```diff
@@ -123,8 +124,11 @@ def tracing_context(
         client: The client to use for logging the run to LangSmith. Defaults to None,
         enabled: Whether tracing is enabled. Defaults to None, meaning it will use the
             current context value or environment variables.
-        replicas: A sequence of tuples containing project names and optional updates for each replica.
-            Example: [("my_experiment", {"reference_example_id": None}), ("my_project", None)]
+        replicas: A sequence of replicas to send runs to. Can be either:
+            - Legacy format: tuples of (project_name, updates)
```
jacoblee93: Meta note: every single time I use a tuple I feel like I end up really regretting it 😂

jacoblee93: Though I don't think we formally deprecated the shorthand in JS

jacoblee93 (commented):
Small comment

python/langsmith/run_trees.py
```diff
@@ -897,6 +937,70 @@ def to_header(self) -> str:
         return ",".join(items)
 
 
+def _get_write_replicas_from_env() -> list[WriteReplica]:
```
hinthornw: We should cache this to avoid json.loads'ing stuff a lot

python/langsmith/run_trees.py
```diff
@@ -855,7 +880,22 @@ def from_header(cls, header_value: Optional[str]) -> _Baggage:
                     project_name = urllib.parse.unquote(value)
                 elif key == LANGSMITH_REPLICAS:
                     replicas_data = json.loads(urllib.parse.unquote(value))
-                    replicas = [(str(proj), updates) for proj, updates in replicas_data]
+                    parsed_replicas: list[Replica] = []
+                    for replica_item in replicas_data:
+                        if isinstance(replica_item, list) and len(replica_item) == 2:
+                            # Legacy format: [project_name, updates]
```
hinthornw: Should we still load as a dict ? Feels nice to have consistent format once it's actually loaded

hinthornw (approved):

PR #1878, opened 2025-07-30
feat(py): Adds new cached_hosts parameter to vcrpy caching

jacoblee93:


python/langsmith/utils.py
```diff
@@ -509,8 +511,33 @@ def with_cache(
     patch_urllib3.patch_urllib3()
 
     def _filter_request_headers(request: Any) -> Any:
+        # Legacy behavior
         if ignore_hosts and any(request.url.startswith(host) for host in ignore_hosts):
             return None
+
+        if allow_hosts:
+            from urllib.parse import urlparse
```
baskaryan: import up top

python/langsmith/utils.py
```diff
@@ -509,8 +511,33 @@ def with_cache(
...
+
+            try:
+                parsed_url = urlparse(request.url)
```
baskaryan: try except can be just around this line

python/langsmith/utils.py
```diff
@@ -509,8 +511,33 @@ def with_cache(
     patch_urllib3.patch_urllib3()
 
     def _filter_request_headers(request: Any) -> Any:
```
baskaryan: can we unnest and test this function

baskaryan (commented):
overall makes sense to me!

python/langsmith/utils.py
jacoblee93: Ya good call - was struggling with properly mocking vcrpy in a test and gave up

python/langsmith/testing/_internal.py
```diff
@@ -947,10 +986,15 @@ def _test():
             "reference_example_id": str(test_case.example_id),
         },
     }
+    # Handle cached_hosts parameter
+    ignore_hosts = [test_case.test_suite.client.api_url]
+    cached_hosts = langtest_extra.get("cached_hosts")
+    allow_hosts = cached_hosts if cached_hosts else None
```
baskaryan:
```suggestion
    allow_hosts = langtest_extra.get("cached_hosts") or None
```

PR #1881, opened 2025-07-31
patch(python, js): Fix sampling logic for patch runs

angus-langchain:
### Description
We were not correctly filtering patch runs when the post would get filtered, meaning large amounts of patch requests were being sent to ingest API and counting toward usage limits, but not actually getting ingested.

### Changes
Check patch trace id against trace id list, only remove trace ids from list for root runs.

### Alternative Solutions
Use consistent hashing based on run_id to sample both post and patch run ids equally, prevents storage of uuids (minimal memory overhead)

## Tests
Added unit tests confirming filtering logic correctly applies to both post and patch

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -937,10 +937,8 @@ export class Client implements LangSmithTracingClientInterface {
       for (const run of runs) {
         if (!this.filteredPostUuids.has(run.trace_id)) {
           sampled.push(run);
-        } else {
-          if (run.id === run.trace_id) {
-            this.filteredPostUuids.delete(run.trace_id);
-          }
+        } else if (run.id === run.trace_id) {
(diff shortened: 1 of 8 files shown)


jacoblee93 (approved):

js/src/client.ts
```diff
@@ -935,10 +935,12 @@ export class Client implements LangSmithTracingClientInterface {
     if (patch) {
       const sampled = [];
       for (const run of runs) {
-        if (!this.filteredPostUuids.has(run.id)) {
+        if (!this.filteredPostUuids.has(run.trace_id)) {
           sampled.push(run);
         } else {
```
jacoblee93: Just `else if`

hinthornw (approved):

PR #1884, opened 2025-07-31
feat(python): Add distributed tracing with sliced traces for child projects

angus-langchain:
### Description
Adds support for distributed tracing where parent projects receive the full trace while child projects receive a sliced trace starting from the distributed parent ID.
Screenshot 2025-07-10 at 9 34 36 AM

## Usage
Requires manual configuration: The `reroot: True` flag must be explicitly set in replica configuration for each child project that should receive sliced traces.

## Features
1. Cross workspace support

## Limitations
2. Tracing will only work for a parent-child relationship. It will not work for chains of 3 or more distributed graphs (e.g., A→B→C). Only the immediate parent-child pair will have proper trace slicing.

### DIFF
--- a/python/langsmith/run_trees.py
+++ b/python/langsmith/run_trees.py
@@ -62,9 +62,10 @@ _DISTRIBUTED_PARENT_ID = contextvars.ContextVar[Optional[str]](
 
 _SENTINEL = cast(None, object())
 
+TIMESTAMP_LENGTH = 36
+
+
 # Note, this is called directly by langchain. Do not remove.
-
-
 def get_cached_client(**init_kwargs: Any) -> Client:
     global _CLIENT
     if _CLIENT is None:
@@ -533,7 +534,7 @@ class RunTree(ls_schemas.RunBase):
             parent_id = str(parent_id)
             # TODO(angus): potentially use binary search to find the index
             for idx, part in enumerate(segs):
-                seg_id = part[-36:]
+                seg_id = part[-TIMESTAMP_LENGTH:]
                 if str(seg_id) == parent_id:
                     start_idx = idx
                     break
@@ -543,7 +544,7 @@ class RunTree(ls_schemas.RunBase):
                 # Rebuild dotted_order
                 run_dict["dotted_order"] = ".".join(trimmed_segs)
                 if trimmed_segs:
-                    run_dict["trace_id"] = UUID(trimmed_segs[0][-36:])
+                    run_dict["trace_id"] = UUID(trimmed_segs[0][-TIMESTAMP_LENGTH:])
                 else:
                     run_dict["trace_id"] = run_dict["id"]
         if str(run_dict.get("parent_run_id")) == parent_id:
@@ -583,9 +584,11 @@ class RunTree(ls_schemas.RunBase):
             segs = run_dict["dotted_order"].split(".")
             rebuilt = []
             for part in segs[:-1]:
-                repl = uuid5(NAMESPACE_DNS, f"{part[-36:]}:{project_name}")
-                rebuilt.append(part[:-36] + str(repl))
-            rebuilt.append(segs[-1][:-36] + str(new_id))
+                repl = uuid5(
+                    NAMESPACE_DNS, f"{part[-TIMESTAMP_LENGTH:]}:{project_name}"
+                )
+                rebuilt.append(part[:-TIMESTAMP_LENGTH] + str(repl))
+            rebuilt.append(segs[-1][:-TIMESTAMP_LENGTH] + str(new_id))
             dotted = ".".join(rebuilt)
         else:
             dotted = None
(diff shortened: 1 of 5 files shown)


python/langsmith/run_trees.py
```diff
@@ -516,6 +520,38 @@ def _get_dicts_safe(self):
...
+
+        Additionally check if the current run is a child of the parent. If so, update
+        the parent_run_id to None, and set the trace id to the new root id after
+        parent_id.
+        """
+        if dotted_order := run_dict.get("dotted_order"):
+            segs = _parse_dotted_order(dotted_order)
+            start_idx = None
+            parent_id = str(parent_id)
+            # TODO(angus): potentially use binary search to find the index
```
jacoblee93, line 535: Feels like `segs` will be max like 100 levels deep? Prob not worth thinking about

python/langsmith/run_trees.py
```diff
@@ -516,6 +520,38 @@ def _get_dicts_safe(self):
...
+                run_dict["dotted_order"] = ".".join(
+                    ts.strftime("%Y%m%dT%H%M%S%fZ") + str(seg_id)
+                    for ts, seg_id in trimmed_segs
+                )
+                if trimmed_segs:
+                    run_dict["trace_id"] = trimmed_segs[0][1]
+                else:
+                    run_dict["trace_id"] = run_dict["id"]
+        # Remove parent_run_id if it matches parent_id
+        if str(run_dict.get("parent_run_id")) == parent_id:
```
jacoblee93, line 550: When would this not be the case?

python/langsmith/run_trees.py
```diff
@@ -516,6 +520,38 @@ def _get_dicts_safe(self):
...
+                if str(seg_id) == parent_id:
+                    start_idx = idx
+                    break
+            if start_idx is not None:
+                # Trim segments to start after parent_id (exclusive)
+                trimmed_segs = segs[start_idx + 1 :]
+                # Rebuild dotted_order
+                run_dict["dotted_order"] = ".".join(
+                    ts.strftime("%Y%m%dT%H%M%S%fZ") + str(seg_id)
+                    for ts, seg_id in trimmed_segs
```
jacoblee93: Shouldn't the timestamps already be properly formatted? Or maybe we should just do `split(".")` instead of `parse_dotted_order`?

jacoblee93 (commented):
Seems fine, do we need to worry about patch at all?

python/langsmith/run_trees.py
angus-langchain: parent_id is the distributed parent id, so its looking for the new root run

angus-langchain: will add comment

python/langsmith/run_trees.py
angus-langchain: good point

python/langsmith/run_trees.py
```diff
@@ -516,6 +520,36 @@ def _get_dicts_safe(self):
...
+        the parent_run_id to None, and set the trace id to the new root id after
+        parent_id.
+        """
+        if dotted_order := run_dict.get("dotted_order"):
+            segs = dotted_order.split(".")
+            start_idx = None
+            parent_id = str(parent_id)
+            # TODO(angus): potentially use binary search to find the index
+            for idx, part in enumerate(segs):
+                seg_id = part[-36:]
```
jacoblee93: Just split on `Z`?

python/langsmith/run_trees.py
```diff
@@ -524,6 +558,12 @@ def _remap_for_project(
         if project_name == self.session_name:
             return run_dict
 
+        if updates and updates.get("reroot", False):
+            distributed_parent_id = _DISTRIBUTED_PARENT_ID.get()
```
jacoblee93, line 563: I'm actually not really following how this gets set - from a header?

python/langsmith/run_trees.py
```diff
@@ -540,13 +580,12 @@ def _remap_for_project(
...
-            segs = _parse_dotted_order(run_dict["dotted_order"])
+            segs = run_dict["dotted_order"].split(".")
             rebuilt = []
-            for ts, seg_id in segs[:-1]:
-                repl = uuid5(NAMESPACE_DNS, f"{seg_id}:{project_name}")
-                rebuilt.append(ts.strftime("%Y%m%dT%H%M%S%fZ") + str(repl))
-            ts_last, _ = segs[-1]
-            rebuilt.append(ts_last.strftime("%Y%m%dT%H%M%S%fZ") + str(new_id))
+            for part in segs[:-1]:
+                repl = uuid5(NAMESPACE_DNS, f"{part[-36:]}:{project_name}")
```
jacoblee93: See above, split on `Z`

python/langsmith/run_trees.py
angus-langchain: it gets set via tracing_context. Need to set via `from_headers` as well

python/langsmith/run_trees.py
angus-langchain: that's annoying cause then I need to add back the Z

angus-langchain: this way I can rejoin the 2 segments without having to do timestamp + 'Z' + id

angus-langchain:
> Seems fine, do we need to worry about patch at all?

wdym here? we use the same remap function for both, so patch should be sliced as well if parent run id is set

python/langsmith/run_trees.py
jacoblee93: Can we then use a variable? `TIMESTAMP_LENGTH = 36`

jacoblee93 (approved):
Good with me

jacoblee93:
> wdym here? we use the same remap function for both, so patch should be sliced as well if parent run id is set

Just checking

PR #1886, opened 2025-08-01
perf(python, js): Pop all allowed oob keys

angus-langchain:
## Description
Pop all OOB keys for better ingestion performance


akira (approved):

jacoblee93 (approved):
Basically we were duplicating some parts of the payload?

angus-langchain:
> Basically we were duplicating some parts of the payload?

We weren't putting them out of band, so we coudln't directly stream them to s3 or redis

PR #1892, opened 2025-08-01
Remove black dependency and update lock file

EugeneJinXin:
as discussed we have both black and ruff check in langsmith-sdk, no objection to leave ruff and remove black

They supposed to be compatible but not always in corner case, make lint fix a bit inconvenient.


jacoblee93 (approved):
They are def not compatible haha

PR #1898, opened 2025-08-04
chore(python): Make multipart endpoint the default batch config

angus-langchain:
### Description
Use multipart endpoint as the default ingest EP since we lazy load info.

Adds fallback to batch for older self hosted deployments


python/langsmith/client.py
```diff
@@ -1792,6 +1792,12 @@ def _multipart_ingest_ops(
         if acc_multipart:
             try:
                 self._send_multipart_req(acc_multipart)
+            except ls_utils.LangSmithNotFoundError:
+                # Fallback to batch ingest if multipart endpoint returns 404
+                # Filter out feedback operations as they're not supported in non-multipart mode
+                run_ops = [op for op in ops if isinstance(op, SerializedRunOperation)]
+                if run_ops:
+                    self._batch_ingest_run_ops(run_ops)
```
jacoblee93: Should this also set `acc_multipart`? Don't think we'd ever 404 if the server was down right?

jacoblee93 (approved):

python/langsmith/client.py
angus-langchain: yeah it should

PR #1917, opened 2025-08-08
feat(js): Allow direct wrapping of Jest/vitest methods, fix module resolution

jacoblee93:
Yarn monorepo hoisting causes issues in some configurations - I was able to repro when I had:

1. Only one version of `langsmith` used in the whole monorepo, which caused `langsmith` to be hoisted to the top of the monorepo
2. If the monorepo then has workspaces with multiple versions of Vitest, the vitest import within `langsmith` resolves to the wrong version (possibly due to a race condition)?

This caused issues where the version of vitest to run tests was not the same as the version used to declare tests with `ls.describe(...)`, resulting in missing test errors.

This PR adds a way for the user to ensure they are wrapping the right version, and adds path based resolution for the reporter so that the import maps to the correct location.

TODO: Test Jest for the same issue


js/src/vitest/utils/esm.mts
```diff
@@ -0,0 +1,41 @@
...
+    const packagePath = require.resolve(`vitest/package.json`, {
+      paths: [process.cwd()],
+    });
+    const pkg = require(packagePath);
+    const pkgDir = packagePath.replace("/package.json", "");
+    const esmEntry = !entrypoint
+      ? pkg.module ?? pkg.exports?.["."]?.import ?? "dist/index.js"
+      : pkg.exports?.[`./${entrypoint}`]?.import ??
+        pkg.exports?.[`./${entrypoint}`]?.default;
+    const path = `file://${pkgDir}/${esmEntry}`;
```
dqbd, line 25: Should we use `node:url` if possible? Think that this might not work for Windows?

jacoblee93: Yeah good point

PR #1920, opened 2025-08-11
feat: update SDK for org-scoped keys

keshivtandon:
### Description
Update SDKs to accept workspace or workspace ID specification for org-scoped keys

Closes INF-602

### Testing

- [x] Python unit tests
- [ ] Javascript unit tests
- [ ] Go unit tests

### DIFF
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -275,6 +275,27 @@ def _validate_api_key_if_hosted(api_url: str, api_key: Optional[str]) -> None:
             )
 
 
+def _validate_workspace_for_org_scoped_key(workspace: Optional[str], workspace_id: Optional[str]) -> None:
+    """Validate that workspace is specified for org-scoped keys.
+    
+    This validation is called when the backend returns an error indicating
+    that the API key is org-scoped and requires workspace specification.
+    
+    Args:
+        workspace (Optional[str]): The workspace name.
+        workspace_id (Optional[str]): The workspace ID.
+        
+    Raises:
+        LangSmithUserError: If neither workspace nor workspace_id is provided.
+    """
+    if not workspace and not workspace_id:
+        raise ls_utils.LangSmithUserError(
+            "This API key is org-scoped and requires workspace specification. "
+            "Please provide either 'workspace' or 'workspace_id' parameter, "
+            "or set LANGSMITH_WORKSPACE or LANGSMITH_WORKSPACE_ID environment variable."
+        )
+
+
 def _format_feedback_score(score: Union[float, int, bool, None]):
     """Format a feedback score by truncating numerical values to 4 decimal places.
 
@@ -313,7 +334,7 @@ def _get_tracing_sampling_rate(
     return sampling_rate
 
 
-def _get_write_api_urls(_write_api_urls: Optional[dict[str, str]]) -> dict[str, str]:
+def _get_write_api_urls(_write_api_urls: Optional[dict[str, str]], workspace: Optional[str] = None, workspace_id: Optional[str] = None) -> dict[str, str]:
     # Note: LANGSMITH_RUNS_ENDPOINTS is now handled via replicas, not _write_api_urls
     _write_api_urls = _write_api_urls or {}
     processed_write_api_urls = {}
@@ -399,6 +420,8 @@ class Client:
         "__weakref__",
         "api_url",
         "_api_key",
+        "_workspace",
+        "_workspace_id",
         "_headers",
         "retry_config",
         "timeout_ms",
@@ -448,6 +471,8 @@ class Client:
         api_urls: Optional[dict[str, str]] = None,
         otel_tracer_provider: Optional[TracerProvider] = None,
         tracing_sampling_rate: Optional[float] = None,
+        workspace: Optional[str] = None,
+        workspace_id: Optional[str] = None,
     ) -> None:
         """Initialize a Client instance.
 
@@ -488,6 +513,8 @@ class Client:
                 overrides the LANGCHAIN_TRACING_SAMPLING_RATE environment variable.
                 Should be a float between 0 and 1, where 1 means trace everything
                 and 0 means trace nothing.
+            workspace (Optional[str]): The workspace name. Required for org-scoped API keys.
+            workspace_id (Optional[str]): The workspace ID. Required for org-scoped API keys.
 
         Raises:
             LangSmithUserError: If the API key is not provided when using the hosted service.
@@ -509,8 +536,12 @@ class Client:
         self.tracing_sample_rate = _get_tracing_sampling_rate(tracing_sampling_rate)
         self._filtered_post_uuids: set[uuid.UUID] = set()
         self._write_api_urls: Mapping[str, Optional[str]] = _get_write_api_urls(
-            api_urls
+            api_urls, workspace, workspace_id
         )
+        # Initialize workspace attributes first
+        self._workspace = ls_utils.get_workspace(workspace)
+        self._workspace_id = ls_utils.get_workspace_id(workspace_id)
+        
         if self._write_api_urls:
             self.api_url = next(iter(self._write_api_urls))
             self.api_key = self._write_api_urls[self.api_url]
@@ -688,7 +719,25 @@ class Client:
         }
         if self.api_key:
             headers[X_API_KEY] = self.api_key
+        if self._workspace:
+            headers["X-Workspace"] = self._workspace
+        if self._workspace_id:
+            headers["X-Workspace-Id"] = self._workspace_id
         return headers
+
+    def _validate_workspace_requirements(self) -> None:
+        """Validate workspace requirements for org-scoped keys."""
+        _validate_workspace_for_org_scoped_key(self._workspace, self._workspace_id)
+
+    def _check_workspace_error(self, response: requests.Response) -> None:
+        """Check if response indicates a workspace-related error for org-scoped keys."""
+        try:
+            error_data = response.json()
+            error_message = error_data.get("detail", "")
+            if "workspace" in error_message.lower() and "org-scoped" in error_message.lower():
+                self._validate_workspace_requirements()
+        except (ValueError, KeyError):
+            pass
 
     @property
     def api_key(self) -> Optional[str]:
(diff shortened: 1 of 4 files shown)


js/src/client.ts
```diff
@@ -756,9 +765,41 @@ export class Client implements LangSmithTracingClientInterface {
...
+          this.validateWorkspaceRequirements();
+        }
+      } catch {
+        // Don't assume workspace error
+      }
+    }
+  }
+
+  public getWorkspaceId(): string | undefined {
+    return this.workspaceId;
```
jacoblee93: Why do we need this?

js/src/client.ts
```diff
@@ -827,6 +868,11 @@ export class Client implements LangSmithTracingClientInterface {
         ...this.fetchOptions,
       }
     );
+
+    if (!response.ok) {
+      void this.checkWorkspaceError(response);
```
jacoblee93: This will throw an uncatchable error if the function throws (which I think is what you intend?), you must await it

Can we also just instead add all this to `raiseForStatus`?

js/src/client.ts
```diff
@@ -756,9 +765,41 @@ export class Client implements LangSmithTracingClientInterface {
...
+    // Check for 400 status code or error messages containing workspace-related content
+    if (response.status === 403) {
+      try {
+        const errorData = await response.json();
+        const errorCode = errorData?.error;
+        if (errorCode === "org_scoped_key_requires_workspace") {
+          this.validateWorkspaceRequirements();
+        }
+      } catch {
+        // Don't assume workspace error
```
jacoblee93: This will just make the error disappear into the ether?

js/src/client.ts
```diff
@@ -1229,6 +1275,10 @@ export class Client implements LangSmithTracingClientInterface {
         ...this.fetchOptions,
       }
     );
+
+    if (!response.ok) {
+      void this.checkWorkspaceError(response);
```
jacoblee93: See above, never use `void` like this unless you know what you're doing

js/src/client.ts
keshivtandon: Wasn't needed, removed it

js/src/client.ts
keshivtandon: Yes, moved to raiseForStatus

js/src/client.ts
keshivtandon: Yeah, if it's a 403 with the specific error message I call validator, otherwise I just ignore the non-JSON error body and throw generic error as usual - specific error is just to call validator

js/src/client.ts
keshivtandon: Thank you, changed it!

js/yarn.lock
jacoblee93, line 1: Think your yarn version is wrong, just revert

js/src/utils/error.ts
```diff
@@ -85,6 +86,18 @@ export async function raiseForStatus(
     }
     return;
   }
+
+  if (response.status === 403 && onWorkspaceError) {
+    try {
+      const errorData = await response.json();
+      const errorCode = errorData?.error;
+      if (errorCode === "org_scoped_key_requires_workspace") {
+        onWorkspaceError();
```
jacoblee93: Do we actually need a separate hook here? What's the point vs just directly throwing the returned API error?

keshivtandon: You're right, it's pointless especially now that I'm using raiseForStatus - removed hook

js/src/utils/error.ts
```diff
@@ -85,6 +85,30 @@ export async function raiseForStatus(
...
+      }
+    } catch (e) {
+      if (
+        e &&
+        typeof e === "object" &&
+        "message" in e &&
+        typeof e.message === "string" &&
+        e.message.includes("org-scoped")
+      ) {
+        throw e;
```
jacoblee93: Why do this? Should probably just throw a generic `Forbidden` error since if this condition isn't met `response.text()` could double read the body and make an even stranger error

Maybe:

```ts
} catch (e: any) {
  throw new Error(`${response.status} ${response.statusText}`);
}
```

https://developer.mozilla.org/en-US/docs/Web/API/Response/statusText

js/src/client.ts
```diff
@@ -836,7 +848,7 @@ export class Client implements LangSmithTracingClientInterface {
         signal: AbortSignal.timeout(this.timeout_ms),
         ...this.fetchOptions,
       });
-      await raiseForStatus(res, `Failed to fetch ${path}`);
+      await raiseForStatus(res, `Failed to fetch ${path}`, false);
```
jacoblee93: Revert, this is the default

python/langsmith/client.py
```diff
@@ -760,8 +786,25 @@ def _compute_headers(self) -> dict[str, str]:
         }
         if self.api_key:
             headers[X_API_KEY] = self.api_key
+        if self._workspace_id:
+            headers["X-Tenant-Id"] = self._workspace_id
         return headers
 
+    def _validate_workspace_requirements(self) -> None:
+        """Validate workspace requirements for org-scoped keys."""
+        _validate_workspace_for_org_scoped_key(self._workspace_id)
```
jacoblee93: Too much indirection

python/langsmith/client.py
```diff
@@ -760,8 +786,25 @@ def _compute_headers(self) -> dict[str, str]:
...
+
+    def _check_workspace_error(self, response: requests.Response) -> None:
+        """Check if response indicates a workspace-related error for org-scoped keys."""
+        if response.status_code == 403:
+            try:
+                error_data = response.json()
+                error_code = error_data.get("error", "")
+                if error_code == "org_scoped_key_requires_workspace":
+                    self._validate_workspace_requirements()
```
jacoblee93: Why not just throw the error like we're doing for JS? Keep it simple

python/langsmith/client.py
```diff
@@ -772,6 +815,16 @@ def api_key(self, value: Optional[str]) -> None:
         object.__setattr__(self, "_api_key", value)
         object.__setattr__(self, "_headers", self._compute_headers())
 
+    @property
+    def workspace_id(self) -> Optional[str]:
```
jacoblee93, line 790: What's the point of this vs just exposing the prop directly?

python/langsmith/client.py
```diff
@@ -772,6 +815,16 @@ def api_key(self, value: Optional[str]) -> None:
...
+        """Return the workspace ID used for API requests."""
+        return self._workspace_id
+
+    @workspace_id.setter
+    def workspace_id(self, value: Optional[str]) -> None:
```
jacoblee93, line 795: See above

python/langsmith/client.py
```diff
@@ -954,13 +1007,21 @@ def request_with_retries(
                             raise ls_utils.LangSmithConflictError(
                                 f"Conflict for {pathname}. {repr(e)}{_context}"
                             )
+                        elif (response.status_code == 400 or response.status_code == 403):
+                            self._check_workspace_error(response)
```
jacoblee93: Why would this happen on 400

js/src/client.ts
keshivtandon: Changed!

js/src/utils/error.ts
keshivtandon: Changed

python/langsmith/client.py
keshivtandon: Yeah this is stupid, removed

python/langsmith/client.py
keshivtandon: Yep, was just waiting for JS revisions to be done before I matched Python side

python/langsmith/client.py
keshivtandon: Added a header setting function and simplified props to make things more direct. This simplifies the api_key and workspace_id setters, at least, since it removes duplication

python/langsmith/client.py
keshivtandon: Mb, it def wouldn't haha

python/langsmith/client.py
```diff
@@ -956,6 +977,22 @@ def request_with_retries(
...
+                            try:
+                                error_data = response.json()
+                                error_code = error_data.get("error", "")
+                                if error_code == "org_scoped_key_requires_workspace":
+                                    raise ls_utils.LangSmithUserError(
+                                        "This API key is org-scoped and requires workspace specification. "
+                                        "Please provide 'workspace_id' parameter, "
+                                        "or set LANGSMITH_WORKSPACE_ID environment variable."
+                                    )
+                            except (ValueError, KeyError):
```
jacoblee93, line 990: I assume this will catch if `response.json()` fails?

python/langsmith/client.py
```diff
@@ -762,17 +768,32 @@ def _compute_headers(self) -> dict[str, str]:
...
         if self.api_key:
             headers[X_API_KEY] = self.api_key
+        if self._workspace_id:
+            headers["X-Tenant-Id"] = self._workspace_id
         return headers
 
+    def _set_header_affecting_attr(self, attr_name: str, value: Any) -> None:
+        """Set attributes that affect headers and recalculate them."""
+        object.__setattr__(self, attr_name, value)
+        object.__setattr__(self, "_headers", self._compute_headers())
```
jacoblee93, line 778: I still don't really understand this (my Python is really bad), maybe @angus-langchain can better judge if this is a good idea/conventional

jacoblee93 (commented):
Good on JS, will let @angus-langchain answer a few questions on Python

python/langsmith/client.py
```diff
@@ -331,7 +331,7 @@ def _get_tracing_sampling_rate(
     return sampling_rate
 
 
-def _get_write_api_urls(_write_api_urls: Optional[dict[str, str]]) -> dict[str, str]:
+def _get_write_api_urls(_write_api_urls: Optional[dict[str, str]], workspace_id: Optional[str] = None) -> dict[str, str]:
```
angus-langchain: these are used anymore, does workspace_id depend on this function?

angus-langchain: seems like we don't actually use the param either

python/langsmith/client.py
angus-langchain: is this just a refactor?

angus-langchain (approved):
approving for python changes

python/langsmith/client.py
keshivtandon: yeah just refactored so I wasn't using it twice

python/langsmith/client.py
keshivtandon: yeah the ValueError catches for failures, KeyError is raised by error_data.get("error", "") if there are dictionary access issues

python/langsmith/client.py
keshivtandon: ah you're right, removed workspace_id

jacoblee93 (approved):

PR #1927, opened 2025-08-12
fix(py): Pass 'project_name' to _get_parent_run method

yakir-nevona:
Using `trace` context manager with "parent" argument (dotted_order) in a project that is not the default is not working due to this little missing parameter.


jacoblee93 (approved):
Thank you!

PR #1928, opened 2025-08-12
feat(js): Allow async processOutputs methods for traceables

jacoblee93:
Also adds to streamObject and streamText AI SDK calls

### DIFF
--- a/js/src/traceable.ts
+++ b/js/src/traceable.ts
@@ -148,11 +148,11 @@ const handleRunInputs = (
 
 const _extractUsage = (runData: {
   runTree: RunTree;
-  outputs: KVMap;
+  outputs?: KVMap;
 }): ExtractedUsageMetadata | undefined => {
   const usageMetadataFromMetadata = (runData.runTree.extra.metadata ?? {})
     .usage_metadata;
-  return runData.outputs.usage_metadata ?? usageMetadataFromMetadata;
+  return runData.outputs?.usage_metadata ?? usageMetadataFromMetadata;
 };
 
 function validateExtractedUsageMetadata(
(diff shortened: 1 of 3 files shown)


js/src/tests/fetch.test.ts
```diff
@@ -99,7 +99,7 @@ describe.each([[""], ["mocked"]])("Client uses %s fetch", (description) => {
     });
   });
 
-  test("basic traceable implementation", async () => {
+  test.only("basic traceable implementation", async () => {
```
dqbd: Remove `only`

js/src/traceable.ts
```diff
@@ -198,28 +245,34 @@ function handleRunOutputs(params: {
 
   try {
     outputs = processOutputsFn(outputs);
+    if (isAsync(processOutputsFn)) {
```
dqbd: _nit_: would also check if it's a promise in case of a sync function that returns a promise

js/src/traceable.ts
```diff
@@ -954,21 +995,25 @@ export function traceable<Func extends (...args: any[]) => any>(
...
                 return rawOutput;
               }
             },
             async (error: unknown) => {
               await currentRunTree?.end(undefined, String(error));
-              await handleEnd();
+              await handleEnd({
+                runTree: currentRunTree,
+                postRunPromise,
+                on_end: config?.on_end,
```
dqbd, line 1019: for the future, can we make this as `onEnd`?

dqbd (approved):

js/src/traceable.ts
jacoblee93: Intentionally did not do this to preserve behavior. Will add a note to consider this on a minor bump

js/src/traceable.ts
jacoblee93: Will make a note

PR #1932, opened 2025-08-14
feat(py): Adds process_buffered_run_ops method to Python SDK

jacoblee93:
At scale, some APIs (such as anonymization) hit rate limits when run over a single run at a time. This PR adds a low-level hook called `process_buffered_run_ops` and a new `run_ops_buffer_size` method that buffers run operations and allows the user to run such APIs over an entire batch of runs at once.

This needed to be a separate buffer because for multipart, we convert runs to multipart and then compress them when batching. Otherwise we'd have to do something really inefficient like decompressing and recompressing batches.

```py
    def track_runs(runs):
        # transform run ops here, e.g. anonymization
        return runs

    client = Client(
        process_buffered_run_ops=track_runs,
        run_ops_buffer_size=1024,
        run_ops_buffer_timeout_ms=1000,
    )
```

### DIFF
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -489,7 +489,10 @@ class Client:
             process_buffered_run_ops (Optional[Callable[[Sequence[dict]], Sequence[dict]]]): A function applied to buffered run operations
                 that allows for modification of the raw run dicts before they are converted to multipart and compressed.
                 This is useful specifically for high throughput tracing where you need to apply a rate-limited API or other
-                costly process to the runs before they are sent to the API.
+                costly process to the runs before they are sent to the API. Note that the buffer will only flush automatically
+                when run_ops_buffer_size is reached or a new run is added to the buffer after run_ops_buffer_timeout_ms
+                has elapsed - it will not flush outside of these conditions unless you manually
+                call client.flush(), so be sure to do this before your code exits.
             run_ops_buffer_size (Optional[int]): Maximum number of run operations to collect in the buffer before applying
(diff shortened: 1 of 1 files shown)


python/langsmith/client.py
```diff
@@ -414,6 +414,10 @@ class Client:
         "_hide_inputs",
         "_hide_outputs",
         "_hide_metadata",
+        "_process_buffered_run_ops",
```
angus-langchain, line 417: do we need to store all this configuration on the client directly?

python/tests/integration_tests/test_client.py
```diff
@@ -3333,3 +3333,105 @@ def child():
     finally:
         if langchain_client.has_project(project_name=project_name):
             langchain_client.delete_project(project_name=project_name)
+
+
+def test_run_ops_buffer_integration(langchain_client: Client) -> None:
```
angus-langchain, line 3338: just checking, did you run these locally and confirm they pass? Can't tell because rate limits on our test runner

python/langsmith/client.py
jacoblee93: What's the alternative?

python/tests/integration_tests/test_client.py
jacoblee93: Yeah it runs fine, can rerun int tests

jacoblee93: Actually failing int test is unrelated

PR #1936, opened 2025-08-14
chore(py): correct references to `langchain-core` and hyperlink

mdrxy:


python/langsmith/async_client.py
```diff
@@ -1712,7 +1712,7 @@ async def pull_prompt(
     ) -> Any:
         """Pull a prompt and return it as a LangChain PromptTemplate.
 
-        This method requires `langchain_core`.
+        This method requires `langchain-core <https://pypi.org/project/langchain-core/>`__.
```
jacoblee93, line 1715: nit: remove trailing `__`? Or intentional

python/langsmith/client.py
```diff
@@ -7259,7 +7259,7 @@ def pull_prompt(
     ) -> Any:
         """Pull a prompt and return it as a LangChain PromptTemplate.
 
-        This method requires `langchain_core`.
+        This method requires `langchain-core <https://pypi.org/project/langchain-core/>`__.
```
jacoblee93, line 7262: See above

PR #1938, opened 2025-08-15
feat(js): Improve tracing for multimodal AI SDK runs

jacoblee93:


js/src/experimental/vercel/utils.ts
```diff
@@ -0,0 +1,188 @@
...
+          length: Object.keys(fileData).length,
+        })
+      );
+    }
+    if (uint8Array) {
+      let binary = "";
+      for (let i = 0; i < uint8Array.length; i++) {
+        binary += String.fromCharCode(uint8Array[i]);
+      }
+      const base64 = btoa(binary);
```
dqbd, line 99: _nit_: could we update `guessMimeType` to handle UInt8Array? Don't think we need to encode to Base64 just to decode it later

jacoblee93: Yeah can do

js/src/experimental/vercel/middleware.ts
```diff
@@ -157,7 +120,8 @@ export function LangSmithMiddleware(config?: {
           },
         }
       );
-      return traceableFunc(params);
+      const res = await traceableFunc(params);
```
dqbd, line 123: _q_: is this intended?

js/src/experimental/vercel/utils.ts
jacoblee93: Well actually, we always send up as base64 so no decoding

js/src/experimental/vercel/utils.ts
```diff
@@ -0,0 +1,188 @@
...
+    return fileData.toString();
+  }
+  let normalizedFileData: string;
+  if (typeof fileData !== "string") {
+    let uint8Array;
+    if (
+      fileData != null &&
+      typeof fileData === "object" &&
+      "type" in fileData &&
+      "data" in fileData
```
dqbd, line 81: _nit_: should we also check if `fileData.data instanceof Uint8Array`?

js/src/experimental/vercel/middleware.ts
jacoblee93: Yeah seeing if it helps some subtle timing issues

js/src/experimental/vercel/utils.ts
```diff
@@ -0,0 +1,188 @@
...
+    ) {
+      // Typing is wrong here if a buffer is passed in
+      uint8Array = new Uint8Array(fileData.data as Uint8Array);
```
dqbd: _q_: isn't NodeJS buffer already an instanceof UInt8Array?

js/src/experimental/vercel/utils.ts
```diff
@@ -0,0 +1,188 @@
...
+      // eslint-disable-next-line no-instanceof/no-instanceof
+    } else if (fileData instanceof ArrayBuffer) {
+      uint8Array = new Uint8Array(fileData);
+    } else if (
+      fileData != null &&
+      typeof fileData === "object" &&
+      Object.keys(fileData).every((key) => !isNaN(Number(key)))
```
dqbd: _q_: is this potentially slow?

dqbd (approved):

PR #1940, opened 2025-08-18
feat(python): Pass api config through background thread

angus-langchain:
### Description
Currently `LANGSMITH_RUN_ENDPOINTS` is not respected. This PR acts fixes the issue by allowing api_url/api_key passthrough in the background thread to multipart/batch ingest methods

### Limitations
Does not work with compression, this is likely possible to implement but would require further scoping and coordination with other sdk projects that will likely change how we handle compression

Does not work with feedback, `LANGSMITH_RUN_ENDPOINTS` did not work completely with feedback either, and only worked for multipart feedback, not feedback created directly via `create_feedback` without backgrounding

Does not work with feedback yet. 
### Tests
- [x] Added unit tests
- [x] Tested locally

### DIFF
--- a/python/langsmith/__init__.py
+++ b/python/langsmith/__init__.py
@@ -21,7 +21,7 @@ if TYPE_CHECKING:
 
 # Avoid calling into importlib on every call to __version__
 
-__version__ = "0.4.13"
+__version__ = "0.4.14"
 version = __version__  # for backwards compatibility
 
 
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "langsmith"
-version = "0.4.13"
+version = "0.4.14"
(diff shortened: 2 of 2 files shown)


python/langsmith/_internal/_background_thread.py
```diff
@@ -43,6 +43,19 @@
...
+    batch: list[TracingQueueItem],
+) -> dict[tuple[Optional[str], Optional[str]], list[TracingQueueItem]]:
+    """Group batch items by (api_url, api_key) combination."""
+    from collections import defaultdict
+
+    grouped = defaultdict(list)
+    for item in batch:
+        key = (item.api_url, item.api_key)
+        grouped[key].append(item)
+    return dict(grouped)
```
hinthornw: ooc: why do we need this?

angus-langchain: good catch, not needed

PR #1941, opened 2025-08-18
fix(js): Fix typing for AI SDK stream methods

jacoblee93:
- Allow passing generics through as expected
- Make `streamText` and `streamObject` return values directly instead of promises
- Pass all args through in addition to just the first one (though no AI SDK methods currently do this)

### DIFF
--- a/js/src/experimental/vercel/index.ts
+++ b/js/src/experimental/vercel/index.ts
@@ -91,15 +91,15 @@ const _formatTracedInputs = (params: Record<string, unknown>) => {
  */
 const wrapAISDK = <
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  WrapLanguageModelType extends (...args: any[]) => any,
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   GenerateTextType extends (...args: any[]) => any,
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   StreamTextType extends (...args: any[]) => any,
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  GenerateObjectType extends (...args: any[]) => any,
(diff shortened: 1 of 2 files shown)


js/src/experimental/vercel/index.ts
```diff
@@ -90,16 +90,16 @@ const _formatTracedInputs = (params: Record<string, unknown>) => {
...
   GenerateTextType extends (...args: any[]) => any,
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   StreamTextType extends (...args: any[]) => any,
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  GenerateObjectType extends (...args: any[]) => any,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   StreamObjectType extends (...args: any[]) => any,
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  GenerateObjectType extends (...args: any[]) => any
+  WrapLanguageModelType extends (...args: any[]) => any
```
jacoblee93, line 102: Unlikely anyone is directly passing generics here at the moment, and this makes the ordering make more sense if anyone needs to in the future

js/src/experimental/vercel/index.ts
```diff
@@ -393,16 +405,16 @@ const wrapAISDK = <
...
-    return traceableFunc(params);
+    return traceableFunc(...args);
   };
 
   return {
-    generateText: wrappedGenerateText,
-    generateObject: wrappedGenerateObject,
-    streamText: wrappedStreamText,
-    streamObject: wrappedStreamObject,
+    generateText: wrappedGenerateText as GenerateTextType,
```
jacoblee93, line 414: This was the key bit to allow generics to be passed

PR #1943, opened 2025-08-19
feat: add extra & metadata into dataset pydantic schema

EugeneJinXin:
Issue: The create_dataset() method accepts a metadata parameter, but the created dataset doesn't expose this metadata through the Dataset schema or any retrieval methods.  [Linear tickert](https://linear.app/langchain/issue/LS-3906/issue-dataset-metadata-parameter-in-create-dataset-not-accessible-via
)

Also addressed LS backend to make sure extra is returned here https://github.com/langchain-ai/langchainplus/pull/11649/files

### DIFF
--- a/python/langsmith/schemas.py
+++ b/python/langsmith/schemas.py
@@ -283,6 +283,7 @@ class Dataset(DatasetBase):
     inputs_schema: Optional[dict[str, Any]] = None
     outputs_schema: Optional[dict[str, Any]] = None
     transformations: Optional[list[DatasetTransformation]] = None
+    extra: Optional[dict[str, Any]] = None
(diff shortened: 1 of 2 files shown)


jacoblee93 (changes requested):
Actually - worth checking to make sure you are able to set metadata fields on the model

EugeneJinXin:
> Actually - worth checking to make sure you are able to set metadata fields on the model

yes good catch a @property would make it read-only. 
Good thing is that in my lastest commit I no longer need to expose 'extra' so i ditched this @property thing all together

EugeneJinXin:
dependency is merged https://github.com/langchain-ai/langchainplus/pull/11649/files

PR #1944, opened 2025-08-19
feat(js): Allow passing runtime LangSmith config to AI SDK wrapper

jacoblee93:


js/src/experimental/vercel/index.ts
```diff
@@ -74,6 +74,99 @@ const _formatTracedInputs = (params: Record<string, unknown>) => {
...
+    | "child_runs"
+    | "parent_run"
+    | "error"
+    | "serialized"
+  >
+> & {
+  processInputs?: (inputs: {
+    formatted: Record<string, unknown>;
+    raw: Record<string, unknown>;
+  }) => Record<string, unknown>;
```
jacoblee93: Provide both formatted and raw values to user override here

PR #1947, opened 2025-08-19
fix(js): Fix retry issues

jacoblee93:


js/src/utils/async_caller.ts
```diff
@@ -99,23 +92,22 @@ export class AsyncCaller {
...
                 throw error;
               }
-              // eslint-disable-next-line @typescript-eslint/no-explicit-any
-              const response: Response | undefined = (error as any)?.response;
-              const status = response?.status;
+              const response: Response | undefined = error?.response;
+              const status = response?.status ?? error?.status;
               if (status) {
-                if (STATUS_NO_RETRY.includes(+status)) {
+                if (!STATUS_RETRYABLE.includes(+status)) {
```
dqbd, line 115: We should handle `STATUS_IGNORE` before `!STATUS_RETRYABLE`

js/src/tests/vercel/vercel_node_otel.int.test.ts
```diff
@@ -12,124 +12,126 @@ import { traceable } from "../../traceable.js";
 
 const client = new Client();
 
-test("nested generateText", async () => {
-  const sdk = new NodeSDK({
-    traceExporter: new AISDKExporter(),
-    instrumentations: [getNodeAutoInstrumentations()],
-  });
+describe.skip("Legacy AI SDK tests", () => {
+  test("nested generateText", async () => {
```
hinthornw, line 16: ooc what's the point of skipping it at this point instead of removing if it's legacy?

jacoblee93: In case there's someone out there still using the old flow in a critical way who we need to debug for ;)

js/src/client.ts
```diff
@@ -2091,16 +2117,18 @@ export class Client implements LangSmithTracingClientInterface {
...
-        body: JSON.stringify(filteredPayload),
-        signal: AbortSignal.timeout(this.timeout_ms),
-        ...this.fetchOptions,
+      const response = await this.caller.call(async () => {
+        const res = await this._fetch(url, {
+          method: "POST",
+          headers: { ...this.headers, "Content-Type": "application/json" },
+          signal: AbortSignal.timeout(this.timeout_ms),
+          ...this.fetchOptions,
+          body: JSON.stringify(filteredPayload),
```
dqbd: Serialization happening inside `this.caller`

jacoblee93: Thanks fixing

js/yarn.lock
```diff
@@ -5099,6 +5099,19 @@ langsmith@^0.3.33:
     semver "^7.6.3"
     uuid "^10.0.0"
 
+langsmith@^0.3.46:
```
dqbd, line 5102: We shouldn't have `langsmith` here I think?

jacoblee93: Circular dep issue - we probably shouldn't have two but it's just for testing anyway so not overly worried.

PR #1948, opened 2025-08-20
chore(python): Prefer provided project_name to environment in tracing context

hinthornw:
When using a dotted_order style parent, the provided project_name wasn't being applied to that constructed run tree.

Also updated lockfile which seems to have given the formatter a field day.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,10 +32,6 @@ jobs:
           enable-cache: true
       - name: Build project for distribution
         run: cd python && uv build
-      - name: Check Version
-        id: check-version
-        run: |
-          cd python && echo version=$(grep '^version =' pyproject.toml | sed 's/version = "\(.*\)"/\1/') >> $GITHUB_OUTPUT
       - name: Create Release
         uses: ncipollo/release-action@v1
         with:
--- a/python/tests/unit_tests/test_run_helpers.py
+++ b/python/tests/unit_tests/test_run_helpers.py
@@ -1613,7 +1613,9 @@ def test_set_parent_and_project_name():
             assert rt.session_name == project_name
             assert str(rt.id) == str(rt.trace_id)
             assert str(rt.trace_id) == trace_id
-            assert rt.start_time == datetime.datetime.fromisoformat(start_time)
+            assert rt.start_time == datetime.datetime.strptime(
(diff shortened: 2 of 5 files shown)


python/langsmith/run_helpers.py
```diff
@@ -141,7 +141,12 @@ def tracing_context(
         )
     current_context = get_tracing_context()
     parent_run = (
-        _get_parent_run({"parent": parent or kwargs.get("parent_run")})
+        _get_parent_run(
+            {
+                "parent": parent or kwargs.get("parent_run"),
+                "project_name": project_name,
```
hinthornw, line 147: This is the change here.

python/tests/unit_tests/test_run_helpers.py
```diff
@@ -1599,6 +1600,22 @@ async def test_trace_respects_env_var(env_var: bool, context: Optional[bool]):
         assert not mock_calls
 
 
+def test_set_parent_and_project_name():
```
hinthornw, line 1603: Added a test

.github/workflows/release.yml
```diff
@@ -32,10 +32,6 @@ jobs:
           enable-cache: true
       - name: Build project for distribution
         run: cd python && uv build
-      - name: Check Version
```
jacoblee93, line 35: Why remove this?

python/langsmith/run_trees.py
```diff
@@ -1113,7 +1113,9 @@ def _parse_dotted_order(dotted_order: str) -> list[tuple[datetime, UUID]]:
     parts = dotted_order.split(".")
     return [
         (
-            datetime.strptime(part[:-TIMESTAMP_LENGTH], "%Y%m%dT%H%M%S%fZ"),
+            datetime.strptime(part[:-TIMESTAMP_LENGTH], "%Y%m%dT%H%M%S%fZ").replace(
```
jacoblee93, line 1116: This makes me slightly nervous

.github/workflows/release.yml
hinthornw: because i switched to dynamic version

python/langsmith/run_trees.py
hinthornw: can remove if you want. we were just ignoring anyway

PR #1958, opened 2025-08-25
fix: move otel check to client init time to avoid timing issue

EugeneJinXin:
**Issue:** for dynamic values like env var, you usually want to check them inside the instance or function, not at import time. because under the hood, python Imports are cached (in sys.modules), so in the case of some customer who's using pytest with otel,  their  second import of langsmith.client (during pytest unit test run time) weren't able to override initial one (during langsmith_plugin pytest plugin import, happened at very first of pytest set up).  This is causing subtle issue like if pytest plugin run first (`has_otel=False`), then later even if env var got reset (`has_otel=True`) the python still doesn't recognize the change.

I also found that the env var cache a bit problematic in such case, but if we fix the import, that might be less concerning, so i just reduced max size a little bit to make it less likely to be an issue.

### DIFF
--- a/python/langsmith/_internal/otel/_otel_client.py
+++ b/python/langsmith/_internal/otel/_otel_client.py
@@ -4,9 +4,9 @@ import os
 
 from langsmith import utils as ls_utils
 
-HAS_OTEL = False
-try:
-    if ls_utils.is_truish(ls_utils.get_env_var("OTEL_ENABLED")):
+def _import_otel_client():
+    """Dynamically import OTEL client modules when needed."""
+    try:
         from opentelemetry.exporter.otlp.proto.http.trace_exporter import (  # type: ignore[import]
             OTLPSpanExporter,
         )
@@ -19,9 +19,12 @@ try:
             BatchSpanProcessor,
         )
 
-        HAS_OTEL = True
-except ImportError:
-    pass
+        return OTLPSpanExporter, SERVICE_NAME, Resource, TracerProvider, BatchSpanProcessor
+    except ImportError:
+        raise ImportError(
+            "OpenTelemetry packages are required to use this function. "
+            "Please install with `pip install langsmith[otel]`"
+        )
 
 
 def get_otlp_tracer_provider() -> "TracerProvider":
@@ -40,12 +43,8 @@ def get_otlp_tracer_provider() -> "TracerProvider":
     Returns:
         TracerProvider: The OTLP tracer provider.
     """
-    # Set LangSmith-specific defaults if not already set in environment
-    if not HAS_OTEL:
-        raise ImportError(
-            "OpenTelemetry packages are required to use this function. "
-            "Please install with `pip install langsmith[otel]`"
-        )
+    # Import OTEL modules dynamically
+    OTLPSpanExporter, SERVICE_NAME, Resource, TracerProvider, BatchSpanProcessor = _import_otel_client()
 
     if "OTEL_EXPORTER_OTLP_ENDPOINT" not in os.environ:
         ls_endpoint = ls_utils.get_api_url(None)
--- a/python/langsmith/_internal/otel/_otel_client.py
+++ b/python/langsmith/_internal/otel/_otel_client.py
@@ -1,8 +1,11 @@
 """Client configuration for OpenTelemetry integration with LangSmith."""
+# ruff: noqa: F821
 
 import os
+import warnings
 
 from langsmith import utils as ls_utils
+
 
 def _import_otel_client():
     """Dynamically import OTEL client modules when needed."""
@@ -19,12 +22,18 @@ def _import_otel_client():
             BatchSpanProcessor,
         )
 
-        return OTLPSpanExporter, SERVICE_NAME, Resource, TracerProvider, BatchSpanProcessor
-    except ImportError:
-        raise ImportError(
-            "OpenTelemetry packages are required to use this function. "
-            "Please install with `pip install langsmith[otel]`"
+        return (
+            OTLPSpanExporter,
+            SERVICE_NAME,
+            Resource,
+            TracerProvider,
+            BatchSpanProcessor,
         )
+    except ImportError as e:
+        warnings.warn(
+            f"OTEL_ENABLED is set but OpenTelemetry packages are not installed: {e}"
+        )
+        return None
 
 
 def get_otlp_tracer_provider() -> "TracerProvider":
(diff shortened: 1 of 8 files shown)


python/langsmith/client.py
```diff
@@ -94,9 +94,15 @@
...
 from langsmith.schemas import AttachmentInfo
 
-HAS_OTEL = False
-try:
-    if ls_utils.is_truish(ls_utils.get_env_var("OTEL_ENABLED")):
+
+def _check_otel_enabled() -> bool:
+    """Check if OTEL is enabled and imports are available."""
+    return ls_utils.is_truish(ls_utils.get_env_var("OTEL_ENABLED"))
+
```
EugeneJinXin, line 101: this is the main change

python/langsmith/_internal/otel/_otel_client.py
```diff
@@ -40,12 +59,20 @@ def get_otlp_tracer_provider() -> "TracerProvider":
     Returns:
         TracerProvider: The OTLP tracer provider.
     """
-    # Set LangSmith-specific defaults if not already set in environment
-    if not HAS_OTEL:
+    # Import OTEL modules dynamically
+    otel_imports = _import_otel_client()
```
jacoblee93, line 62: We are now importing these modules directly many times and I think it's in the tracing hot path - would check into whether there's a perf hit or if we should be caching this after the first import

Not sure about Python's module loading perf

python/langsmith/utils.py
```diff
@@ -364,7 +364,7 @@ def is_base_message_like(obj: object) -> bool:
     )
 
 
-@functools.lru_cache(maxsize=100)
+@functools.lru_cache(maxsize=10)
```
jacoblee93: Why change this?

python/uv.lock
jacoblee93, line 1: Why is this changing?

python/tests/unit_tests/test_multiple_endpoints.py
jacoblee93, line 1: Why is this changing?

python/langsmith/_internal/otel/_otel_exporter.py
```diff
@@ -107,26 +129,43 @@ def __init__(self, tracer_provider=None):
...
-        self._tracer = trace.get_tracer("langsmith", tracer_provider=tracer_provider)
-        self._spans = {}
+            self._spans = {}
+            self._otel_available = True
 
     def export_batch(
         self,
         operations: list[SerializedRunOperation],
-        otel_context_map: dict[uuid.UUID, Optional[Context]],
+        otel_context_map: dict[uuid.UUID, Optional[Any]],  # type: ignore[misc]
```
jacoblee93: Why are we broadening this type?

python/langsmith/_internal/otel/_otel_exporter.py
```diff
@@ -168,8 +207,8 @@ def _create_span_for_run(
         self,
         op: SerializedRunOperation,
         run_info: dict,
-        otel_context: Optional[Context] = None,
-    ) -> Optional[Span]:
+        otel_context: Optional[Context] = None,  # type: ignore
```
jacoblee93: Why are we ignoring types?

python/langsmith/_internal/otel/_otel_client.py
EugeneJinXin: Think this _import_otel_client is mostly just one time import at client __init__ time, unless customer init the client multiple times.  
This PR is just changing when that import happens, not the frequency of it.

python/langsmith/_internal/otel/_otel_exporter.py
EugeneJinXin: there's definitely ways to deal with the mypy issues better, let me do a push on mypy issues

python/langsmith/_internal/otel/_otel_client.py
jacoblee93: Ok cool, my bad then

python/langsmith/_internal/otel/_otel_exporter.py
jacoblee93: Yeah if too hard it's fine but don't love the idea of broadening if it can be avoided

python/langsmith/_internal/otel/_otel_exporter.py
```diff
@@ -127,6 +163,10 @@ def export_batch(
         Args:
             operations: List of serialized run operations to export.
         """
+        # Return early if OTEL not available
+        if not self._otel_available:
```
jacoblee93: What was this doing before?

python/langsmith/client.py
```diff
@@ -105,11 +111,12 @@
         )
         from langsmith._internal.otel._otel_exporter import OTELExporter
 
-        HAS_OTEL = True
-except ImportError:
-    raise ImportError(
-        "To use OTEL tracing, you must install it with `pip install langsmith[otel]`"
-    )
+        return otel_trace, set_span_in_context, get_otlp_tracer_provider, OTELExporter
```
jacoblee93, line 114: Should we refactor everything to use a single import method?

python/langsmith/client.py
```diff
@@ -665,33 +684,48 @@ def __init__(
...
+
+                # Store imports for later use
+                self._otel_trace = otel_trace
+                self._set_span_in_context = set_span_in_context
+
+            except ImportError as e:
                 warnings.warn(
-                    "LANGSMITH_OTEL_ENABLED is set but OpenTelemetry packages are not installed. "
-                    "Install with `pip install langsmith[otel]`"
+                    f"LANGSMITH_OTEL_ENABLED is set but OpenTelemetry packages are not installed: {e}"
```
jacoblee93: Why remove the install prompt?

python/tests/unit_tests/test_multiple_endpoints.py
jacoblee93, line 1: Is our CI broken?

jacoblee93 (approved):
Thank you!

python/langsmith/client.py
EugeneJinXin: my bad, added back

python/tests/unit_tests/test_multiple_endpoints.py
EugeneJinXin: passed in the first run, seems just very flaky

EugeneJinXin:
Merging since the integration tests passes and my [script](https://docs.google.com/document/d/1hgS-YPAqvavbkYN0DyrFkjd91lD-n0n6kAedAfAiCFM/edit?tab=t.0) based tests passed, also wanna have a version for Apple to test

python/tests/unit_tests/test_multiple_endpoints.py
jacoblee93: No I mean why did this get reformatted, CI should reject non-formatted files

PR #1959, opened 2025-08-25
fix(js): Allow null to be passed as a first arg to traceable

### LINKED ISSUE #1911: Issue: Passing null to a traced function crashes with hard-to-debug error

This code reliably throws a hard-to-understand error:
```ts
import { traceable } from "langsmith/traceable"
await traceable(
    async (foo?: string) => foo
)(null)
```
```
TypeError: Cannot read properties of null (reading 'callbacks')
    at isRunnableConfigLike ({redacted}/node_modules/langsmith/dist/run_trees.cjs:748:18)
    at {redacted}/node_modules/langsmith/dist/traceable.cjs:395:57
    at traceableFunc {redacted}/node_modules/langsmith/dist/traceable.cjs:441:11)
   {app code}
```

However, passing `undefined` works:
```ts
import { traceable } from "langsmith/traceable"
await traceable(
    async (foo?: string) => foo
)(undefined)
```

Now, I'm aware that this isn't strictly legal typescript if you enable strict null checks (this library should probably support older TypeScript codebases that don't enable this check). In any case, this should just compile down to `await traceable(async (foo) => foo)(null);` which is perfectly valid javascript.

### Suggestion:

I'll leave the intended behavior to the experts - though at the very least, detecting this case and throwing a useful error would save folks from some hair-pulling.

jacoblee93:
Fixes #1911

### DIFF
--- a/js/src/run_trees.ts
+++ b/js/src/run_trees.ts
@@ -766,7 +766,7 @@ export class RunTree implements BaseRun {
 
 export function isRunTree(x?: unknown): x is RunTree {
   return (
-    x !== undefined &&
+    x != null &&
     typeof (x as RunTree).createChild === "function" &&
     typeof (x as RunTree).postRun === "function"
   );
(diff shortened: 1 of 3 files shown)


EugeneJinXin (approved):
lgtm

PR #1961, opened 2025-08-26
fix(py,js): Trace AzureOpenAI clients with proper provider for playground

jacoblee93:
CC @catherine-langchain

### DIFF
--- a/js/src/wrappers/openai.ts
+++ b/js/src/wrappers/openai.ts
@@ -1,4 +1,4 @@
-import { OpenAI, AzureOpenAI } from "openai";
+import type { OpenAI } from "openai";
 import type { APIPromise } from "openai";
 import type { RunTreeConfig } from "../index.js";
 import {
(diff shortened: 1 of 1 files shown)


python/langsmith/wrappers/_openai.py
```diff
@@ -393,13 +393,26 @@ def wrap_openai(
     """  # noqa: E501
     tracing_extra = tracing_extra or {}
 
+    ls_provider = "openai"
+    try:
+        from openai import AsyncAzureOpenAI, AzureOpenAI
+
+        if isinstance(client, AzureOpenAI) or isinstance(client, AsyncAzureOpenAI):
+            ls_provider = "azure"
```
EugeneJinXin, line 401: Screenshot 2025-08-25 at 7 03 05 PM
should we call it azure or  azure_openai? i really don't know, think i've heard the latter more and more aligned with our playground naming?

jacoblee93: This is respected by our backend already

PR #1966, opened 2025-08-26
fiz(python): Alleviate memory allocations from zstd flush

angus-langchain:
### Description
Reduce calls to flush zstd buffer

### Changes
- Use uncompressed size to determine if we should flush or not.

### Results
15% decrease in peak memory usage

### DIFF
--- a/python/langsmith/_internal/_background_thread.py
+++ b/python/langsmith/_internal/_background_thread.py
@@ -129,9 +129,6 @@ def _tracing_thread_drain_compressed_buffer(
         if client.compressed_traces is None:
             return None, None
         with client.compressed_traces.lock:
-            client.compressed_traces.compressor_writer.flush()
-            current_size = client.compressed_traces.buffer.tell()
-
             pre_compressed_size = client.compressed_traces.uncompressed_size
 
             if size_limit is not None and size_limit <= 0:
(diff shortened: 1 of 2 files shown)


python/langsmith/_internal/_background_thread.py
```diff
@@ -141,7 +138,10 @@ def _tracing_thread_drain_compressed_buffer(
                     f"size_limit_bytes must be nonnegative; got {size_limit_bytes}"
                 )
 
-            if (size_limit_bytes is None or current_size < size_limit_bytes) and (
+            if (
+                size_limit_bytes is None
+                or pre_compressed_size < size_limit_bytes
```
jacoblee93, line 143: So before we were batching based on the compressed size, now we're going by the uncompressed size?

This will be smaller batches in general right?

python/langsmith/_internal/_background_thread.py
```diff
@@ -129,9 +129,6 @@ def _tracing_thread_drain_compressed_buffer(
         if client.compressed_traces is None:
             return None, None
         with client.compressed_traces.lock:
-            client.compressed_traces.compressor_writer.flush()
```
jacoblee93, line 132: This is safe to remove?

angus-langchain: close() also flushes.

python/langsmith/_internal/_background_thread.py
angus-langchain: Yeah it will, but won’t be calling flush unnecessarily. We can also raise the batch config size

angus-langchain: We also already trace every half second

jacoblee93 (approved):

PR #1967, opened 2025-08-26
feat(js): Add OpenAI service tier to input and output token details for AI SDK

jacoblee93:


python/tests/integration_tests/wrappers/test_openai.py
```diff
@@ -271,6 +271,17 @@ def _collect_requests(mock_session: mock.MagicMock, filename: str):
         "expect_usage_metadata": True,
         "check_reasoning_tokens": True,
     },
+    # just test flex as priority can randomly downgrade
```
EugeneJinXin, line 274: interesting, so tier can downgrade from 'priority' to to 'flex' even if you paying more (CORRECTION: wanted to say "even if you choose priority initially") ? if such downgrade happen the charge will be only for flex, is that correct?

python/langsmith/wrappers/_openai.py
```diff
@@ -229,7 +250,9 @@ def _process_chat_completion(outputs: Any):
         rdict = outputs.model_dump()
         oai_token_usage = rdict.pop("usage", None)
         rdict["usage_metadata"] = (
-            _create_usage_metadata(oai_token_usage) if oai_token_usage else None
+            _create_usage_metadata(oai_token_usage, rdict.get("service_tier"))
```
EugeneJinXin, line 253: sure no but just double check,  do we need to worry about _process_responses_api_output() of this file which also uses _create_usage_metadata

python/tests/integration_tests/wrappers/test_openai.py
jacoblee93: `priority` will download to `default` if capacity isn't there

python/langsmith/wrappers/_openai.py
jacoblee93: Ah yeah good catch

EugeneJinXin:
strange to see experimental word, any legacy reason for that? 
js/src/**experimental**/vercel/middleware.ts

jacoblee93:
> strange to see experimental word, any legacy reason for that? js/src/**experimental**/vercel/middleware.ts

Because AI SDK is looking into other ways of non-OTEL official telemetry

PR #1977, opened 2025-08-28
feat(py,js): Adds split parameter to test runners

jacoblee93:


python/langsmith/testing/_internal.py
```diff
@@ -69,6 +69,9 @@ def test(
     output_keys: Optional[Sequence[str]] = None,
     client: Optional[ls_client.Client] = None,
     test_suite_name: Optional[str] = None,
+    metadata: Optional[dict] = None,
+    repetitions: Optional[int] = None,
```
jacoblee93, line 73: These were missing, it still worked but this should help with typing

js/src/utils/jestlike/index.ts
```diff
@@ -198,21 +199,33 @@ export function generateWrapperFromJestlikeMethods(
...
+        : undefined;
+      const { dataset_split, ...restMetadata } = example.metadata ?? {};
       if (
         _objectHash(example.inputs) !== _objectHash(inputs) ||
         _objectHash(example.outputs ?? {}) !== _objectHash(outputs ?? {}) ||
-        example.dataset_id !== datasetId
+        example.dataset_id !== datasetId ||
+        (normalizedSplit !== undefined &&
+          _objectHash(dataset_split ?? []) !==
+            _objectHash(normalizedSplit ?? [])) ||
```
madams0013, line 221: Not super clear on what _objectHash does but does the example have to match the specified splits or just include one of them or what

jacoblee93: I think we want to update if there's any difference at all

madams0013: ohhh I see i was misinterpreting, I understand now

madams0013 (approved):

PR #1979, opened 2025-08-28
fix: memory leak risk from otel exporter persistance

EugeneJinXin:
Problem: 
Memory leak when otel exporter process traces that never ended. Specifically,  the OTEL exporter stores spans for correlation between post (create) and patch (update) operations, but spans are only cleaned up when patch operations contain end_time. Those span that doesn't have end_time stays there forever now  

Solution:
Adds a proactive span cleanup:
- TTL-based cleanup: Remove spans older than 2 hours
- Count-based limits: Cap maximum spans at 100,000 
- Proactive processing: Clean up before each batch to prevent accumulation

Want to see if this remediates problem for customer OOM, if it works we can follow up on further fine tune the clean up logic such as make max_count and ttl configurable  

Test: 
- test hybrid and otel-only mode still sends trace to right party(s) 
- asked claude to help stress testing to confirm this seems to bound the memory usage with 10k incomplete spans 

  1,000 Incomplete Spans:

  | Metric         | Before Fix | After Fix     | Improvement          |
  |----------------|------------|---------------|----------------------|
  | Hybrid Mode    | 21.5 MB    | 26.5 MB       | Bounded (no runaway) |
  | OTEL-ONLY Mode | 4.5 MB     | 5.7 MB        | Bounded (no runaway) |
  | Active Spans   | Unlimited  | Exactly 1,000 | ✅ Controlled         |

  10,000 Incomplete Spans (Stress Test):

  | Metric         | Before Fix (Projected) | After Fix      | Improvement  |
  |----------------|------------------------|----------------|--------------|
  | Hybrid Mode    | ~215 MB → ∞            | 136.8 MB       | ✅ Bounded    |
  | OTEL-ONLY Mode | ~45 MB → ∞             | 23.4 MB        | ✅ Bounded    |
  | Active Spans   | Unlimited growth       | Exactly 10,000 | ✅ Controlled |

### DIFF
--- a/python/langsmith/_internal/otel/_otel_exporter.py
+++ b/python/langsmith/_internal/otel/_otel_exporter.py
@@ -130,14 +130,14 @@ class OTELExporter:
     ]
     """OpenTelemetry exporter for LangSmith runs."""
 
-    def __init__(self, tracer_provider=None, max_spans=100000, span_ttl_seconds=3600):
+    def __init__(self, tracer_provider=None, max_spans=100000, span_ttl_seconds=5400):
         """Initialize the OTEL exporter.
 
         Args:
             tracer_provider: Optional tracer provider to use. If not provided,
                 the global tracer provider will be used.
             max_spans: Maximum number of spans to keep in memory (default: 100000)
-            span_ttl_seconds:TTL for incomplete traces in seconds (default: 7200s)
+            span_ttl_seconds:TTL for incomplete traces in seconds (default: 5400s)
         """
         otel_imports = _import_otel_exporter()
         if otel_imports is None:
--- a/python/langsmith/_internal/otel/_otel_exporter.py
+++ b/python/langsmith/_internal/otel/_otel_exporter.py
@@ -130,14 +130,14 @@ class OTELExporter:
     ]
     """OpenTelemetry exporter for LangSmith runs."""
 
-    def __init__(self, tracer_provider=None, max_spans=100000, span_ttl_seconds=5400):
+    def __init__(self, tracer_provider=None, max_spans=100000, span_ttl_seconds=3600):
         """Initialize the OTEL exporter.
 
         Args:
             tracer_provider: Optional tracer provider to use. If not provided,
                 the global tracer provider will be used.
             max_spans: Maximum number of spans to keep in memory (default: 100000)
-            span_ttl_seconds:TTL for incomplete traces in seconds (default: 5400s)
+            span_ttl_seconds:TTL for incomplete traces in seconds (default: 3600s)
         """
         otel_imports = _import_otel_exporter()
         if otel_imports is None:
--- a/python/langsmith/_internal/otel/_otel_exporter.py
+++ b/python/langsmith/_internal/otel/_otel_exporter.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import datetime
 import logging
+import time
 import uuid
 import warnings
 from typing import TYPE_CHECKING, Any, Optional
@@ -118,22 +119,35 @@ def _get_operation_name(run_type: str) -> str:
 
 
 class OTELExporter:
-    __slots__ = ["_tracer", "_spans", "_otel_available", "_trace"]
+    __slots__ = [
+        "_tracer",
+        "_spans",
+        "_span_timestamps",
+        "_otel_available",
+        "_trace",
+        "_max_spans",
+        "_span_ttl_seconds",
+    ]
     """OpenTelemetry exporter for LangSmith runs."""
 
-    def __init__(self, tracer_provider=None):
+    def __init__(self, tracer_provider=None, max_spans=100000, span_ttl_seconds=3600):
         """Initialize the OTEL exporter.
 
         Args:
             tracer_provider: Optional tracer provider to use. If not provided,
                 the global tracer provider will be used.
+            max_spans: Maximum number of spans to keep in memory (default: 100000)
+            span_ttl_seconds:TTL for incomplete traces in seconds (default: 7200s)
         """
         otel_imports = _import_otel_exporter()
         if otel_imports is None:
             self._tracer = None
             self._spans = {}
+            self._span_timestamps = {}
             self._otel_available = False
             self._trace = None
+            self._max_spans = max_spans
+            self._span_ttl_seconds = span_ttl_seconds
         else:
             (
                 trace,
@@ -150,8 +164,11 @@ class OTELExporter:
                 "langsmith", tracer_provider=tracer_provider
             )
             self._spans = {}
+            self._span_timestamps = {}
             self._otel_available = True
             self._trace = trace
+            self._max_spans = max_spans
+            self._span_ttl_seconds = span_ttl_seconds
 
     def export_batch(
         self,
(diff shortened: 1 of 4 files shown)


python/langsmith/client.py
```diff
@@ -559,7 +559,7 @@ def __init__(
         )
         # Initialize workspace attribute first
         self._workspace_id = ls_utils.get_workspace_id(workspace_id)
-
```
jacoblee93: Format

python/langsmith/_internal/otel/_otel_exporter.py
```diff
@@ -118,22 +119,35 @@ def _get_operation_name(run_type: str) -> str:
...
+        "_span_timestamps",
+        "_otel_available",
+        "_trace",
+        "_max_spans",
+        "_span_ttl_seconds",
+    ]
     """OpenTelemetry exporter for LangSmith runs."""
 
-    def __init__(self, tracer_provider=None):
+    def __init__(self, tracer_provider=None, max_spans=100000, span_ttl_seconds=3600):
```
jacoblee93: Should make these kwargs?

python/langsmith/_internal/otel/_otel_exporter.py
```diff
@@ -176,6 +194,7 @@ def export_batch(
                     )
                     if span:
                         self._spans[op.id] = span
+                        self._span_timestamps[op.id] = time.time()
```
jacoblee93: Can we just have these share a single dict for simplicity?

`self._span_info = { "span": span, "created_at": time.time() }`

python/langsmith/_internal/otel/_otel_exporter.py
```diff
@@ -332,12 +351,64 @@ def _update_span_for_run(self, op: SerializedRunOperation, run_info: dict) -> No
                     span.end(end_time=end_time_utc_nano)
                 else:
                     span.end()
-                # Remove the span from our dictionary
+                # Remove the span and timestamp from our dictionaries
                 del self._spans[op.id]
```
jacoblee93: Read an article this morning about how `delete` in JS is slow, not sure about Python

Maybe good to combine and have both use one method?

python/langsmith/_internal/otel/_otel_exporter.py
```diff
@@ -163,7 +180,8 @@ def export_batch(
         Args:
             operations: List of serialized run operations to export.
         """
-        # Create a dictionary mapping run IDs to their operations
+        # Proactive cleanup of expired and excess spans before new operations
+        self._cleanup_stale_spans()
```
jacoblee93, line 187: This seems to iterate over all spans in memory, not sure we want to do that on every batch?

Could we somehow track the size of all in-memory spans and cleanup if it goes above a certain point?

python/langsmith/_internal/otel/_otel_exporter.py
```diff
@@ -332,12 +351,64 @@ def _update_span_for_run(self, op: SerializedRunOperation, run_info: dict) -> No
...
+    def _remove_span(self, span_id: uuid.UUID) -> None:
+        """Remove a single span and clean up resources."""
+        if span_id not in self._spans:
+            return
+
+        try:
+            # End the span gracefully
+            span = self._spans[span_id]
+            if hasattr(span, "end"):
+                span.end()
```
jacoblee93, line 410: Why `end()` it? Is that the right thing to do?

python/langsmith/_internal/otel/_otel_exporter.py
```diff
@@ -332,12 +351,64 @@ def _update_span_for_run(self, op: SerializedRunOperation, run_info: dict) -> No
...
+
+        # If still over limit, remove oldest spans
+        if len(self._spans) - len(spans_to_remove) > self._max_spans:
+            # Get remaining spans sorted by age (oldest first)
+            remaining_spans = [
+                (span_id, ts)
+                for span_id, ts in self._span_timestamps.items()
+                if span_id not in spans_to_remove
+            ]
+            remaining_spans.sort(key=lambda x: x[1])
```
jacoblee93: All this seems rather heavy - should we instead just remove everything older than some threshold?

python/langsmith/client.py
EugeneJinXin: this is actually done by ruff when `make format `, ah strange

jacoblee93: I think something is wrong with our CI not validating this, will fix

python/langsmith/_internal/otel/_otel_exporter.py
EugeneJinXin: Think it might be fine 

**Traditional OTEL:** 

 Application Code → span.end(), span completed and immutable → SpanProcessor → Exporter → Backend
                                                                 
**LangSmith:** 

otel_exporter.py creates AND manages spans

```
def _create_span_for_run(self, op, run_info, otel_context):
    span = self._tracer.start_span(...)  # ← Creates span
    return span

def _update_span_for_run(self, op, run_info):
    if end_time:
        span.end(end_time=...)  # ← Ends span when data complete
```

EugeneJinXin: LangSmith otel is such an unique architecture:  
- otel_exporter is creating spans from serialized operations (not traditional app spans)
- It owns the entire span lifecycle from creation to completion
- Acting as both span creator and span processor
- Calling span.end() is our responsibility since we created the spans; unlike otel which usually context manager such as trace.start_span() handles it

python/langsmith/_internal/otel/_otel_exporter.py
EugeneJinXin: do you mean key-word only? that sounds a good idea

python/langsmith/_internal/otel/_otel_exporter.py
EugeneJinXin: made clean up logic only run per 10 seconds to reduce overhead; 
and simplified clean up to be ttl-based only rather than ttl+volume which seems not necessary at this moment

EugeneJinXin:
made ttl configurable as well

python/langsmith/_internal/otel/_otel_exporter.py
EugeneJinXin: i found size-based a bit hard to judge, because spans have both good spans (just waiting for their patch operation) and zombie, so in high-traffic it's hard to say when to trigger it. 
Think let's just ship something useful enough for customer to test and tune from here.

python/langsmith/_internal/otel/_otel_exporter.py
```diff
@@ -257,11 +281,15 @@ def _create_span_for_run(
 
             # Start the span with appropriate context
             parent_run_id = run_info.get("parent_run_id")
-            if parent_run_id is not None and uuid.UUID(parent_run_id) in self._spans:
+            if (
+                parent_run_id is not None
+                and uuid.UUID(parent_run_id) in self._span_info
+            ):
                 # Use the parent span context
+                parent_span = self._span_info[uuid.UUID(parent_run_id)]["span"]
```
jacoblee93, line 289: Should we be checking `self._span_info[uuid.UUID(parent_run_id)]` is `None`?

jacoblee93 (approved):
Would be nice to have a test or two

python/langsmith/_internal/otel/_otel_exporter.py
EugeneJinXin: probably not? think line 286 `and uuid.UUID(parent_run_id) in self._span_info` will make sure it won't enter the if statement

PR #1980, opened 2025-08-28
fix(js): Remove validation of usage metadata params

jacoblee93:
We no longer reject runs with invalid usage metadata fields on the backend, shouldn't do this on the client either.

### DIFF
--- a/js/src/tests/traceable.test.ts
+++ b/js/src/tests/traceable.test.ts
@@ -2231,3 +2231,51 @@ test("passing null doesn't throw an error", async () => {
   expect(tree.data["i_traceable:0"].inputs).toEqual({ inputs: null });
   expect(tree.data["i_traceable:0"].outputs).toEqual({ outputs: null });
 });
+
+test("traceable with invalid properties in usage metadata", async () => {
+  const { client, callSpy } = mockClient();
+  const traceableLLM = traceable(
+    (_input: Record<string, unknown>) => {
+      return {
+        messages: [
+          {
+            role: "assistant",
+            content: "Hello, world!",
+          },
+        ],
+        usage_metadata: {
+          foo: "bar",
+          input_tokens: 10,
+          output_tokens: 10,
+          total_tokens: 20,
+        },
+      };
+    },
+    {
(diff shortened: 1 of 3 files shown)


js/src/tests/utils/tree.ts
```diff
@@ -40,16 +40,23 @@ export function getAssumedTreeFromCalls(calls: unknown[][]) {
       }
     }
 
-    if (req === "POST /runs") {
+    if (req === "POST /runs" || req === "POST /api/v1/runs") {
```
EugeneJinXin, line 43: ooc why we only adding this now; i thought v1/runs has been a thing for long time (it talks to platform backend right?

jacoblee93: This is a test utility, so doesn't affect prod

Unit tests were always failing if you have localhost as your API url it seems

js/src/tests/traceable.test.ts
```diff
@@ -2231,3 +2231,51 @@ test("passing null doesn't throw an error", async () => {
...
+    (_input: Record<string, unknown>) => {
+      return {
+        messages: [
+          {
+            role: "assistant",
+            content: "Hello, world!",
+          },
+        ],
+        usage_metadata: {
+          foo: "bar",
```
EugeneJinXin, line 2247: oh wow, why no longer check this in BE+SDK; is this to be future-proof since it keeps changing or we had other reason?

EugeneJinXin: like allow customer to specify their own things

jacoblee93: A while back we would silently drop runs with bad values here

This seems to have been changed (was bad behavior anyway to not tolerate extra keys here)

jacoblee93: And anyway, in Python we just pass it directly through:

https://github.com/langchain-ai/langsmith-sdk/blob/main/python/langsmith/run_helpers.py#L290

EugeneJinXin: cool thanks

PR #1985, opened 2025-08-29
feat(py): Allow overriding max batch size parameter on client, fix for disabled run compression

jacoblee93:
If you are tracing many runs on multiple threads, due to threading batches can exceed stated size limits

This is more complex to fix when run compression is enabled due to incremental compression, but this PR allows for restricting this when run compression is disabled


EugeneJinXin:
is this ready to review?

jacoblee93:
No will ping

jacoblee93:
Good to go now

EugeneJinXin:
discussed offline, benchling will disable compression to try this, and we'll do the follow up internally to enable this for compression depends on how testing with customer goes

jacoblee93:
CC @angus-langchain

PR #1997, opened 2025-09-03
release(js): 0.3.67

jacoblee93:


jacoblee93:
> do you want bump python as well?

Not as pressing

PR #1998, opened 2025-09-04
fix: storage of already ended spans causes memory growth

EugeneJinXin:
fix long standing memory issue 

- Storage of Already-Ended Spans: _create_span_for_run() would create and immediately end spans (for fast operations with end_time), but export_batch() would still store these ended spans in _span_info, where they would never be cleaned up. Found this when i manually log every added / completed span and found that pretty much most of the fast operations are not cleaned up. 

Fix:  prevent storage of ended spans by checking if it's still active

### DIFF
--- a/python/langsmith/_internal/otel/_otel_exporter.py
+++ b/python/langsmith/_internal/otel/_otel_exporter.py
@@ -207,9 +207,7 @@ class OTELExporter:
                             )
                         else:
                             # Span ended in _create_span_for_run (had end_time)
-                            logger.debug(
-                                "completed span (not tracked - already ended)"
-                            )
+                            logger.debug("completed span (not tracked - already ended)")
                 else:
                     self._update_span_for_run(op, run_info)
             except Exception as e:
--- a/python/langsmith/_internal/otel/_otel_exporter.py
+++ b/python/langsmith/_internal/otel/_otel_exporter.py
@@ -203,12 +203,12 @@ class OTELExporter:
                                 "created_at": time.time(),
                             }
                             logger.debug(
-                                f"Created active span, total spans: {len(self._span_info)}"
+                                f"Created active span, total: {len(self._span_info)}"
                             )
                         else:
                             # Span was already ended in _create_span_for_run (had end_time)
                             logger.debug(
-                                "Created and completed span (not tracked - already ended)"
+                                "completed span (not tracked - already ended)"
                             )
                 else:
                     self._update_span_for_run(op, run_info)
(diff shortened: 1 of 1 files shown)


python/langsmith/_internal/otel/_otel_exporter.py
```diff
@@ -196,10 +197,20 @@ def export_batch(
...
-                            "span": span,
-                            "created_at": time.time(),
-                        }
+                        # Only store spans that are still active (not already ended)
+                        if hasattr(span, "is_recording") and span.is_recording():
+                            self._span_info[op.id] = {
+                                "span": span,
+                                "created_at": time.time(),
+                            }
+                            logger.debug(
```
angus-langchain, line 205: do we want to keep the debug logging?

python/langsmith/_internal/otel/_otel_exporter.py
```diff
@@ -388,10 +403,12 @@ def _cleanup_stale_spans(self) -> None:
         ]
 
         if stale_span_ids:
-            logger.debug(f"Cleaning up {len(stale_span_ids)} stale spans")
+            logger.info(
```
angus-langchain, line 403: This comment feel like AI, can we remove?

python/langsmith/_internal/otel/_otel_exporter.py
```diff
@@ -406,14 +423,29 @@ def _remove_span(self, span_id: uuid.UUID) -> None:
...
+            # Check if span is still active before ending it
+            if (
+                hasattr(span, "end")
+                and hasattr(span, "is_recording")
+                and span.is_recording()
+            ):
                 span.end()
+                logger.debug(f"Ended orphaned span {span_id}")
+            elif hasattr(span, "end"):
+                # Span already ended, just log it
```
angus-langchain, line 433: how do we know the span ended here?

angus-langchain (approved):

python/langsmith/_internal/otel/_otel_exporter.py
EugeneJinXin: i'm debating, i wanted to keep in case future strange bug happen, wdyt?

python/langsmith/_internal/otel/_otel_exporter.py
EugeneJinXin: same reason, wdyt

python/langsmith/_internal/otel/_otel_exporter.py
EugeneJinXin: because is_recording() tells if an span is active

PR #2000, opened 2025-09-04
release(py): 0.4.25

EugeneJinXin:


python/langsmith/__init__.py
```diff
@@ -21,7 +21,7 @@
```
angus-langchain, line 21: you need to update uv.lock as well

PR #2001, opened 2025-09-04
otel developer improvements

angus-langchain:


hinthornw:
driveby: we already have 'langsmith.configure()'; maybe we should align naming or whatnot if we land this type of hting?

python/langsmith/integrations/otel/__init__.py
```diff
@@ -0,0 +1,110 @@
...
+from langsmith import utils as ls_utils
+
+from .processor import OtelExporter, OtelSpanProcessor
+
+logger = logging.getLogger(__name__)
+
+__all__ = ["configure", "OtelSpanProcessor", "OtelExporter"]
+
+
+def configure(
```
jacoblee93, line 15: > driveby: we already have 'langsmith.configure()'; maybe we should align naming or whatnot if we land this type of hting?

Not sure how this works (don't we already pull from env?) but should we just add this as a separate flag in that method if it already exists instead of a separately imported method?

python/langsmith/integrations/otel/__init__.py
```diff
@@ -0,0 +1,110 @@
...
+        >>> configure(  # doctest: +SKIP
+        ...     api_key="your-api-key", project_name="your-project"
+        ... )
+
+        Using environment variables:
+        >>> # Set LANGSMITH_API_KEY and LANGSMITH_PROJECT
+        >>> configure()  # Will use env vars  # doctest: +SKIP
+
+    WARNING: This function is only for when LangSmith is your ONLY OpenTelemetry source.
+    It sets the global TracerProvider, which can only be done once per application.
```
jacoblee93, line 35: Nice hope people read it

python/langsmith/integrations/otel/processor.py
```diff
@@ -0,0 +1,219 @@
...
+
+        if project:
+            exporter_headers["Langsmith-Project"] = project
+
+        self.project = project
+
+        super().__init__(endpoint=endpoint, headers=exporter_headers, **kwargs)
+
+
+class OtelSpanProcessor:
```
jacoblee93, line 124: Feel like this should be subclassing something/implementing some interface?

python/langsmith/integrations/otel/processor.py
```diff
@@ -0,0 +1,219 @@
...
+except ImportError:
+    warnings.warn(
+        "OpenTelemetry packages are not installed. "
+        "Install optional OpenTelemetry dependencies with: "
+        "pip install langsmith[otel]",
+        UserWarning,
+        stacklevel=2,
+    )
+
+    class OTLPSpanExporter:  # type: ignore[no-redef]
```
jacoblee93, line 24: Is the idea that these mocks override the below classes if the import fails?

python/langsmith/integrations/otel/processor.py
```diff
@@ -0,0 +1,219 @@
...
+        provider.add_span_processor(langsmith_processor)
+    """
+
+    def __init__(
+        self,
+        api_key: Optional[str] = None,
+        project: Optional[str] = None,
+        url: Optional[str] = None,
+        headers: Optional[dict[str, str]] = None,
+        SpanProcessor: Optional[type] = None,
```
jacoblee93, line 159: OOC why would someone want something other than `BatchSpanProcessor`? I guess if you had more than one exporter?

jacoblee93 (approved):
Think this is fine

python/langsmith/integrations/otel/processor.py
angus-langchain: yup!

python/langsmith/integrations/otel/processor.py
angus-langchain: yeah if you have more than 1 or custom exporters

python/langsmith/integrations/otel/__init__.py
angus-langchain: this feels more explicit and it in a different integrations package. I don't want to weave in otel setup with our regular tracing configs either

PR #2003, opened 2025-09-05
fix(js): Fix retries for empty bodies

jacoblee93:
`await response.text()` will throw an error if the body is empty. If this happens, it overrides the existing error and does not set a status code, causing it to be retried inappropriately.

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -848,7 +848,7 @@ export class Client implements LangSmithTracingClientInterface {
         signal: AbortSignal.timeout(this.timeout_ms),
         ...this.fetchOptions,
       });
-      await raiseForStatus(res, `Failed to fetch ${path}`);
+      await raiseForStatus(res, `fetch ${path}`);
       return res;
     });
     return response;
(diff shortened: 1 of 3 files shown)


js/src/client.ts
```diff
@@ -917,7 +917,7 @@ export class Client implements LangSmithTracingClientInterface {
           ...this.fetchOptions,
           body,
         });
-        await raiseForStatus(res, `Failed to fetch ${path}`);
+        await raiseForStatus(res, `fetch ${path}`);
```
jacoblee93, line 920: `Failed to` is already added in the `error.ts` handling

js/src/utils/error.ts
```diff
@@ -91,24 +91,37 @@ export async function raiseForStatus(
...
+      (errorWithStatus as any).status = response?.status;
+      throw errorWithStatus;
     }
   }
-  errorBody = await response.text();
+  if (errorBody === undefined) {
+    try {
+      errorBody = await response.text();
+    } catch (e: any) {
+      errorBody = "Empty";
```
EugeneJinXin: ah not familiar with JS convention, hope this is ok,

EugeneJinXin (approved):

js/src/utils/error.ts
jacoblee93: What's your concern?

EugeneJinXin: nah no concern, just have seen more "" as empty haha

jacoblee93: Ah, yeah maybe less misleading

PR #2012, opened 2025-09-11
Non-beta parse awareness in the OpenAI wrapper

### LINKED ISSUE #2011: Issue: Non-beta parse method not wrapped when using the wrap_openai function

Quite simply while the JS client seems to have the awareness of the non-beta methods to wrap (`js/src/wrappers/openai.ts:L385`) the python version of the wrapper does not have this in the current version. This just needs to be added so we don't need to force our users to have to use the beta when they should not.

Separate from #1588 as that pertains to a separate part of the API.

May be related to #1984 if they happened to not use the beta methods

### Suggestion:

_No response_

cr-klarna:
Fixes #2011
Adds a check for clients that have the non-beta parse functionality for OpenAI.
Ensures that the following code runs equivalently to the legacy `beta` methods and traces properly:

```
completion = client.chat.completions.parse(
    model="gpt-4o-2024-08-06",
    messages=[
        {"role": "system", "content": "Extract the event information."},
        {
            "role": "user",
            "content": "Alice and Bob are going to a science fair on Friday.",
        },
    ],
    response_format=CalendarEvent,
)
```

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "langsmith",
-  "version": "0.3.67",
+  "version": "0.3.68",
   "description": "Client library to connect to the LangSmith LLM Tracing and Evaluation Platform.",
   "packageManager": "yarn@1.22.19",
   "files": [
--- a/js/src/experimental/vercel/index.ts
+++ b/js/src/experimental/vercel/index.ts
@@ -500,7 +500,7 @@ const wrapAISDK = <
             }
             const { content } = lastStep;
             return convertMessageToTracedFormat({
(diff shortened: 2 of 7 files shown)


cr-klarna:
Not sure what the JS tests are failing on. Willing to take a look, but will need some extra guidance here on those errors

jacoblee93:
Hey @cr-klarna! Thanks a bunch for this. Yes, integration tests are not enabled for contributors, but I will take a look and get this merged.

jacoblee93 (approved):
Very sorry for the delay!

PR #2021, opened 2025-09-15
Get experiment

EugeneJinXin:


vendor/orjson/test/test_api.py
```diff
@@ -111,7 +111,7 @@ def test_loads_recursion_valid_limit_mixed_pretty(self):
         loads() recursion limit at limit mixed pretty
         """
         n = LOADS_RECURSION_LIMIT
-        value = b"[\n  " b'{"key":' * n + b'{"key":true}' + b"}" * n + b"]"
+        value = b'[\n  {"key":' * n + b'{"key":true}' + b"}" * n + b"]"
```
EugeneJinXin, line 114: strange this is done by formatter, will check why

python/script_get_experiment.py
jacoblee93, line 1: Remove this

python/langsmith/schemas.py
```diff
@@ -1261,3 +1261,18 @@ class UpsertExamplesResponse(TypedDict):
...
+    runs: list[Run] = Field(default_factory=list)
+
+    """The runs of the example."""
+
+
+class ExperimentResults(TypedDict):
+    """Results container for experiment data with stats and examples."""
+
+    stats: TracerSessionResult
+    examples_with_runs: Iterator[ExampleWithRuns]
```
jacoblee93, line 1278: This is kind of an awkward name

Should we try to closer match what we do in `get_test_results()`?

vendor/orjson/test/test_api.py
jacoblee93: More concerning that we don't enforce formatting in CI - thought I fixed this

python/langsmith/client.py
```diff
@@ -8269,6 +8269,132 @@ async def helpfulness(outputs: dict) -> dict:
...
+            for batch in self._paginate_dataset_runs(
+                dataset_id=dataset_id,
+                session_id=session_id,
+                preview=preview,
+                comparative_experiment_id=comparative_experiment_id,
+                filters=filters,
+                limit=limit,
+            ):
+                for result in batch:
+                    yield _extract_run_metrics_from_example(result)
```
jacoblee93: Is `result` not a run object? Where does the `example` come into play?

python/langsmith/client.py
```diff
@@ -8309,6 +8435,41 @@ def convert_prompt_to_openai_format(
         raise ls_utils.LangSmithError(f"Error converting to OpenAI format: {e}")
 
 
+def _extract_run_metrics_from_example(result: dict) -> dict:
```
jacoblee93: Naming is confusing, you're actually transforming the `dict` here instead of just returning it

Would be nicer to use real types

python/langsmith/schemas.py
EugeneJinXin: changed examples_with_runs to experiment_runs, does it sound better?  

class ExperimentResults(TypedDict):
    """Results container for experiment data with stats and examples."""

    stats: TracerSessionResult
    experiment_runs: Iterator[ExampleWithRuns]

python/langsmith/client.py
EugeneJinXin: > Is `result` not a run object? Where does the `example` come into play?

i see the confusion.. i think it's indeed more a run than example, although example came into play was due to  return type of `POST /datasets/{dataset_id}/runs ` is List[ExampleWithRuns] here is python endpoint  https://github.com/langchain-ai/langchainplus/blob/main/smith-backend/app/api/endpoints/datasets.py#L388

EugeneJinXin: renamed to `for run in batch`, you right it's more clear

python/langsmith/client.py
EugeneJinXin: removed this extraction as a whole, because i feel it's kinda not necessary to extract each run's stats from Run object, and it makes the already complex typing even more frustrating, user can do calculation after getting Runs

python/langsmith/evaluation/_arunner.py
```diff
@@ -648,7 +648,6 @@ def _reset_example_attachments(self, example: schemas.Example) -> schemas.Exampl
             outputs=example.outputs,
             metadata=example.metadata,
             modified_at=example.modified_at,
-            runs=example.runs,
```
EugeneJinXin, line 651: my PR introducing ExampleWithRuns made MyPy strict enough to catch this dormant bug,   before it could cause runtime failures. Example pydantic type does not have `runs`

python/langsmith/client.py
```diff
@@ -8269,6 +8269,130 @@ async def helpfulness(outputs: dict) -> dict:
...
+                f"/datasets/{dataset_id}/runs",
+                request_kwargs={"data": _dumps_json(body)},
+            )
+
+            batch = response.json()
+            if not batch:
+                break
+
+            # Transform raw dictionaries to ExampleWithRuns instances
+            examples_batch = [ls_schemas.ExampleWithRuns(**result) for result in batch]
```
jacoblee93, line 8321: What's the max number of runs this returns per example?

python/langsmith/client.py
```diff
@@ -8269,6 +8269,130 @@ async def helpfulness(outputs: dict) -> dict:
             **kwargs,
         )
 
+    def _paginate_dataset_runs(
```
jacoblee93: Call this `_paginate_examples_with_runs`

python/langsmith/client.py
```diff
@@ -8269,6 +8269,130 @@ async def helpfulness(outputs: dict) -> dict:
...
+                session_id=project_id,
+                preview=preview,
+                comparative_experiment_id=comparative_experiment_id,
+                filters=filters,
+                limit=limit,
+            ):
+                yield from batch
+
+        return ls_schemas.ExperimentResults(
+            stats=project_stats[0], experiment_runs=_get_dataset_runs_iterator()
```
jacoblee93: Sorry I think actually `examples_with_runs` is nicer - if we do this it'll be:

```py
results.experiment_runs[0].runs` which is kind of silly
```

python/langsmith/evaluation/_arunner.py
jacoblee93: Cool, definitely not breaking since this would error right?

jacoblee93 (commented):
Nice looks pretty good

python/langsmith/client.py
EugeneJinXin: That's good thought, will adopt

python/langsmith/evaluation/_arunner.py
EugeneJinXin: Yes this is safe;  this code path seems have never worked in runtime, so this is actually a fix

python/langsmith/client.py
EugeneJinXin: each example has at max (The Number of repetition) runs.

Kinda related,  we not allowing multiple experiment sessions to be query in this method because that is both slow and could make Stats a bit confusing, user can do call this method multi times with different experiments if they want

jacoblee93: Unlikely to be an issue but are those paginated too?

EugeneJinXin: when calling the endpoint, we do provide offset+limit, so yes, unless i misunderstood your question [code link ](https://github.com/langchain-ai/langsmith-sdk/pull/2021/files#diff-aa866ec6b1a81eca98b083280a30d539b2c154d4c62bd58fdfc09e06714a873bR8303)

PR #2022, opened 2025-09-16
feat(py,js): Adds list_experiments method

jacoblee93:


EugeneJinXin:
this python side looks good actually, let me know when it's ready for review

PR #2025, opened 2025-09-16
feat(py): Adds polling to blocking experiments

jacoblee93:


python/langsmith/evaluation/_arunner.py
```diff
@@ -1230,8 +1230,61 @@ def __len__(self) -> int:
...
         return f"<AsyncExperimentResults {self.experiment_name}>"
 
-    async def wait(self) -> None:
+    async def wait(self, timeout: float = 30.0) -> None:
+        """Wait for the experiment to complete and for runs to be available.
+
+        Args:
+            timeout: Maximum time to wait in seconds (default: 30.0)
+        """
+        import time
```
hinthornw: don't import in functions

PR #2027, opened 2025-09-17
fix(js): Fix output processing for AI SDK streamText

### LINKED ISSUE #2023: Issue: [JS SDK] wrapper for Vercel SDK (wrapAISDK) causing uncaughtException when aborting request to LLM

When using JS SDK,
When wrapping `streamText` method from Vercel's AI SDK and aborting request to LLM at some point, this produces `uncaughtException` in process.

code example:
```
import { anthropic } from '@ai-sdk/anthropic';
import * as ai from 'ai';
import 'dotenv/config';
import { wrapAISDK } from 'langsmith/experimental/vercel';

const wrap = true;

const { streamText } = wrap ? wrapAISDK(ai) : ai;

let res: ai.StreamTextResult<ai.ToolSet, unknown>;

void (async () => {
    const abortController = new AbortController();

    try {
        res = streamText({
            model: anthropic('claude-sonnet-4-20250514'),
            messages: [
                {
                    role: 'user',
                    content: [
                        {
                            type: 'text',
                            text: 'tell me lorem ipsum poem',
                        },
                    ],
                },
            ],
            abortSignal: abortController.signal,
        });

        let i = 0;
        for await (const p of res.fullStream) {
            if (p.type === 'text-delta') {
                console.log(p.text);

                if (i++ > 10) {
                    abortController.abort();
                    console.log('aborted');
                }
            }
        }
    } catch {
        console.log('catch try/catch block');
    }
})()
    .catch((e) => {
        console.log(`catch async function`);
    })
    .then(() => {
        console.log('then async function');
    });

while (true) {
    console.log('running...');
    await new Promise((r) => setTimeout(r, 1000));
}

```

please note, using `streamText` without wrapping works good and process continue running after cancelling request.

### Suggestion:

I suppose the root cause of this issue is this line of code: https://github.com/langchain-ai/langsmith-sdk/blob/main/js/src/experimental/vercel/index.ts#L661-L663

```
          if ("content" in outputs.outputs && outputs.outputs.content != null) {
            content = await outputs.outputs.content;
          } else if (
```

it accesses content property of output to check if it is not null. That property is implemented as chain of promises: https://github.com/vercel/ai/blob/main/packages/ai/src/generate-text/stream-text.ts#L1543-L1557

```
  get steps() {
    // when any of the promises are accessed, the stream is consumed
    // so it resolves without needing to consume the stream separately
    this.consumeStream();

    return this._steps.promise;
  }

  private get finalStep() {
    return this.steps.then(steps => steps[steps.length - 1]);
  }

  get content() {
    return this.finalStep.then(step => step.content);
  }
```

This is not obvious, but checking if that property is not equal to null initializes a new promise that lacks catch block. Thus, once request is cancelled it fails whole process with uncaughtError caused by that implicit promise

jacoblee93:
Fixes #2023

### DIFF
--- a/.github/workflows/js_test.yml
+++ b/.github/workflows/js_test.yml
@@ -110,5 +110,27 @@ jobs:
         run: yarn run check-version
       - name: Test
         run: yarn run test
+
+
+  vitest:
+    name: Vitest
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: "js"
+    steps:
+      - uses: actions/checkout@v4
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: "yarn"
+          cache-dependency-path: "js/yarn.lock"
+      - name: Install dependencies
+        run: yarn install --immutable
+      - name: Build
+        run: yarn run build
+      - name: Check version
+        run: yarn run check-version
       - name: Test: Vitest
         run: yarn run test:vitest
--- a/.github/workflows/js_test.yml
+++ b/.github/workflows/js_test.yml
@@ -110,3 +110,5 @@ jobs:
         run: yarn run check-version
       - name: Test
         run: yarn run test
+      - name: Test: Vitest
+        run: yarn run test:vitest
--- a/.github/workflows/js_test.yml
+++ b/.github/workflows/js_test.yml
@@ -7,9 +7,8 @@ permissions:
 
 on:
   pull_request:
-    paths:
-      - ".github/**"
-      - "js/**"
+    branches:
+      - main
   workflow_dispatch:
     inputs:
       run_integration_tests:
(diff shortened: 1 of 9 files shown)


.github/workflows/js_test.yml
```diff
@@ -110,3 +110,5 @@ jobs:
         run: yarn run check-version
       - name: Test
         run: yarn run test
+      - name: Test: Vitest
+        run: yarn run test:vitest
```
jacoblee93, line 116: AI SDK's mock providers don't seem to support Jest anymore

We should honestly get everything over to Vitest at some point anyway, but doing this as a stopgap

js/package.json
```diff
@@ -178,6 +179,7 @@
     "eslint-plugin-prettier": "^4.2.1",
     "jest": "^29.5.0",
     "langchain": "^0.3.29",
+    "msw": "^2.11.2",
```
jacoblee93, line 182: AI SDK's mock provider seems to need this mock in new versions

PR #2047, opened 2025-10-02
feat: add batching to create_examples

EugeneJinXin:
LS-4410 
Enhance upload performance and reduce chances of hitting the current 25MB max payload limit on our backend. 
Configurable batch_size to allow user adjust based on example payload size

### DIFF
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -4854,6 +4854,7 @@ class Client:
         dataset_id: Optional[ID_TYPE] = None,
         examples: Optional[Sequence[ls_schemas.ExampleCreate | dict]] = None,
         dangerously_allow_filesystem: bool = False,
+        batch_size: int = 200,
         **kwargs: Any,
     ) -> ls_schemas.UpsertExamplesResponse | dict[str, Any]:
         """Create examples in a dataset.
@@ -4869,6 +4870,8 @@ class Client:
                 The examples to create.
             dangerously_allow_filesystem (bool):
                 Whether to allow uploading files from the filesystem.
+            batch_size (int):
+                upload examples in batches of this size, default is 200.
             **kwargs (Any): Legacy keyword args. Should not be specified if 'examples' is specified.
 
                 - inputs (Sequence[Mapping[str, Any]]): The input values for the examples.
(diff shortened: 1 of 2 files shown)


python/langsmith/client.py
```diff
@@ -4854,6 +4854,7 @@ def create_examples(
         dataset_id: Optional[ID_TYPE] = None,
         examples: Optional[Sequence[ls_schemas.ExampleCreate | dict]] = None,
         dangerously_allow_filesystem: bool = False,
+        batch_size: int = 200,
```
jacoblee93: Think we need to consider max size in bytes too?

.mlc_config.json
```diff
@@ -1,3 +1,8 @@
 {
-  "aliveStatusCodes": [429, 200]
+  "aliveStatusCodes": [429, 200],
+  "ignorePatterns": [
+    {
+      "pattern": "^https://www\\.npmjs\\.com/"
```
EugeneJinXin, line 5: just to fix ci/cd failure that complains langsmith npm link returns 403, not our fault

python/langsmith/client.py
```diff
@@ -4930,6 +5004,9 @@ def create_examples(
                 response = client.create_examples(dataset_name="agent-qa", examples=examples)
                 # -> {"example_ids": [...
         """  # noqa: E501
+        if not 1 <= max_concurrency <= 3:
```
EugeneJinXin, line 5028: think @validate_call pydantic runtime is ugly, so i prefer this since it works as simple and effective runtime check

python/langsmith/client.py
```diff
@@ -4994,35 +5071,109 @@ def create_examples(
...
 
+        if not uploads:
+            return ls_schemas.UpsertExamplesResponse(example_ids=[], count=0)
+
+        # Use size-aware batching to prevent payload limit errors
+        batches = self._batch_examples_by_size(uploads)
+
+        # sequential upload
+        if len(batches) <= 1 or max_concurrency <= 1:
+            return self._upload_examples_batches_sequential(
```
jacoblee93: Why can't this use the same method as parallel?

EugeneJinXin: good idea, addressed

jacoblee93 (approved):

PR #2049, opened 2025-10-02
fix(py): return session feedback with experiment results

baskaryan:


EugeneJinXin (approved):
LGTM and tested with below script, works OK, feel free to ship
```

from pprint import pprint

res = client.get_experiment_results(name="My Experiment-9dbd1c61")
pprint(res)
pprint(res["feedback_stats"])
pprint(res["run_stats"])

 
for i, res in enumerate(res["examples_with_runs"]):
    if i == 0:
        pprint(res.dict())
```

PR #2057, opened 2025-10-06
fix ugly rendering of 'omit' in UI due to incorrect strip

EugeneJinXin:
## Problem

  LangSmith UI was showing ugly serialized objects like:
  "seed": "<openai.Omit object at 0x105d91940>",
  "stop": "<openai.Omit object at 0x105d91940>",

image

  This happened because OpenAI SDK v1.106+ uses both NotGiven and
  Omit sentinel types to represent optional parameters that weren't
  provided, but the code only checked for NotGiven.

after fix: 
image

### DIFF
--- a/python/tests/integration_tests/wrappers/test_azure_openai.py
+++ b/python/tests/integration_tests/wrappers/test_azure_openai.py
@@ -34,14 +34,14 @@ def test_chat_sync_api(stream: bool):
     )
     messages = [{"role": "user", "content": "Say 'foo'"}]
     original = original_client.chat.completions.create(
-        messages=messages,  # noqa: [arg-type]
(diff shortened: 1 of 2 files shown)


python/langsmith/wrappers/_openai.py
```diff
@@ -35,24 +35,33 @@
...
 
-        return NotGiven
+        types.append(NotGiven)
     except ImportError:
-        return None
+        pass
+    try:
+        from openai._types import Omit
+
+        types.append(Omit)
```
jacoblee93, line 45: nit: just combine this with the above import so you only need one try/catch

jacoblee93 (approved):

PR #2058, opened 2025-10-06
feat: add log of base_url

EugeneJinXin:
Worth adding this openai.base_url since it's important metadata, helping usage analytics of openAI format-compliant providers such as gemini   

idea credits to jake https://langchain.slack.com/archives/C08RJTF3QS2/p1758664251472349?thread_ts=1758663268.667749&cid=C08RJTF3QS2

### DIFF
--- a/js/src/wrappers/openai.ts
+++ b/js/src/wrappers/openai.ts
@@ -346,6 +346,7 @@ export const wrapOpenAI = <T extends OpenAIType>(
           params.max_completion_tokens ?? params.max_tokens ?? undefined,
         ls_temperature: params.temperature ?? undefined,
         ls_stop,
+        openai_base_url: (openai as any).baseURL,
       };
     },
     processOutputs: processChatCompletion,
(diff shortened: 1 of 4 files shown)


js/src/wrappers/openai.ts
```diff
@@ -346,6 +346,7 @@ export const wrapOpenAI = <T extends OpenAIType>(
           params.max_completion_tokens ?? params.max_tokens ?? undefined,
         ls_temperature: params.temperature ?? undefined,
         ls_stop,
+        openai_base_url: (openai as any).baseURL,
```
jacoblee93: Can we just pull this in from the parameters the model is called with? If `baseURL` isn't in the types, it could potentially be removed in a patch version change

EugeneJinXin: good catch, yeah it'd be safer to capture baseURL once at wrap time and spread it defensively, addressed in latest commit

Issue #1722, opened 2025-05-09
Native Wrapper for Google Gemini (@google/genai) in LangSmith JS SDK


reedchan7:
### Feature request

Hi LangSmith Team,

I'm currently using the `@google/genai` client for interacting with Google's Gemini models and would like to integrate these calls with LangSmith for tracing.

I've noticed that the langsmith TypeScript SDK provides a convenient `wrapOpenAI` function for OpenAI clients, which makes integration straightforward by automatically handling much of the tracing setup. This is a fantastic feature!

However, I couldn't find a similar dedicated wrapper for the `@google/genai` client. While I understand the generic traceable function can be used to wrap any asynchronous function, a dedicated wrapper (e.g., `wrapGemini` or `wrapGoogleGenAI`) would offer a more streamlined and consistent developer experience, similar to what's available for OpenAI.

### Motivation

With the increasing adoption and capabilities of Google's Gemini models, providing first-class, easy-to-use tracing support within LangSmith is crucial for developers who choose Gemini as their LLM provider. A dedicated wrapper would significantly lower the barrier to entry for integrating Gemini with LangSmith, encouraging broader adoption of LangSmith within the Gemini ecosystem and ensuring developers can easily monitor, debug, and evaluate their Gemini-powered applications. This parity with the OpenAI integration would make LangSmith a more versatile and appealing observability tool for a wider range of LLM users.

Eunchan24:
Is developing these features progressing?

PR #2059, opened 2025-10-07
feat: wrap_gemini beta python

EugeneJinXin:
wrap_gemini has been requested quite many times (23 [upvotes](  https://github.com/langchain-ai/langsmith-sdk/issues/1722) from sdk issues); we wanted to make it in September  

Tests: integration test (still need to add google_api_key in github settings) and inspection in langsmith https://smith.langchain.com/public/d3958002-41cd-4df2-9711-91689ec3ea88/r 

Note: mark as beta because even though the messages goes up to langsmith, i expect there'll be edges cases on rendering we need to take care of; don't really expect much breaking changes but wanna be proactive in setting expectation

image

### DIFF
--- a/python/langsmith/wrappers/_gemini.py
+++ b/python/langsmith/wrappers/_gemini.py
@@ -342,7 +342,9 @@ def _process_generate_content_response(response: Any) -> dict:
 
         # Extract and convert usage metadata
         usage_metadata = rdict.get("usage_metadata")
-        usage_dict = {}
+        usage_dict: UsageMetadata = UsageMetadata(
+            input_tokens=0, output_tokens=0, total_tokens=0
+        )
         if usage_metadata:
             usage_dict = _create_usage_metadata(usage_metadata)
             # Add usage_metadata to both run.extra AND outputs
@@ -378,8 +380,12 @@ def _process_generate_content_response(response: Any) -> dict:
                 }
             else:
                 # For multimodal content, return structured format with usage metadata
-                result["usage_metadata"] = usage_dict
-                return result
+                return {
+                    "content": result["content"],
+                    "role": "assistant",
+                    "finish_reason": finish_reason,
+                    "usage_metadata": usage_dict,
+                }
     except Exception as e:
         logger.debug(f"Error processing Gemini response: {e}")
         return {"output": response}
@@ -388,7 +394,12 @@ def _process_generate_content_response(response: Any) -> dict:
 def _reduce_generate_content_chunks(all_chunks: list) -> dict:
     """Reduce streaming chunks into a single response."""
     if not all_chunks:
-        return {"content": "", "usage_metadata": {}}
+        return {
+            "content": "",
+            "usage_metadata": UsageMetadata(
+                input_tokens=0, output_tokens=0, total_tokens=0
+            ),
+        }
 
     # Accumulate text from all chunks
     full_text = ""
@@ -402,11 +413,12 @@ def _reduce_generate_content_chunks(all_chunks: list) -> dict:
         except Exception as e:
             logger.debug(f"Error processing chunk: {e}")
 
-    # Build chat-like response format
-    result = {"content": full_text, "role": "assistant"}
+    # Extract usage metadata from the last chunk
+    usage_metadata: UsageMetadata = UsageMetadata(
+        input_tokens=0, output_tokens=0, total_tokens=0
+    )
     if last_chunk:
         try:
-            # Extract usage metadata from the last chunk
             if hasattr(last_chunk, "usage_metadata") and last_chunk.usage_metadata:
                 if hasattr(last_chunk.usage_metadata, "to_dict"):
                     usage_dict = last_chunk.usage_metadata.to_dict()
@@ -432,7 +444,6 @@ def _reduce_generate_content_chunks(all_chunks: list) -> dict:
                     }
                 # Add usage_metadata to both run.extra AND outputs
                 usage_metadata = _create_usage_metadata(usage_dict)
-                result["usage_metadata"] = usage_metadata
                 current_run = run_helpers.get_current_run_tree()
                 if current_run:
                     try:
@@ -448,8 +459,8 @@ def _reduce_generate_content_chunks(all_chunks: list) -> dict:
 
     # Return minimal structure with usage_metadata in outputs
     return {
-        "content": result["content"],
-        "usage_metadata": result.get("usage_metadata", {}),
+        "content": full_text,
+        "usage_metadata": usage_metadata,
     }
 
 


python/langsmith/wrappers/_gemini.py
```diff
@@ -0,0 +1,426 @@
...
+            Defaults to "ChatGoogleGenerativeAI".
+
+    Returns:
+        genai.Client: The patched client.
+
+    Example:
+
+        .. code-block:: python
+
+            from google import genai
```
jacoblee93, line 558: I know they have a bunch of different clients and they were unifying them - but is it worth calling this `wrap_google_genai` instead?

Think it's also better because I'm pretty sure you can access non-Gemini models (e.g. Gemma, other Model Garden stuff)

python/langsmith/wrappers/_gemini.py
```diff
@@ -0,0 +1,426 @@
...
+            from langsmith import wrappers
+
+            # Use Google Gen AI client same as you normally would.
+            client = wrappers.wrap_gemini(genai.Client(api_key="your-api-key"))
+
+            # Non-streaming:
+            response = client.models.generate_content(
+                model="gemini-2.0-flash-001",
```
jacoblee93: nit: use latest model string

python/langsmith/wrappers/_gemini.py
```diff
@@ -0,0 +1,426 @@
...
+C = TypeVar("C", bound=Union["genai.Client", Any])
+logger = logging.getLogger(__name__)
+
+
+def _strip_none(d: dict) -> dict:
+    """Remove None values from dictionary."""
+    return {k: v for k, v in d.items() if v is not None}
+
+
+def _process_gemini_inputs(inputs: dict) -> dict:
```
jacoblee93, line 42: Any special consideration needed for tool calls?

python/tests/integration_tests/wrappers/test_gemini.py
jacoblee93, line 1: Would definitely add a tool calling test too

python/langsmith/wrappers/_gemini.py
```diff
@@ -0,0 +1,426 @@
...
+        temperature = config.temperature
+        max_tokens = getattr(config, "max_output_tokens", None)
+        stop = getattr(config, "stop_sequences", None)
+    else:
+        temperature = config.get("temperature")
+        max_tokens = config.get("max_output_tokens")
+        stop = config.get("stop_sequences")
+
+    return {
+        "ls_provider": "google",
```
jacoblee93, line 189: Worth checking to make sure this is the same as LangChain

python/tests/integration_tests/wrappers/test_gemini.py
EugeneJinXin: good pointed added too

python/langsmith/wrappers/_gemini.py
EugeneJinXin: per our discussion it's either wrap_google or wrap_gemini, do you still hold strongly for wrap_google? i think users probably won't be too surprised if they see we still name wrap_gemini, see https://github.com/langchain-ai/langsmith-sdk/issues/1722

python/langsmith/wrappers/_gemini.py
EugeneJinXin: yeap the format is very different so added extraction logic, FE still need to do a better job rendering but think we have a good starting point showing all tool calls in json and default rendering at least

python/langsmith/wrappers/_gemini.py
```diff
@@ -0,0 +1,606 @@
...
+                            last_chunk.usage_metadata, "cached_content_token_count", 0
+                        ),
+                        "thoughts_token_count": getattr(
+                            last_chunk.usage_metadata, "thoughts_token_count", 0
+                        ),
+                        "total_token_count": getattr(
+                            last_chunk.usage_metadata, "total_token_count", 0
+                        ),
+                    }
+                result["usage_metadata"] = _create_usage_metadata(usage_dict)  # type: ignore[assignment]
```
angus-langchain: if this isn't possible lmk, but we are standardizing on storing usage_metadata in extra.metadata and not outputs.

angus-langchain: same for js

python/langsmith/wrappers/_gemini.py
```diff
@@ -0,0 +1,606 @@
+from __future__ import annotations
```
angus-langchain, line 1: is it possible to send the first event to langsmith so we can calculate time to first token?

python/langsmith/wrappers/_gemini.py
```diff
@@ -0,0 +1,606 @@
...
+                                }
+                            )
+        elif "text" in rdict:
+            content_result = rdict["text"]
+            content_parts.append({"type": "text", "text": content_result})
+
+        # Build chat-like response format - use OpenAI-compatible format for tool calls
+        tool_calls = [p for p in content_parts if p.get("type") == "function_call"]
+
+        if tool_calls:
```
angus-langchain, line 307: do we need to serialize the tool calls?
```
from google.genai.models import (
          _GenerateContentParameters_to_mldev,
          _GenerateContentParameters_to_vertex,
      )
```

python/langsmith/wrappers/_gemini.py
angus-langchain: and capture finish_reason. neither blocking but nice to have

python/langsmith/wrappers/_gemini.py
```diff
@@ -0,0 +1,606 @@
...
+    if not all_chunks:
+        return {"content": "", "usage_metadata": {}}
+
+    # Accumulate text from all chunks
+    full_text = ""
+    last_chunk = None
+
+    for chunk in all_chunks:
+        try:
+            if hasattr(chunk, "text") and chunk.text:
```
angus-langchain, line 410: the chunk can also have "thought": chunk.thought

python/langsmith/wrappers/_gemini.py
```diff
@@ -0,0 +1,606 @@
...
+
+    # Handle list of content objects (multimodal case)
+    if isinstance(contents, list):
+        # Check if it's a simple list of strings
+        if all(isinstance(item, str) for item in contents):
+            # Each string becomes a separate user message (matches Gemini's behavior)
+            return {
+                "messages": [{"role": "user", "content": item} for item in contents],
+                "model": inputs.get("model"),
+                **({k: v for k, v in inputs.items() if k not in ("contents", "model")}),
```
angus-langchain, line 72: thinking you need to serialize tools - could be function objects, dicts, etc

python/langsmith/wrappers/_gemini.py
EugeneJinXin: added finish_reason

But time to first token seems not straightforward it needs to work on traceable?

python/langsmith/wrappers/_gemini.py
EugeneJinXin: good reminder, added!

python/langsmith/wrappers/_gemini.py
EugeneJinXin: done!

python/langsmith/wrappers/_gemini.py
EugeneJinXin: ahh this one actually not very clear from python API ref and doc, so might want to do it later

python/langsmith/wrappers/_gemini.py
EugeneJinXin: addressed for both languages!

PR #2067, opened 2025-10-10
support dataframe in get_experiment_results and deprecate get_test_results

EugeneJinXin:
closes https://linear.app/langchain/issue/LS-3161/clientget-test-results-not-returning-feedback

Change summary
- Added response_format to get_experiment_results with overloads:
"ExperimentResults" → ls_schemas.ExperimentResults (default)
"DataFrame" → pd.DataFrame

- Implemented DataFrame conversion:
One row per run; columns: example_id, run_id, inputs (dict), outputs (dict), execution_time, error; optional: reference (dict), notes, feedback.*
Attached experiment metadata via df.attrs["feedback_stats"] and df.attrs["run_stats"] 
Empty DataFrame returns with full expected columns and attrs

- Improved pandas import UX:
try/except ImportError with a clear warning and actionable error message

Key decisions
- One row per run: Chosen for clarity and fidelity to underlying data; avoids lossy aggregation. This is different from default response format, but hypothesis here is 1. user migrated from get_test_results want to see same granularity 2. user of dataframe is analytics driven, so let's make it easy to use. 

- (same as get_test_resutls and csv download of experiment view) Dict columns for inputs/outputs/reference, not flatten

-  run_id naming: Standardized column name to run_id (not id) for explicitness.

-  Metadata via df.attrs: Avoids duplicating experiment-level stats per row; consistent even for empty results.

- Overloads 

- dependency handling: Warn-and-raise pattern for pandas optional dependency 

Test:  All lint checks pass; integration test covers the DataFrame path and critical invariants. we significantly blocked by integration test issues someone need to fix that maybe me but i want to stay focus on bug this month

### DIFF
--- a/python/langsmith/wrappers/_openai.py
+++ b/python/langsmith/wrappers/_openai.py
@@ -514,8 +514,24 @@ def wrap_openai(
     return client
 
 
-def _process_responses_api_output(response: Any) -> dict:
+def _process_responses_api_output(response: Any) -> Any:
     if response:
+        # Check if this is a Stream object - if so, return it as-is
+        # When streaming, the reduce_fn will consume the stream and call this function
+        # again on the completed response object
+        try:
+            from openai import Stream
+            from openai.lib.streaming import AsyncStream
+
+            if isinstance(response, (Stream, AsyncStream)):
+                # Return the stream as-is so reduce_fn can consume it
+                # reduce_fn (_reduce_response_events) will call this function
+                # again on the completed response
+                return response
+        except ImportError:
+            pass
+
+        # Process non-stream response objects
         try:
             output = response.model_dump(exclude_none=True, mode="json")
             if usage := output.pop("usage", None):
@@ -524,5 +540,5 @@ def _process_responses_api_output(response: Any) -> dict:
                 )
             return output
         except Exception:
-            return {"output": response}
+            return response
     return {}


python/tests/integration_tests/test_client.py
```diff
@@ -782,8 +782,9 @@ def test_list_datasets(langchain_client: Client) -> None:
         assert dataset1.url is not None
         assert dataset2.url is not None
 
-        # Test datasets without metadata return empty metadata
-        assert dataset2.metadata is None  # dataset2 has no metadata
+        # Dataset2 has no user-provided metadata, only automatic runtime metadata
+        assert dataset2.metadata is not None
```
EugeneJinXin, line 786: fix flaky

python/langsmith/client.py
```diff
@@ -8343,6 +8356,115 @@ def _paginate_examples_with_runs(
...
+            A pandas DataFrame where 'inputs', 'outputs', and 'reference' are dict
+            columns. Experiment-level metadata is
+            available via DataFrame attributes:
+                - df.attrs["feedback_stats"]: Combined feedback statistics including
+                  session-level feedback
+                - df.attrs["run_stats"]: Aggregated run statistics (latency, tokens,
+                  cost, etc.)
+            Required columns:
+                - example_id: UUID of the example
+                - run_id: UUID of the run
```
EugeneJinXin, line 8379: another way is represent each row as Example level, and do this:

"""
    If max_reps > 1, creates columns like:
    "session_name_rep_1_outputs", "session_name_rep_1_run", ...
   "session_name_rep_2_outputs", "session_name_rep_2_run", ...
"""

Above Example-per-row logic is what we have for experiment view's download csv results, so i think it's a good way as well, but less analytics friendly from what i've seen on data engineering world when you try to apply simple filtering/groupby/agg across runs


Example-level wide is better when it's about UI display:
You want side-by-side comparison per example (mirrors UI “Download CSV”)

PR #2070, opened 2025-10-12
feat(python): OpenAI Agents Tracing Interoperability

### LINKED ISSUE #2063: Feature request: Improved tracing with Open AI Agent SDK
I have been diving deep into LangSmith documentation on tracing as its strongly related to some of the projects I have been working on. I am using OpenAI Agent SDK for agentic AI.

I see that a lot of tracing functionality (ex. passing  custom metadata, or anonymising I/O) is integrated with `@traceable` decorators or openAI wrappers `wrap_openai` which doesn't work well with global tracing via `set_trace_processors([OpenAIAgentsTracingProcessor()])` -  as they duplicate the traces.

Would love to see in the future a bit more flexibility in integrating the customisation and fine control of tracing provided by langsmith with OpenAI Agent SDK framework. 

https://docs.langchain.com/langsmith/trace-with-openai-agents-sdk

angus-langchain:
## Description
Use RunTree API for openai agents sdk to improve tracing devex by allowing for input/output processing, nested traces with `@traceable` and `wrap_openai`.

## Test Plan
- [x] Added unit tests

Resolves #2063

### DIFF
--- a/python/langsmith/wrappers/_openai_agents.py
+++ b/python/langsmith/wrappers/_openai_agents.py
@@ -205,7 +205,7 @@ if HAVE_AGENTS:
                     }
                     if self._project_name is not None:
                         run_kwargs["project_name"] = self._project_name
-                    new_run = rt.RunTree(**run_kwargs)
+                    new_run = rt.RunTree(**run_kwargs)  # type: ignore[arg-type]
 
                 new_run.post()
                 _context._PARENT_RUN_TREE.set(new_run)


python/langsmith/__init__.py
```diff
@@ -21,7 +21,7 @@
 
 # Avoid calling into importlib on every call to __version__
 
-__version__ = "0.4.34"
+__version__ = "0.4.35rc1"
```
jacoblee93: Revert before merging

python/langsmith/wrappers/_openai_agents.py
```diff
@@ -178,118 +175,101 @@ def __init__(
...
+                        "client": self.client,
+                    }
+                    if self._project_name is not None:
+                        run_kwargs["project_name"] = self._project_name
+                    new_run = rt.RunTree(**run_kwargs)
+
+                new_run.post()
+
+                ctx = copy_context()
+                ctx.run(_context._PARENT_RUN_TREE.set, new_run)
```
jacoblee93: Why is storing context necessary? We're just pulling the parent out of the span maps aren't we?

python/langsmith/wrappers/_openai_agents.py
```diff
@@ -91,13 +93,8 @@ def __init__(self, *args, **kwargs):
...
     class RunData(TypedDict):
-        id: str
-        trace_id: str
-        start_time: datetime
-        dotted_order: str
-        parent_run_id: Optional[str]
-        project_name: Optional[str]
-        name: Optional[str]
+        run_tree: rt.RunTree
+        context: contextvars.Context
```
jacoblee93: See below, do we need to store `context` here?

python/langsmith/wrappers/_openai_agents.py
angus-langchain: modifying the context is what enables `@traceable` and `wrap_openai` to automatically find the agent span as a parent without us having to pass it explicitly

angus-langchain: via setting `_PARENT_RUN_TREE`

python/langsmith/wrappers/_openai_agents.py
angus-langchain: This is correct though, we don't need to actually store it in the RunData map

angus-langchain: removed

python/langsmith/wrappers/_openai_agents.py
```diff
@@ -91,13 +93,7 @@ def __init__(self, *args, **kwargs):
...
 
     class RunData(TypedDict):
-        id: str
-        trace_id: str
-        start_time: datetime
-        dotted_order: str
-        parent_run_id: Optional[str]
-        project_name: Optional[str]
-        name: Optional[str]
+        run_tree: rt.RunTree
```
jacoblee93: nit: why not just store RunTrees directly

angus-langchain: we are, just importing `from langsmith import run_trees as rt`

PR #2073, opened 2025-10-15
feat(python): Claude Agent SDK wrapper

### LINKED ISSUE #2062: Claude Agent SDK tracing
### Feature request

Any plans to support tracing agent runs with the Claude Agent SDK?

### Motivation

I'm frustrated by the seeming lack of options to trace claude agent sdk runs.

angus-langchain:
## Description
Adds a Claude Agent SDK Wrapper to python

Partially resolves #2062

### DIFF
--- a/python/langsmith/__init__.py
+++ b/python/langsmith/__init__.py
@@ -21,7 +21,7 @@ if TYPE_CHECKING:
 
 # Avoid calling into importlib on every call to __version__
 
-__version__ = "0.4.35"
+__version__ = "0.4.36"
 version = __version__  # for backwards compatibility
 
 
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "langsmith"
-version = "0.4.35"
+version = "0.4.36"
 description = "Client library to connect to the LangSmith LLM Tracing and Evaluation Platform."
 authors = [
     {name = "LangChain", email = "support@langchain.dev"},
--- a/python/uv.lock
+++ b/python/uv.lock
@@ -738,7 +738,7 @@ wheels = [
 
 [[package]]
 name = "langsmith"
-version = "0.4.35"
+version = "0.4.36"
 source = { editable = "." }
 dependencies = [
     { name = "httpx" },


python/langsmith/integrations/claude_agent_sdk/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+
+        # Best-effort propagation to already-imported modules.
+        for module in list(sys.modules.values()):
+            try:
+                if module and getattr(module, symbol, None) is original:
+                    setattr(module, symbol, wrapped)
+            except Exception:
+                continue
+
+    return True
```
jacoblee93, line 88: nit: Can we avoid altering the original client and return a wrapped version?

python/langsmith/wrappers/_openai_agents.py
```diff
@@ -1,347 +1,9 @@
...
-from datetime import datetime
-from typing import Optional
+"""Tombstone module for backward compatibility.
 
-from langsmith import run_trees as rt
-from langsmith._internal import _context
-from langsmith.run_helpers import get_current_run_tree
+This module has been moved to langsmith.integrations.openai_agents.
+Imports from this location are deprecated but will continue to work.
+"""
```
jacoblee93, line 5: Think generally there's a way to log a deprecation warning

python/langsmith/integrations/claude_agent_sdk/_client.py
```diff
@@ -0,0 +1,198 @@
...
+        self.next_start_time: Optional[float] = query_start_time
+
+    def start_llm_run(
+        self, message: Any, prompt: Any, history: list[dict[str, Any]]
+    ) -> Optional[dict[str, Any]]:
+        """Begin a new model run, ending any existing one."""
+        start = self.next_start_time or time.time()
+
+        if self.current_run:
+            self.current_run.end()
```
jacoblee93, line 38: Is this really the only way to get end time?

This feels brittle/inaccurate

python/langsmith/integrations/claude_agent_sdk/_client.py
```diff
@@ -0,0 +1,198 @@
...
+        if hasattr(last_msg, "content")
+        else None
+    )
+    return final_content, llm_run
+
+
+def instrument_claude_client(original_class: Any) -> Any:
+    """Wrap ClaudeSDKClient to trace both query() and receive_response()."""
+
+    class TracedClaudeSDKClient:
```
jacoblee93: Would it make sense for this to inherit from the original class instance so that you get all the original client methods?

python/langsmith/integrations/claude_agent_sdk/_client.py
```diff
@@ -0,0 +1,198 @@
...
+            return await self._client.query(*args, **kwargs)
+
+        async def receive_response(self) -> AsyncGenerator[Any, None]:
+            """Intercept message stream and record chain run activity."""
+            messages = self._client.receive_response()
+            async with trace(
+                name=TRACE_CHAIN_NAME,
+                run_type="chain",
+                inputs={"prompt": self._prompt} if self._prompt else None,
+            ) as run:
```
jacoblee93, line 180: Shouldn't this be created in `query`?

Can we instead do something similar to the OpenAI Agents implementation and keep a map of current runs and their parents?

python/langsmith/integrations/claude_agent_sdk/_tools.py
```diff
@@ -0,0 +1,105 @@
...
+
+import logging
+import threading
+from typing import Any, Callable
+
+from langsmith.run_helpers import get_current_run_tree, trace
+
+logger = logging.getLogger(__name__)
+
+# Thread-local store for passing the parent run tree into tool handlers.
```
jacoblee93, line 11: So tools are run on separate threads?

angus-langchain: yes, they're run on the Claude Code CLI, which handles the tool execution.

python/langsmith/integrations/claude_agent_sdk/__init__.py
angus-langchain: good point

python/langsmith/integrations/claude_agent_sdk/_client.py
angus-langchain: discessed offline

python/langsmith/wrappers/_openai_agents.py
```diff
@@ -1,347 +1,18 @@
...
-        "Trace",
-        "Span",
-        "ResponseSpanData",
-    )
-    if not all(hasattr(tracing, name) for name in required):
-        raise ImportError("The `agents` package is not installed.")
+from langsmith.integrations.openai_agents_sdk import OpenAIAgentsTracingProcessor
 
-    import langsmith.wrappers._agent_utils as agent_utils
+warnings.warn(
```
jacoblee93, line 11: Just making sure this will only log once?

Also do we need to update docs?

angus-langchain: yeah need to update docs, will do before merging this one

angus-langchain: and this should only log once

PR #2074, opened 2025-10-15
feat(py,js): Unnest example inputs in evaluations

jacoblee93:
Existing format is unusable for certain flows such as adding experiment runs to annotation queues and adding those to another dataset

Should release in a minor bump

### DIFF
--- a/js/src/evaluation/_runner.ts
+++ b/js/src/evaluation/_runner.ts
@@ -1027,7 +1027,6 @@ async function _forward(
     },
     client,
     tracingEnabled: true,
-    processInputs: (inputs: KVMap) => inputs?.inputs,
   };
 
   const wrappedFn = isTraceableFunction(fn)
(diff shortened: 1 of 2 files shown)


js/src/evaluation/_runner.ts
```diff
@@ -1027,36 +1027,38 @@ async function _forward(
     },
     client,
     tracingEnabled: true,
+    processInputs: (inputs: KVMap) => inputs?.inputs,
```
dqbd: _nit_: should we fallback to `inputs` if `inputs?.inputs` is not defined? Or are we 100% certain that `inputs?.inputs` do exist?


```suggestion
    processInputs: (inputs: KVMap) => inputs?.inputs ?? inputs,
```

dqbd (approved):

js/src/evaluation/_runner.ts
jacoblee93: I think all useful examples have inputs right?

jacoblee93: Hmm this seems to actually work in JS

PR #2076, opened 2025-10-16
fix: support response API stream object parse

EugeneJinXin:
OpenAI Streaming Response Serialization Bug
When using OpenAI's streaming APIs (both chat.completions.create(stream=True) and responses.create(stream=True))
 with LangSmith tracing via wrap_openai(), the trace output was being serialized as a string representation of the Stream object:  "<openai.Stream object at 0x113355010>" instead of the actual response data.

The stream isn't really a pydantic model and won't have model_dump(), so  Exception is caught, and {"output": response} is returned,  When serialized to JSON for LangSmith, Python converts the Stream object to its string representation: "<openai.Stream object at 0x...>"

### DIFF
--- a/python/langsmith/wrappers/_openai.py
+++ b/python/langsmith/wrappers/_openai.py
@@ -514,8 +514,24 @@ def wrap_openai(
     return client
 
 
-def _process_responses_api_output(response: Any) -> dict:
+def _process_responses_api_output(response: Any) -> Any:
     if response:
+        # Check if this is a Stream object - if so, return it as-is
+        # When streaming, the reduce_fn will consume the stream and call this function
+        # again on the completed response object
+        try:
+            from openai import Stream
+            from openai.lib.streaming import AsyncStream
+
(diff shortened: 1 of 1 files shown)


python/langsmith/wrappers/_openai.py
```diff
@@ -514,8 +514,24 @@ def wrap_openai(
...
+                # Return the stream as-is so reduce_fn can consume it
+                # reduce_fn (_reduce_response_events) will call this function
+                # again on the completed response
+                return response
+        except ImportError:
+            pass
+
+        # Process non-stream response objects
         try:
             output = response.model_dump(exclude_none=True, mode="json")
```
EugeneJinXin, line 536: The stream isn't really a pydantic model and won't have model_dump(), so Exception is caught, and {"output": response} is returned, When serialized to JSON for LangSmith, Python converts the Stream object to its string representation: "<openai.Stream object at 0x...>"

PR #2081, opened 2025-10-20
fix(python): Fix CI by bumping deps

jacoblee93:


.github/workflows/python_test.yml
```diff
@@ -38,7 +38,6 @@ jobs:
     strategy:
       matrix:
         python-version:
-          - "3.9"
           - "3.10"
           - "3.11"
           - "3.12"
```
angus-langchain, line 43: should we add 3.14

jacoblee93: Think that's a little bit hairier, but can try it!

PR #2088, opened 2025-10-25
Allow trace.end(outputs=str)

xl0:
Just a bit of sugar that lets you pass a string to tr.end(). I suspect a string is the most common thing to pass here:
```py
    with ls.trace("Template", "prompt", inputs=input) as trace:
        prompt = template.render(**input)
        trace.end(outputs=prompt)
```
Same for llm calls that don't use a schema.

A related request - add a button to render text blocks in the output as markdown in the web ui.

### DIFF
--- a/python/langsmith/run_trees.py
+++ b/python/langsmith/run_trees.py
@@ -440,7 +440,7 @@ class RunTree(ls_schemas.RunBase):
     def end(
         self,
         *,
-        outputs: Optional[dict] = None,
+        outputs: Optional[Union[dict, str]] = None,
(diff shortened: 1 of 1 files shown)


xl0:
I think I can just allow str to all functions that take inputs/outputs. I converted them, most were trivial. If you like the idea, I'll add it to the PR.

```python
utils.py
def str_to_dict(item: Union[dict, str], key: str) -> dict:
    if isinstance(item, str):
        item = {key: item}
    return item
...

inputs = str_to_dict(inputs, "input")
```

Issue #2091, opened 2025-10-27
Issue: LangSmith tracing does not capture subagent tool calls in Claude Agent SDK


Z-Jr10-K:
### Issue you'd like to raise.

When using `configure_claude_agent_sdk()` with Claude Agent SDK, tool invocations made by **subagents** (defined via `agents` and `mcp_servers`) are **not recorded** in LangSmith traces.
Only the main agent conversation appears.

### **Reproduction**

```python
import asyncio
from typing import Any
from claude_agent_sdk import (
    ClaudeAgentOptions,
    ClaudeSDKClient,
    AgentDefinition,
    tool,
    create_sdk_mcp_server
)
from langsmith.integrations.claude_agent_sdk import configure_claude_agent_sdk

configure_claude_agent_sdk()

@tool("add_numbers", "Add two integers.", {"a": int, "b": int})
async def add_numbers(args: dict[str, Any]) -> dict[str, Any]:
    a, b = args["a"], args["b"]
    return {"content": [{"type": "text", "text": f"{a} + {b} = {a + b}"}]}

math_helper = create_sdk_mcp_server(name="math-helper", version="1.0.0", tools=[add_numbers])

async def main():
    subagent = AgentDefinition(
        description="Performs arithmetic tasks.",
        prompt="You are a math specialist.",
        tools=["add_numbers"],
    )

    options = ClaudeAgentOptions(
        mcp_servers={"math-helper": math_helper},
        allowed_tools=["mcp__math-helper__add_numbers"],
        agents={"math-helper": subagent},
        system_prompt="You are the main orchestrator. Delegate math to math-helper.",
    )

    async with ClaudeSDKClient(options=options) as client:
        await client.query("Ask the math-helper agent to add 3 and 5.")
        async for msg in client.receive_response():
            print(msg)

if __name__ == "__main__":
    asyncio.run(main())
```

### **Observed**

- Only main `claude.conversation` trace appears in LangSmith.
- No trace for subagent `math-helper` or its tool `add_numbers`.

*Example output:*

<img width="1143" height="661" alt="Image" src="https://github.com/user-attachments/assets/c0951e28-6a92-4d8c-9488-a74c6934ebec" />

LangSmith should record nested traces for subagents and their tool calls, e.g.:

```
claude.conversation
 └── math-helper
      └── Tool: add_numbers(a=3, b=5)
```

### **Environment**

| Component        | Version           |
| ---------------- | ----------------- |
[...]

PR #2102, opened 2025-11-03
feat(python): Fix claude tool tracing by using hooks

angus-langchain:
Resolves https://github.com/langchain-ai/langsmith-sdk/issues/2091 by hooking into pre and post tool call hooks instead of tracing the tool directly.


jacoblee93 (approved):
Nice

PR #2103, opened 2025-11-03
v2 API Changes

angus-langchain:
### Description
For v2 Ingestion APIs, we need to support uuidv7 for run and trace identifiers. Additionally, we need to ensure that `run_type`,  `name`, and `start_time` are passed in both the POST and PATCH request.

### Changes
1. Vendors uuidv7 from python 3.14 standard lib
2. Modified ingestion to use uuidv7 with the time identifier seeded by the run `start_time` if possible. For manually provided run_ids, I added a simple `~isUUID7~` check that emits a warning.
3. Ensured run_type, name, and start_time are passed through both js and python PATCH requests
4. Removes run id remapping between project replicas since no requirements exists for unique run ids for a single tenant in between projects

### Tests
Added unit tests in both js and python for uuid7 compliance and sending run_type, name, and start_time in patch

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "langsmith",
-  "version": "0.3.79-rc.0",
+  "version": "0.3.79",
   "description": "Client library to connect to the LangSmith Observability and Evaluation Platform.",
   "packageManager": "yarn@1.22.19",
   "files": [
--- a/js/src/experimental/vercel/index.ts
+++ b/js/src/experimental/vercel/index.ts
@@ -525,6 +525,18 @@ const wrapAISDK = <
           if (outputs.outputs == null || typeof outputs.outputs !== "object") {
             return outputs;
           }
+          // If experimental_output is present, return it directly at top level (like generateObject)
+          // Note: accessing experimental_output throws if not specified, so wrap in try-catch
+          try {
+            if ("experimental_output" in outputs.outputs) {
+              const experimentalOutput = outputs.outputs.experimental_output;
+              if (experimentalOutput != null) {
+                return experimentalOutput;
+              }
+            }
+          } catch (e) {
+            // experimental_output not specified, continue with normal processing
+          }
           const { steps } = outputs.outputs;
           if (Array.isArray(steps)) {
             const lastStep = steps.at(-1);
@@ -721,6 +733,17 @@ const wrapAISDK = <
               !["object", "string"].includes(typeof content)
             ) {
               return outputs;
+            }
+            try {
+              if (
+                "experimental_partialOutputStream" in outputs.outputs &&
+                outputs.outputs.experimental_partialOutputStream != null
+              ) {
+                const textContent = await outputs.outputs.text;
+                return JSON.parse(textContent);
+              }
+            } catch (e) {
+              // experimental_partialOutputStream not specified, continue with normal processing
             }
             let responseMetadata: Record<string, unknown> | undefined =
               undefined;
--- a/js/src/index.ts
+++ b/js/src/index.ts
@@ -22,4 +22,4 @@ export { getDefaultProjectName } from "./utils/project.js";
 export { uuid7, uuid7FromTime } from "./uuid.js";
 
 // Update using yarn bump-version
-export const __version__ = "0.3.79-rc.0";
+export const __version__ = "0.3.79";
--- a/js/src/tests/vercel/wrapper/methods.vitesttest.ts
+++ b/js/src/tests/vercel/wrapper/methods.vitesttest.ts
@@ -1237,4 +1237,156 @@ describe("wrapAISDK", () => {
       expect(mockHttpRequests[0].body.name).toBe("custom-provider");
     });
   });
+
+  describe("experimental_output handling", () => {
+    it("should preserve experimental_output as structured object in outputs", async () => {
+      const wrappedMethods = wrapAISDK(
+        {
+          wrapLanguageModel: ai.wrapLanguageModel,
+          generateText: ai.generateText,
+          streamText: ai.streamText,
+          generateObject: ai.generateObject,
+          streamObject: ai.streamObject,
+        },
+        { client: mockClient as any }
+      );
+
+      const mockLangModel = new MockLanguageModelV2({
+        modelId: "experimental-output-test",
+        doGenerate: async () => ({
+          content: [
+            {
+              type: "text" as const,
+              text: '{"city":"Prague","temperature":15,"unit":"celsius","conditions":"sunny"}',
+            },
+          ],
+          finishReason: "stop" as const,
+          usage: {
+            promptTokens: 10,
+            completionTokens: 20,
+            inputTokens: 10,
+            outputTokens: 20,
+            totalTokens: 30,
+          },
+          warnings: [],
+        }),
+      });
+
+      await wrappedMethods.generateText({
+        model: mockLangModel,
+        prompt: "What's the weather?",
+        experimental_output: ai.Output.object({
+          schema: z.object({
+            city: z.string(),
+            temperature: z.number(),
+            unit: z.enum(["celsius", "fahrenheit"]),
+            conditions: z.string(),
+          }),
+        }),
+      });
+
+      await new Promise((resolve) => setTimeout(resolve, 1));
+
+      // Find the updateRun call for generateText (parent run)
+      const updateGenerateTextCall = mockHttpRequests.find(
+        (req) =>
+          req.type === "updateRun" &&
+          req.body.extra?.metadata?.ai_sdk_method === "ai.generateText"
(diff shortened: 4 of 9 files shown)


js/src/run_trees.ts
```diff
@@ -456,63 +474,11 @@ export class RunTree implements BaseRun {
     excludeChildRuns = true
   ): RunCreate & { id: string } {
     const baseRun = this._convertToCreate(this, runtimeEnv, excludeChildRuns);
-    if (projectName === this.project_name) {
-      return baseRun;
-    }
-
-    // Create a deterministic UUID mapping for this project
-    const createRemappedId = (originalId: string): string => {
```
angus-langchain, line 467: we dont need this as there's no unique constraint on ids between projects on the same tenant

python/langsmith/run_trees.py
```diff
@@ -552,54 +550,17 @@ def _slice_parent_id(self, parent_id: str, run_dict: dict) -> None:
...
-        if project_name == self.session_name:
-            return run_dict
 
         if updates and updates.get("reroot", False):
             distributed_parent_id = _DISTRIBUTED_PARENT_ID.get()
-
             if distributed_parent_id:
                 self._slice_parent_id(distributed_parent_id, run_dict)
 
-        old_id = run_dict["id"]
```
angus-langchain, line 565: same here

js/src/client.ts
```diff
@@ -1282,7 +1282,7 @@ export class Client implements LangSmithTracingClientInterface {
       session_name,
       ...run,
       start_time: run.start_time ?? Date.now(),
-    });
+    } as RunCreate);
```
jacoblee93, line 1285: nit: does `as const` work? What changed about this cast?

python/langsmith/_internal/_uuid.py
```diff
@@ -0,0 +1,140 @@
...
+    """Generate random counter and tail values for UUID v7."""
+    rand = int.from_bytes(os.urandom(10), byteorder="big")
+    # 42-bit counter with MSB set to 0
+    counter = (rand >> 32) & 0x1FF_FFFF_FFFF
+    # 32-bit random data
+    tail = rand & 0xFFFF_FFFF
+    return counter, tail
+
+
+def uuid7(nanoseconds: int | None = None) -> uuid.UUID:
```
angus-langchain, line 36: this is vendored from the python standard lib with the option of passing in a seed start_time

https://github.com/python/cpython/blob/main/Lib/uuid.py

js/src/run_trees.ts
```diff
@@ -246,13 +246,30 @@ export class RunTree implements BaseRun {
     };
     config.extra = { ...config.extra, metadata: dedupedMetadata };
     Object.assign(this, { ...defaultConfig, ...config, client });
+
+    // If start_time was provided but id was not, regenerate id from start_time
+    if (config.start_time !== undefined && !config.id) {
+      this.id = uuidFromTime(config.start_time);
```
jacoblee93: There's new hard requirement that the run id must be generated from start time then right?

jacoblee93: If so you should actually use `this._serialized_start_time` below:

https://github.com/langchain-ai/langsmith-sdk/pull/2103/files#diff-fa3bb02ee6fa794ab2494fe9c12ab64d8f1daac80bbffcf54ae88f50934ae20bR290

js/src/run_trees.ts
```diff
@@ -289,16 +306,17 @@ export class RunTree implements BaseRun {
   }
 
   private static getDefaultConfig(): object {
+    const start_time = Date.now();
     return {
-      id: uuid.v4(),
+      id: uuidFromTime(start_time),
```
jacoblee93: This is misleading, let's refactor this a bit to use `_serialized_start_time` as I mentioned above

Or just remove `id` from this altogether

js/src/run_trees.ts
jacoblee93: Oh so two different projects can have two different runs with the same id?

js/src/schemas.ts
```diff
@@ -221,6 +221,9 @@ export interface RunCreate extends BaseRun {
 
 export interface RunUpdate {
   id?: string;
+  name?: string;
+  run_type?: string;
+  start_time?: number | string;
```
jacoblee93, line 226: Don't think we want to add `start_time` here since we won't allow those to change on update?

js/src/utils/_uuid.ts
```diff
@@ -14,3 +19,54 @@ export function assertUuid(str: string, which?: string): string {
...
+ * @param idType - The type of ID (e.g., "run_id", "trace_id") for the warning message
+ */
+export function warnIfNotUuidV7(uuidStr: string, _idType: string): void {
+  const version = getUuidVersion(uuidStr);
+  if (version !== null && version !== 7 && !UUID7_WARNING_EMITTED) {
+    UUID7_WARNING_EMITTED = true;
+    warnOnce(
+      `LangSmith now uses UUID v7 for run and trace identifiers. ` +
+        `Please migrate to using UUID v7. ` +
+        `Use { uuid7, uuidFromTime } from 'langsmith'. ` +
```
jacoblee93: nit: show an actual code sample for this and mention this can happen if you are passing custom ids

js/src/utils/_uuid.ts
```diff
@@ -14,3 +19,54 @@ export function assertUuid(str: string, which?: string): string {
...
   return str;
 }
+
+/**
+ * Generate a UUID v7 from a timestamp.
+ *
+ * @param timestamp - The timestamp in milliseconds
+ * @returns A UUID v7 string
+ */
+export function uuidFromTime(timestamp: number | string): string {
```
jacoblee93: nit: Keep naming standard with Python

Also let's export these from the main `langsmith` entrypoint

python/langsmith/run_trees.py
```diff
@@ -693,6 +671,8 @@ def patch(self, *, exclude_inputs: bool = False) -> None:
             self.client.update_run(
                 name=self.name,
                 run_id=self.id,
+                run_type=cast(RUN_TYPE_T, self.run_type),
+                start_time=self.start_time,
```
jacoblee93, line 671: Why send this twice?

js/src/schemas.ts
jacoblee93: Also what's the benefit to allowing `name` and `run_type` after the fact?

js/src/run_trees.ts
```diff
@@ -825,36 +797,6 @@ export function isRunnableConfigLike(x?: unknown): x is RunnableConfigLike {
   );
 }
 
-function _parseDottedOrder(dottedOrder: string): [Date, string][] {
```
jacoblee93, line 831: Why remove this? Just unused?

js/src/run_trees.ts
```diff
@@ -561,6 +527,9 @@ export class RunTree implements BaseRun {
         const runData = this._remapForProject(projectName ?? this.project_name);
         const updatePayload: RunUpdate = {
           id: runData.id,
+          name: runData.name,
+          run_type: runData.run_type,
+          start_time: runData.start_time,
```
jacoblee93, line 549: Let's not double send this

js/src/run_trees.ts
angus-langchain: yup!

js/src/client.ts
angus-langchain: didn't change anything here, was complaining and I fixed. `prepareRunCreateOrUpdateInputs` has an overload for RunCreate and RunUpdate

js/src/run_trees.ts
angus-langchain: its not a hard requirement, its soft, but we should generate from start_time

js/src/schemas.ts
angus-langchain: we dont allow changes these, we need to send the same values in the patch

js/src/run_trees.ts
angus-langchain: yes unused

js/src/run_trees.ts
jacoblee93: Nice

js/src/run_trees.ts
```diff
@@ -246,13 +246,30 @@ export class RunTree implements BaseRun {
     };
     config.extra = { ...config.extra, metadata: dedupedMetadata };
     Object.assign(this, { ...defaultConfig, ...config, client });
+
+    // Generate id from start_time if not provided
+    if (!this.id) {
+      this.id = uuid7FromTime(this._serialized_start_time ?? this.start_time);
```
jacoblee93, line 274: `_serialized_start_time` gets set below, need to take that into account

PR #2107, opened 2025-11-04
fix(python): reduce memory of pending traces

angus-langchain:
### Description
Remove `EXCLUDE_CHILD_RUNS` env var to make memory optimizations default behavior


python/langsmith/run_trees.py
```diff
@@ -631,9 +630,6 @@ def post(self, exclude_child_runs: bool = True) -> None:
         if not exclude_child_runs:
             for child_run in self.child_runs:
                 child_run.post(exclude_child_runs=False)
-        elif utils.is_env_var_truish("EXCLUDE_CHILD_RUNS"):
```
angus-langchain, line 634: this code is never hit

angus-langchain: confirmed in langchain as well

hinthornw (approved):

PR #2114, opened 2025-11-06
fix(py): pairwise eval concurrency

baskaryan:


EugeneJinXin (approved):
provded this is the right fix by adding print on active threads count. LGTM as this fixes issue. 
```
        if futures:
            print(f"waiting for {len(futures)} futures")
            running = sum([1 for f in futures if f.running()])
            print(f"running threads count: {running}")
            cf.wait(futures)
            for future in futures:
                example_id, result = future.result()
                results[example_id][f"feedback.{result.key}"] = result

```
^ This would give active threads equal to concurrency

PR #2115, opened 2025-11-07
feat(py): add metadata helper

bees:
Adds a convenience method for setting metadata on the current run tree.

Example usage:
```python
@traceable
def func(x):
    langsmith.set_run_metadata(foo=1)
    return {"bar": 2}
```


samnoyes (approved):
Can you run `poetry version patch` in python dir to release this

PR #2121, opened 2025-11-12
beta wrap_gemini ts wrapper

EugeneJinXin:
Adds beta TypeScript wrapper for Google Gemini API with automatic LangSmith tracing.

Features
- Text generation (streaming & non-streaming)
- Multimodal inputs (text + images)
- Image generation output (gemini-2.5-flash-image)
- Function calling
- Usage metadata extraction
- compatible format for LangSmith UI rendering (this is beta so it's only a starting point)

### DIFF
--- a/js/src/wrappers/gemini.ts
+++ b/js/src/wrappers/gemini.ts
@@ -5,6 +5,14 @@ import {
   type TraceableConfig,
 } from "../traceable.js";
 import { KVMap, InvocationParamsSchema } from "../schemas.js";
+import type {
+  Content,
+  GenerateContentParameters,
+  GenerateContentResponse,
+  GenerateContentResponseUsageMetadata,
+  SafetyRating,
+  UsageMetadata,
+} from "@google/genai";
 
 type GoogleGenAIType = {
   models: {
@@ -22,7 +30,7 @@ type PatchedGeminiClient<T extends GoogleGenAIType> = T & {
   };
 };
 
-interface UsageMetadata {
+interface LangSmithUsageMetadata {
   input_tokens: number;
   output_tokens: number;
   total_tokens: number;
@@ -37,10 +45,22 @@ interface UsageMetadata {
   };
 }
 
-const _createUsageMetadata = (usage: Record<string, any>): KVMap => {
-  const usageMetadata: UsageMetadata = {
+const _createUsageMetadata = (
+  usage: UsageMetadata | GenerateContentResponseUsageMetadata
+): KVMap => {
+  const usageMetadata: LangSmithUsageMetadata = {
     input_tokens: usage.promptTokenCount || 0,
-    output_tokens: usage.responseTokenCount || usage.candidatesTokenCount || 0,
+    output_tokens: (() => {
+      if ("responseTokenCount" in usage) {
+        return usage.responseTokenCount || 0;
+      }
+
+      if ("candidatesTokenCount" in usage) {
+        return usage.candidatesTokenCount || 0;
+      }
+
+      return 0;
+    })(),
     total_tokens: usage.totalTokenCount || 0,
   };
 
@@ -59,9 +79,10 @@ const _createUsageMetadata = (usage: Record<string, any>): KVMap => {
 
   // Add output token details if available
   usageMetadata.output_token_details = {
-    ...(usage.candidatesTokenCount && {
-      over_200k: Math.max(0, usage.candidatesTokenCount - 200000),
-    }),
+    ...("candidatesTokenCount" in usage &&
+      usage.candidatesTokenCount != null && {
+        over_200k: Math.max(0, usage.candidatesTokenCount - 200000),
+      }),
     ...(usage.thoughtsTokenCount && {
       reasoning: usage.thoughtsTokenCount,
     }),
@@ -70,7 +91,13 @@ const _createUsageMetadata = (usage: Record<string, any>): KVMap => {
   return usageMetadata;
 };
 
-const chatAggregator = (chunks: Record<string, any>[]): KVMap => {
+const chatAggregator = (input: unknown): KVMap => {
+  const chunks =
+    Array.isArray(input) &&
+    input.every((item) => typeof item === "object" && item !== null)
+      ? (input as GenerateContentResponse[])
+      : [];
+
   if (!chunks || chunks.length === 0) {
     return { content: "", role: "assistant" };
   }
@@ -79,9 +106,9 @@ const chatAggregator = (chunks: Record<string, any>[]): KVMap => {
   let thoughtText = "";
   const toolCalls: Array<Record<string, unknown>> = [];
   const otherParts: Array<Record<string, unknown>> = [];
-  let usageMetadata: Record<string, any> | null = null;
+  let usageMetadata: GenerateContentResponseUsageMetadata | null = null;
   let finishReason: string | null = null;
-  let safetyRatings: Array<Record<string, any>> | null = null;
+  let safetyRatings: SafetyRating[] | null = null;
 
   for (const chunk of chunks) {
     if (chunk?.usageMetadata) {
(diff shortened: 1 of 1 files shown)


js/src/wrappers/gemini.ts
```diff
@@ -0,0 +1,410 @@
...
+    if (!currentRun.extra) currentRun.extra = {};
+    if (!currentRun.extra.metadata) currentRun.extra.metadata = {};
+    if (!currentRun.extra.metadata.usage_metadata) {
+      currentRun.extra.metadata.usage_metadata = {};
+    }
+
+    Object.assign(currentRun.extra.metadata.usage_metadata, usageMetadata);
+
+    // Fire-and-forget patch to persist changes asynchronously
+    currentRun.patchRun().catch(() => {});
```
jacoblee93: This is already backgrounded, just await it

js/src/wrappers/gemini.ts
```diff
@@ -0,0 +1,410 @@
...
+    if (!currentRun) return;
+
+    // Initialize nested structure
+    if (!currentRun.extra) currentRun.extra = {};
+    if (!currentRun.extra.metadata) currentRun.extra.metadata = {};
+    if (!currentRun.extra.metadata.usage_metadata) {
+      currentRun.extra.metadata.usage_metadata = {};
+    }
+
+    Object.assign(currentRun.extra.metadata.usage_metadata, usageMetadata);
```
jacoblee93: Just spread it instead, this method can have some funny side-effects

js/src/wrappers/gemini.ts
```diff
@@ -0,0 +1,410 @@
...
+    ),
+  };
+
+  if (gemini.aio?.models) {
+    tracedGeminiClient.aio = {
+      ...gemini.aio,
+      models: {
+        ...gemini.aio.models,
+        generateContent: traceable(
+          gemini.aio.models.generateContent.bind(gemini.aio.models),
```
jacoblee93: Is it really `aio`? What does this mean?

js/src/wrappers/gemini.ts
```diff
@@ -0,0 +1,410 @@
...
+
+              contentParts.push({
+                type: "image_url",
+                image_url: {
+                  url: `data:${mimeType};base64,${data}`,
+                  detail: "high",
+                },
+              });
+            } else if (
+              "functionResponse" in part ||
```
jacoblee93: Why both?

js/src/wrappers/gemini.ts
jacoblee93: Actually - why do you need this at all?

js/src/wrappers/gemini.ts
EugeneJinXin: you are right, i double checked and only snake case is needed as expected per JS convention. addressed

js/src/wrappers/gemini.ts
EugeneJinXin: my bad overlooking this.  fixed.

js/src/wrappers/gemini.ts
EugeneJinXin: makes sense

EugeneJinXin: addressed

js/src/wrappers/gemini.ts
EugeneJinXin: we need the usageMetadata to store in currentRun.extra.metadata.usage_metadata since that's new requirement from the DB team.

.github/actions/js-integration-tests/action.yml
```diff
@@ -10,6 +10,9 @@ inputs:
   openai-api-key:
     description: "OpenAI API key"
     required: false
+  gemini-api-key:
```
jacoblee93, line 13: Is this populated in CI? I don't think so?

Let's just omit for now?

js/src/wrappers/gemini.ts
```diff
@@ -0,0 +1,424 @@
...
+  input_tokens: number;
+  output_tokens: number;
+  total_tokens: number;
+  input_token_details?: {
+    cache_read?: number;
+    cache_read_over_200k?: number;
+    over_200k?: number;
+  };
+  output_token_details?: {
+    reasoning?: number;
```
jacoblee93, line 43: `over_200k` needs to be on output too

js/src/wrappers/gemini.ts
```diff
@@ -0,0 +1,424 @@
...
+      currentRun.extra.metadata.usage_metadata = {};
+    }
+
+    currentRun.extra.metadata.usage_metadata = {
+      ...currentRun.extra.metadata.usage_metadata,
+      ...usageMetadata,
+    };
+
+    // Await patch to ensure proper completion
+    await currentRun.patchRun();
```
jacoblee93: You shouldn't need to do this since this `patch` is handled by `traceable`

.github/actions/js-integration-tests/action.yml
EugeneJinXin: oh, actually populated already; i added GEMINI_API_KEY in github action

EugeneJinXin: it is used for integration test of gemini wrapper

js/src/wrappers/gemini.ts
EugeneJinXin: tested, you are right! think me and Angus didn't realize this when we orginally discuss

js/src/wrappers/gemini.ts
```diff
@@ -0,0 +1,406 @@
...
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const chatAggregator = (chunks: any[]): any => {
+  const fullText = chunks
+    .filter((chunk) => chunk?.text)
+    .map((chunk) => chunk.text)
+    .join("");
+
+  const result: any = {
```
jacoblee93: Try to avoid `any`, do you even need this cast?

js/src/wrappers/gemini.ts
```diff
@@ -0,0 +1,406 @@
...
+        input_tokens: usage.promptTokenCount || 0,
+        output_tokens:
+          usage.responseTokenCount || usage.candidatesTokenCount || 0,
+        total_tokens: usage.totalTokenCount || 0,
+      };
+
+      // Add input token details if available
+      usageMetadata.input_token_details = {
+        ...(usage.cachedContentTokenCount && {
+          cache_read_over_200k: Math.max(0, usage.promptTokenCount - 200000),
```
jacoblee93: The math here is a bit trickier I think - `cache_read_over_200k` and `over_200k` need to be mutually exclusive for our cost tracking to work out

e.g. if you had a 400k token prompt and 100k were cached the correct usage metadata would be:

```
{
  input_tokens: 400000,
  input_token_details: {
    cache_read_over_200k: 100000,
    over_200k: 300000,
  }
}
```

At least that's my read based on: https://ai.google.dev/gemini-api/docs/pricing

We could ask for confirmation there?

js/src/wrappers/gemini.ts
```diff
@@ -0,0 +1,406 @@
...
+    "content" in response &&
+    "role" in response &&
+    !("candidates" in response)
+  ) {
+    return response;
+  }
+
+  // Process raw Gemini response
+  let content = "";
+  const toolCalls: Array<any> = [];
```
jacoblee93: See above, try to avoid `any`

This is better if it works and you don't want to import the real type:
```
const toolCalls: Array<Record<string, unknown>> = [];
```

js/src/wrappers/gemini.ts
```diff
@@ -0,0 +1,406 @@
...
+  }
+
+  const result: KVMap = {
+    content: resultContent,
+    role: "assistant",
+  };
+
+  // Extract and store usage metadata
+  if ("usageMetadata" in response && response.usageMetadata) {
+    const usage = response.usageMetadata;
```
jacoblee93: Let's factor the usage metadata construction into a shared method

EugeneJinXin: yeah great idea should have done that

js/src/wrappers/gemini.ts
EugeneJinXin: fixed

js/src/wrappers/gemini.ts
dqbd: Decided to instead import the real type, since real types are omitted from the output while allowing us to rely on GenAI SDK to uncover potential issues

PR #2124, opened 2025-11-13
chore: Xin/improve unnamed runtree naming

EugeneJinXin:
https://langchain.slack.com/archives/C04GMBZ9Y1Y/p1763061096088929

### DIFF
--- a/python/langsmith/run_trees.py
+++ b/python/langsmith/run_trees.py
@@ -233,13 +233,53 @@ class RunTree(ls_schemas.RunBase):
     @root_validator(pre=True)
     def infer_defaults(cls, values: dict) -> dict:
         """Assign name to the run."""
+        # Try to get name from serialized data (legacy LangChain support)
         if values.get("name") is None and values.get("serialized") is not None:
-            if "name" in values["serialized"]:
-                values["name"] = values["serialized"]["name"]
-            elif "id" in values["serialized"]:
-                values["name"] = values["serialized"]["id"][-1]
+            serialized = values["serialized"]
+            values["name"] = serialized.get("name") or (
+                serialized["id"][-1] if "id" in serialized else None
+            )
+
+        # If still no name, try to infer from caller
         if values.get("name") is None:
-            values["name"] = "Unnamed"
+            name = None
+            try:
+                import inspect
+                import os
+
+                # Look up the stack, skipping internal frames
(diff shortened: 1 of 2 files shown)


python/langsmith/run_trees.py
```diff
@@ -233,13 +233,53 @@ class Config:
...
+            values["name"] = serialized.get("name") or (
+                serialized["id"][-1] if "id" in serialized else None
+            )
+
+        # If still no name, try to infer from caller
         if values.get("name") is None:
-            values["name"] = "Unnamed"
+            name = None
+            try:
+                import inspect
```
jacoblee93, line 247: Imports like this should be top level right?

python/langsmith/run_trees.py
```diff
@@ -233,13 +233,53 @@ class Config:
...
+                import os
+
+                # Look up the stack, skipping internal frames
+                for frame_info in inspect.stack()[1:8]:
```
jacoblee93, line 251: Any perf concerns here?

python/langsmith/run_trees.py
```diff
@@ -233,13 +233,53 @@ class Config:
...
+                    }:
+                        continue
+
+                    # Try to get class name from self or cls
+                    frame_locals = frame_info.frame.f_locals
+                    if "self" in frame_locals:
+                        name = frame_locals["self"].__class__.__name__
+                        break
+                    elif "cls" in frame_locals:
+                        name = frame_locals["cls"].__name__
```
jacoblee93, line 272: I don't know enough about Python to know if this is a good idea

Is there no easier fallback?

python/langsmith/run_trees.py
EugeneJinXin: it should be rare case we even need to enter "make a name" mode. so i don't worry too much about perf. 
yeah inspect can go to top

python/langsmith/run_trees.py
EugeneJinXin: it should be rare case we even need to enter "make a name" mode. i.e, this conditoin should not be entered.
  so i don't worry too much about perf.

python/langsmith/run_trees.py
EugeneJinXin: there's try catch so at least it won't error, no better fallback afaik

PR #2131, opened 2025-11-18
feat(py): Adds upper bound for tracing queue size

jacoblee93:


python/langsmith/__init__.py
```diff
@@ -23,7 +23,7 @@
 
 # Avoid calling into importlib on every call to __version__
 
-__version__ = "0.4.42"
+__version__ = "0.4.43"
```
jacoblee93, line 26: Seems this didn't get bumped alongside the version bump in `main`

python/langsmith/client.py
```diff
@@ -1443,6 +1455,67 @@ def _filter_for_sampling(
...
+
+        Returns:
+            True if the trace should be compressed, False if it should be dropped.
+        """
+        if self.compressed_traces is None:
+            return False
+
+        item_size = serialized_op.calculate_serialized_size()
+
+        # Allow the item if the buffer is empty (to support large single traces)
```
angus-langchain, line 1503: we won't allow traces over 200MB anyway, do we want to allow a single trace if its explicitly over the memory limit a user sets?

python/langsmith/client.py
```diff
@@ -1443,6 +1455,67 @@ def _filter_for_sampling(
...
+        if (
+            self.compressed_traces.uncompressed_size + item_size
+            > self._tracing_queue_size_limit_bytes
```
angus-langchain, line 1506: is _tracing_queue_size_limit_bytes global, compressed queue is tracked via `CompressedTraces.uncompressed_size`

python/langsmith/client.py
```diff
@@ -1443,6 +1455,67 @@ def _filter_for_sampling(
...
+                    f"Dropping run with id: {item.item.id}. "
+                    f"Current queue size: {self._tracing_queue_size_bytes} bytes, "
+                    f"attempted addition: {item_size} bytes."
+                )
+                return
+
+            self.tracing_queue.put(item)
+            self._tracing_queue_size_bytes += item_size
+
+    def _should_compress_trace(
```
angus-langchain, line 1486: do we need to take the lock here?

PR #2134, opened 2025-11-19
Fix:  Add with_raw_response support to OpenAI & Anthropic wrapper

victorm-lc:
Using `with_raw_response` to access HTTP headers was breaking traces - LangSmith showed the wrapper object instead of the actual response:

```python
raw_response = client.chat.completions.with_raw_response.create(...)
# Trace output: {"output": "<APIResponse [200 OK] type=...>"}  # useless
```

Now we call `.parse()` on the wrapper before logging, so you get the actual completion data in traces while still having access to headers.

### Changes:
- `_openai.py`: wrap `with_raw_response.create` for chat/completions, parse `APIResponse` in `_process_chat_completion`
- `_anthropic.py`: same thing for `messages.with_raw_response.create` with `LegacyAPIResponse`
- Added tests for both sync/async in `tests/integration_tests/wrappers/`

### Testing locally:

```python
# OpenAI
from dotenv import load_dotenv; load_dotenv()
import openai
from langsmith.wrappers import wrap_openai

client = wrap_openai(openai.Client())
raw = client.chat.completions.with_raw_response.create(
    model="gpt-4o-mini",
    messages=[{"role": "user", "content": "hi"}],
)
print(raw.headers.get('x-ratelimit-remaining-requests'))  # headers work
print(raw.parse().choices[0].message.content)  # parsing works
# check LangSmith - should show actual completion, not <APIResponse>
```

```python
# Anthropic
from dotenv import load_dotenv; load_dotenv()
import anthropic
from langsmith.wrappers import wrap_anthropic

client = wrap_anthropic(anthropic.Anthropic())
raw = client.messages.with_raw_response.create(
    model="claude-haiku-4-5",
    messages=[{"role": "user", "content": "hi"}],
)
print(raw.headers.get('request-id'))
print(raw.parse().content[0].text)
```

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "langsmith",
-  "version": "0.4.5",
+  "version": "0.4.7",
   "description": "Client library to connect to the LangSmith Observability and Evaluation Platform.",
   "packageManager": "yarn@1.22.19",
   "files": [
--- a/js/src/experimental/vercel/utils.ts
+++ b/js/src/experimental/vercel/utils.ts
@@ -1,9 +1,6 @@
 /* eslint-disable import/no-extraneous-dependencies */
-import type {
-  LanguageModelV2DataContent,
-  LanguageModelV2Message,
-} from "@ai-sdk/provider";
-import type { ModelMessage, ToolCallPart } from "ai";
+import type { LanguageModelV2DataContent } from "@ai-sdk/provider";
+import type { ToolCallPart } from "ai";
 
 const guessMimetypeFromBase64 = (data: string) => {
   // Check magic bytes from base64 data
@@ -122,7 +119,7 @@ export const normalizeFileDataAsDataURL = (
 };
 
 export const convertMessageToTracedFormat = (
-  message: LanguageModelV2Message | ModelMessage,
+  message: Record<string, unknown>,
   responseMetadata?: Record<string, unknown>
 ) => {
   const formattedMessage: Record<string, unknown> = {
--- a/js/src/index.ts
+++ b/js/src/index.ts
@@ -28,4 +28,4 @@ export {
 } from "./utils/prompts_cache.js";
 
 // Update using yarn bump-version
-export const __version__ = "0.4.5";
+export const __version__ = "0.4.7";
--- a/js/src/run_trees.ts
+++ b/js/src/run_trees.ts
@@ -165,6 +165,22 @@ type WriteReplica = {
 };
 type Replica = ProjectReplica | WriteReplica;
 
+const HEADER_SAFE_REPLICA_FIELDS = new Set([
+  "projectName",
+  "updates",
+  "reroot",
+]);
+
+function filterReplicaForHeaders(replica: WriteReplica): WriteReplica {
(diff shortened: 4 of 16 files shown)


python/langsmith/wrappers/_openai.py
```diff
@@ -302,6 +302,18 @@ def _create_usage_metadata(
...
     try:
+        # Check if outputs is an APIResponse wrapper (from with_raw_response)
+        # APIResponse has a .parsed attribute or .parse() method
+        if hasattr(outputs, "parsed") and hasattr(outputs.parsed, "model_dump"):
+            # Extract the parsed response from the wrapper
+            outputs = outputs.parsed
+        elif hasattr(outputs, "parse") and callable(outputs.parse):
+            # Some versions use .parse() method
+            try:
+                outputs = outputs.parse()
```
jacoblee93, line 316: This is a bit unclear, what versions use `.parse()`? Older OpenAI SDK versions?

python/langsmith/wrappers/_openai.py
```diff
@@ -492,6 +516,19 @@ def wrap_openai(
         process_outputs=_process_chat_completion,
     )
 
+    # Wrap with_raw_response.create for chat completions
+    if hasattr(client.chat.completions, "with_raw_response"):
+        client.chat.completions.with_raw_response.create = _get_wrapper(  # type: ignore[method-assign]
```
jacoblee93: Do we need to wrap `.parse` here too? And what about the responses API?

python/langsmith/wrappers/_openai.py
victorm-lc: added better comments but comes from here: github.com/openai/openai-python/blob/main/src/openai/_response.py#L285

python/langsmith/wrappers/_openai.py
victorm-lc: I don't think so, will remove now!

python/langsmith/wrappers/_anthropic.py
```diff
@@ -181,6 +181,16 @@ def _reduce_completions(all_chunks: list[Completion]) -> dict:
 
 def _process_chat_completion(outputs: Any):
     try:
+        # Check if outputs is a LegacyAPIResponse wrapper (from with_raw_response).
+        # The Anthropic SDK's LegacyAPIResponse wraps the actual response object.
+        # Call .parse() to extract the Message for tracing.
+        # See: anthropics/anthropic-sdk-python _legacy_response.py#L102
+        if hasattr(outputs, "parse") and callable(outputs.parse):
+            try:
+                outputs = outputs.parse()
```
jacoblee93, line 191: One thing I thought of here - I think most often when people use `.with_raw_response` it's because they want to defer this later for some reason

Won't this consume the body?

victorm-lc: Good catch, both the OpenAI and Anthropic SDKs cache the `.parse()` result in `_parsed_by_type`. So when we call it in `_process_chat_completion`, subsequent user calls to `.parse()` just return the cached value (same object). Verified locally that calling `.parse()` multiple times returns the same object each time.

PR #2139, opened 2025-11-20
Xin/js ci cd

EugeneJinXin:
1. Vitest/Jest Conflict Error (6 test files)
  TypeError: Cannot redefine property: Symbol($$jest-matchers-object)    at Function.defineProperty (<anonymous>)  at node_modules/@vitest/expect/dist/index.js:601:9
  Affected files:
  ```
  src/tests/otel/vercel/anthropic_cache_otel.int.test.ts
  src/tests/otel/vercel/openai_cache_otel.int.test.ts
  src/tests/otel/rate_limit_otel.int.test.ts
  src/tests/vercel/vercel_rate_limit.int.test.ts
  src/tests/vercel/anthropic_cache.int.test.ts
  src/tests/vercel/openai_cache.int.test.ts
  ```
  Root cause: Jest was loading these test files (even though they have describe.skip), which indirectly caused vitest's expect module to load and conflict with Jest's expect.

Fix: The testPathIgnorePatterns excludes these files entirely so Jest never tries to load them.

2. o1-mini deprecated -> fix
3. few shot "data already indexed" and the api is basically broken  --> skip 
4. broken readme link -> fix
5. ai_sdk_evaluate.int.test.ts - Import Error. Problem: Jest was trying to load a skipped test file with import issues. Fix is exclude it since already skipped.
6.  Test skipped due to rate limit (429). This is expected in CI with API rate limits.

### DIFF
--- a/python/tests/integration_tests/wrappers/test_openai.py
+++ b/python/tests/integration_tests/wrappers/test_openai.py
@@ -256,7 +256,7 @@ test_cases = [
     {
         "description": "reasoning",
         "params": {
-            "model": "o1-mini",
+            "model": "o4-mini",
             "messages": [
                 {
                     "role": "user",


js/src/tests/wrapped_openai.int.test.ts
```diff
@@ -947,7 +947,7 @@ const usageMetadataTestCases = [
   {
     description: "reasoning",
     params: {
-      model: "o1-mini",
+      model: "o1",
```
jacoblee93, line 950: Let's use `o4-mini`!

jacoblee93 (approved):
Very nice thank you!

PR #2156, opened 2025-11-25
fix(js): add reporter support for Vitest 4.x

dqbd:
Vitest 4.x removes `onFinished` method in favour of `onTestRunEnd` method


js/src/vitest/reporter.mts
```diff
@@ -11,6 +15,16 @@ class LangSmithEvalReporter extends DefaultReporter {
     super.onFinished(files, errors);
     await printVitestReporterTable(files, this.ctx);
   }
+
+  // `onFinished` is removed in Vitest 4.x, so we use `onTestRunEnd` instead.
+  async onTestRunEnd(
```
jacoblee93, line 26: Is this a straight rename? Will this run twice on Vitest 3.x?

PR #2166, opened 2025-12-02
feat: Use uuid-utils

hinthornw:


python/tests/unit_tests/test_uuid_v7.py
```diff
@@ -42,6 +44,7 @@ def fn(x: int, *, run_tree: RunTree) -> int:  # type: ignore[misc]
     assert id_ms == start_ms
 
 
+@pytest.mark.repeat(100)
```
angus-langchain: why are we repeating this test?

angus-langchain (approved):

python/tests/unit_tests/test_uuid_v7.py
hinthornw: ha was just trying to be more rigorous locally but we don't need to do this in ci

hinthornw: removed

PR #2189, opened 2025-12-09
remove duplicate patch from claude agent sdk

angus-langchain:
Removes unnecessary duplicate patch. patch is called on close.

### DIFF
--- a/python/langsmith/integrations/claude_agent_sdk/_client.py
+++ b/python/langsmith/integrations/claude_agent_sdk/_client.py
@@ -62,10 +62,6 @@ class TurnLifecycle:
             "usage_metadata", {}
         )
         meta.update(metrics)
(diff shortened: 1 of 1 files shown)


python/langsmith/integrations/claude_agent_sdk/_client.py
```diff
@@ -62,10 +62,6 @@ def add_usage(self, metrics: dict[str, Any]) -> None:
             "usage_metadata", {}
         )
         meta.update(metrics)
-        try:
-            self.current_run.patch()
-        except Exception as e:
-            logger.warning(f"Failed to update usage metrics: {e}")
 
     def close(self) -> None:
```
angus-langchain, line 70: patch already called here on close()

PR #2228, opened 2025-12-22
fix(js): Fix JS Memory Leak clean child_runs

### LINKED ISSUE #2227: Issue: Memory Leaks in JS Library
Hey Team, I want to ask if there's a possible memory leak in the JS LangSmith library when using traceable

We've been seeing memory grow continuously in our production Node.js service and think there might be a memory leak in the library when using nested traces.

## What We Were Doing

We have a service wrapper that adds tracing to our operations:

```typescript
// Called on every request
async trace(fn, input, options) {
  const tracedFn = traceable(
    async (input) => fn(input),
    {
      name: options.name,
      run_type: options.run_type,
      tags: options.tags,
      metadata: options.metadata,
    }
  );
  
  return tracedFn(input);
}
```

Each request creates a parent trace with a few child traces. After running for a while, memory grows until we eventually OOM.

### Library Bug - child_runs Never Cleared

In `run_trees.ts` line 442, when you create a child trace it gets pushed to the parent's `child_runs` array:

```typescript
this.child_runs.push(child);
```

This array is never cleared - not after `postRun()` and not after `patchRun()`. Every parent RunTree keeps references to all its children forever, preventing garbage collection.

We wrote a test to verify this and saw significant memory growth from child references that should have been released.

### Our Pattern - Creating Wrappers Per Request

We also realized we weren't following the documented pattern. The docs show you should create the wrapper once and reuse it:

```typescript
// Create ONCE
const tracedFn = traceable(fn, { name, run_type });

// Reuse many times
await tracedFn(input1);
await tracedFn(input2);
```

But we were creating a new wrapper on every request, which adds overhead and makes the library issue worse.

## What We Did to Fix It

We refactored to use RunTree directly with explicit lifecycle management:

```typescript
async trace(fn, input, options) {
  const parentRun = getCurrentRunTree(true);
  
  const runTree = new RunTree({
    name: options.name,
    run_type: options.run_type,
    parent_run: parentRun,
    inputs: { input },
  });
  
  await runTree.postRun();
  
  try {
    const result = await withRunTree(runTree, () => fn(input));
    await runTree.end({ output: result });
    await runTree.patchRun();
    return result;
  } catch (error) {
    await runTree.end(undefined, String(error));
    await runTree.patchRun();
    throw error;
  }
}
```

This works for us, but the library should still handle the child_runs cleanup.

## Proposed Library Fix

Simple fix - clear the child_runs array after posting:

```typescript
async postRun() {
  // ... existing posting code ...
  this.child_runs = [];
}

async patchRun() {
  // ... existing posting code ...
  this.child_runs = [];
}
```

We tested this and it significantly reduces memory growth and completely eliminates the child reference leak.

## Why This Matters

The library bug affects everyone using nested traces, regardless of usage pattern. Even users following the documented pattern correctly will see some memory leak, though it's much worse if you create traceable wrappers per request.

We tested different combinations and the library fix helps in all cases. The best results come from both fixing the library and following the correct pattern.

## Connection to Python SDK

The Python SDK had the same issue and fixed it a few months ago:
- Issue #2009 (Sept 2025)
- PR #2033 (Sept 22, 2025) - cleared child_runs after posting
- Issue #2097 (Oct 2025)
- PR #2107 (Nov 3, 2025)

Same problem - child_runs accumulation.

Thanks

amodelaweb:
Fixes #2227

Hey! This PR fixes the memory leak we found where `child_runs` was never getting cleared.

## The Fix

Super simple - just clear the `child_runs` array after we're done posting the runs to LangSmith. Added two lines:

```typescript
// In js/src/run_trees.ts

async postRun() {
  // ... existing code that posts the run ...
  this.child_runs = [];  // Line 775
}

async patchRun() {
  // ... existing code that patches the run ...
  this.child_runs = [];  // Line 863
}
```

Once the runs are posted, we don't need to keep those child references around anymore. This lets them get garbage collected properly.

## Testing

Added unit tests in `js/src/tests/run_trees.test.ts` to make sure the array gets cleared. All existing tests still pass.

We tested this in our production-like scenario and it fixed the memory growth we were seeing - went from continuously growing memory to stable usage.

## Why It's Safe

The `child_runs` array is only used during trace construction (for the replica feature). After we post to LangSmith, we don't need those references anymore, so clearing them is totally safe.

This is the same fix the Python SDK got a few months ago for the same issue (#2033, #2107).

---

Let me know if you need anything else! Happy to adjust the approach or add more tests.

### DIFF
--- a/js/src/run_trees.ts
+++ b/js/src/run_trees.ts
@@ -771,7 +771,7 @@ export class RunTree implements BaseRun {
           await childRun.postRun(false);
         }
       }
-      
+
       this.child_runs = [];
     } catch (error) {
       console.error(`Error in postRun for run ${this.id}:`, error);
@@ -859,7 +859,7 @@ export class RunTree implements BaseRun {
         console.error(`Error in patchRun for run ${this.id}`, error);
       }
     }
-    
+
     this.child_runs = [];
   }
 


js/src/run_trees.ts
```diff
@@ -857,6 +859,8 @@ export class RunTree implements BaseRun {
         console.error(`Error in patchRun for run ${this.id}`, error);
       }
     }
+
+    this.child_runs = [];
```
jacoblee93, line 863: I don't actually think this is necessary but I guess you could have some weird cases where you create a child after post

amodelaweb: Yes !, I added for that specific case. I also think we could just remove the push from the child_runs (as the Python SDK does right now) however I just decided to add the clean in those two code paths to avoid breaking anything.

PR #2232, opened 2025-12-22
fix(js): Avoid waiting for child runs to complete if parent throws an error

### LINKED ISSUE #2229: Traceable function never throws (AI SDK)
If code inside `traceable` throws while LLM processing request traceable function waits until LLM finishes (most probably it waits until child run is finished), which is not expected behaviour.

Code example:
```
import { createAnthropic } from '@ai-sdk/anthropic';
import * as ai from 'ai';
import { wrapAISDK } from 'langsmith/experimental/vercel';
import { Client } from 'langsmith';
import { traceable } from 'langsmith/traceable';

const { streamText } = wrapAISDK(ai);
const langsmithClient = new Client();

const anthropicProvider = createAnthropic({ apiKey: process.env.ANTHROPIC_API_KEY });
const tracedOperation = traceable(
    async () => {
          const { fullStream } = streamText({
                model: anthropicProvider('claude-sonnet-4-5'),
                messages: [{ role: 'user', content: 'Hello, world!' }],
          });

          throw new Error('TEST');

          for await (const chunk of fullStream) {
               console.log(chunk);
          }
    },
    { name: 'test-trace', client: langsmithClient },
);

try {
    await tracedOperation();
} catch (error) {
    // control will be here not after error is thrown but after LLM response stream is completed
}
```

Most probably this behaviour was introduced in this PR - https://github.com/langchain-ai/langsmith-sdk/pull/2094

jacoblee93:
Fixes #2229

### DIFF
--- a/js/src/traceable.ts
+++ b/js/src/traceable.ts
@@ -168,7 +168,13 @@ async function handleEnd(params: {
   deferredInputs?: boolean;
   skipChildPromiseDelay?: boolean;
 }) {
-  const { runTree, on_end, postRunPromise, deferredInputs, skipChildPromiseDelay } = params;
+  const {
+    runTree,
+    on_end,
+    postRunPromise,
+    deferredInputs,
+    skipChildPromiseDelay,
+  } = params;
   const onEnd = on_end;
   if (onEnd) {
     onEnd(runTree);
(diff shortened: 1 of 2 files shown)


js/src/traceable.ts
```diff
@@ -1130,21 +1145,14 @@ export function traceable<Func extends (...args: any[]) => any>(
               }
             },
             async (error: unknown) => {
-              if (
```
jacoblee93, line 1133: We should not have been awaiting this directly in code here to begin with - it should have just been handled by `handleEnd`.

PR #2246, opened 2025-12-28
feat: add prompt caching to python sdk

langchain-infra:
## Add Prompt Caching for Python SDK

  ### Summary

  Adds an optional in-memory caching layer for `pull_prompt` and `pull_prompt_commit` methods in the Python SDK. The cache uses a stale-while-revalidate (SWR) pattern with background refresh, reducing API calls and improving latency for prompt lookups.

  ### Features

  - **In-memory LRU cache** with configurable max size (default: 100 entries)
  - **Background refresh** - Stale entries are refreshed asynchronously while serving cached data
  - **Offline mode** - Set `ttl_seconds=None` for infinite TTL, load from file on init
  - **Cache metrics** - Track hits, misses, refresh counts, and errors
  - **Thread-safe** for concurrent access
  - **Error-resilient** - Refresh loops recover from errors and continue running

  ### Usage

  ```python
  from langsmith import Client

  # Enable caching via constructor
  client = Client(
      prompt_cache_enabled=True,
      prompt_cache_max_size=100,
      prompt_cache_ttl_seconds=3600,  # 1 hour
  )

  # Or via environment variables
  # LANGSMITH_PROMPT_CACHE_ENABLED=true
  # LANGSMITH_PROMPT_CACHE_MAX_SIZE=100
  # LANGSMITH_PROMPT_CACHE_TTL_SECONDS=3600

  # Use normally - caching is transparent
  prompt = client.pull_prompt("my-org/my-prompt")

  # Check cache metrics
  print(client._prompt_cache.metrics.hit_rate)

  # Save cache for offline use
  client._prompt_cache.dump("prompts.json")

  # Later: load from file on init with infinite TTL for offline mode
  offline_client = Client(
      prompt_cache_enabled=True,
      prompt_cache_ttl_seconds=None,  # Never expire
      prompt_cache_path="prompts.json",  # Load on init
  )
```

Closes INF-1336

### DIFF
--- a/python/tests/unit_tests/test_prompt_cache.py
+++ b/python/tests/unit_tests/test_prompt_cache.py
@@ -153,7 +153,6 @@ class TestPromptCacheBackgroundRefresh:
             ttl_seconds=0.1,  # Very short TTL
             refresh_interval_seconds=0.2,
             fetch_func=mock_fetch,
-            
         )
         try:
             cache.set("test-key", sample_prompt_commit)
@@ -179,7 +178,6 @@ class TestPromptCacheBackgroundRefresh:
             ttl_seconds=0.1,
             refresh_interval_seconds=0.2,
             fetch_func=failing_fetch,
-            
         )
         try:
             cache.set("test-key", sample_prompt_commit)
@@ -203,7 +201,6 @@ class TestPromptCacheBackgroundRefresh:
             ttl_seconds=0.1,
             refresh_interval_seconds=0.1,
             # No fetch_func
-            
         )
         try:
             cache.set("test-key", sample_prompt_commit)
@@ -225,7 +222,6 @@ class TestPromptCacheBackgroundRefresh:
             ttl_seconds=0.1,
             refresh_interval_seconds=0.1,
             fetch_func=mock_fetch,
-            
         )
 
         # Verify thread is running
@@ -307,7 +303,6 @@ class TestAsyncPromptCache:
             ttl_seconds=0.1,  # Very short TTL
             refresh_interval_seconds=0.2,
             fetch_func=mock_fetch,
-            
         )
         try:
             cache.set("test-key", sample_prompt_commit)
@@ -337,7 +332,6 @@ class TestAsyncPromptCache:
             ttl_seconds=0.1,
             refresh_interval_seconds=0.2,
             fetch_func=failing_fetch,
-            
         )
         try:
             cache.set("test-key", sample_prompt_commit)
@@ -367,7 +361,6 @@ class TestAsyncPromptCache:
             ttl_seconds=0.1,
             refresh_interval_seconds=0.1,
             fetch_func=mock_fetch,
-            
         )
 
         await cache.start()
@@ -388,7 +381,6 @@ class TestAsyncPromptCache:
             ttl_seconds=0.1,
             refresh_interval_seconds=0.1,
             # No fetch_func
-            
         )
 
         await cache.start()
@@ -576,7 +568,6 @@ class TestOfflineMode:
         cache = PromptCache(
             ttl_seconds=None,
             fetch_func=mock_fetch,
-            
         )
         try:
             # No refresh thread should be started
@@ -626,7 +617,6 @@ class TestOfflineMode:
         cache = AsyncPromptCache(
             ttl_seconds=None,
             fetch_func=mock_fetch,
-            
         )
 
         await cache.start()
@@ -690,7 +680,6 @@ class TestCacheMetrics:
             ttl_seconds=0.1,
             refresh_interval_seconds=0.2,
             fetch_func=mock_fetch,
-            
         )
         try:
             cache.set("key1", sample_prompt_commit)
@@ -713,7 +702,6 @@ class TestCacheMetrics:
             ttl_seconds=0.1,
             refresh_interval_seconds=0.2,
             fetch_func=failing_fetch,
-            
         )
         try:
             cache.set("key1", sample_prompt_commit)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: python/langsmith/_internal/_cache.py [hunk 1]
```diff
@@ -0,0 +1,532 @@
...
+                    loaded += 1
+                except Exception as e:
+                    logger.warning(f"Failed to load cache entry {key}: {e}")
+                    continue
+
+        logger.debug(f"Loaded {loaded} cache entries from {path}")
+        return loaded
+
+
+class PromptCache(_BasePromptCache):
```
langchain-infra, line 274: wasn't sure if this should be one class with async methods combined with sync but split for now.

python/langsmith/utils.py
```diff
@@ -370,9 +371,29 @@ def is_base_message_like(obj: object) -> bool:
 
 def is_env_var_truish(value: Optional[str]) -> bool:
     """Check if the given environment variable is truish."""
+    if value is None:
+        return False
     return is_truish(get_env_var(value))
 
 
+@overload
```
langchain-infra, line 379: if default is specified, we should always return a str. this is strictly for mypy.

python/langsmith/client.py
```diff
@@ -702,6 +704,11 @@ def __init__(
         max_batch_size_bytes: Optional[int] = None,
         headers: Optional[dict[str, str]] = None,
         tracing_error_callback: Optional[Callable[[Exception], None]] = None,
+        prompt_cache_enabled: Optional[bool] = None,
```
langchain-infra, line 707: default to disabled until we do another minor release

python/langsmith/_internal/_cache.py
```diff
@@ -0,0 +1,532 @@
...
+
+        Args:
+            key: The cache key (prompt identifier like "owner/name:hash").
+
+        Returns:
+            The cached value or None if not found.
+            Stale entries are still returned (background refresh handles updates).
+        """
+        if not self._enabled:
+            self._metrics.misses += 1
```
angus-langchain: Does this need to be thread safe? Right now it is not atomic

python/langsmith/async_client.py
```diff
@@ -1640,22 +1699,56 @@ async def delete_prompt(self, prompt_identifier: str) -> None:
         )
         response.raise_for_status()
 
-    async def pull_prompt_commit(
+    def _get_prompt_cache_key(
```
angus-langchain, line 1701: Did you want to expose `use_cache` param here?

python/langsmith/_internal/_cache.py
langchain-infra: actually i can just remove the enabled within the cache. we can treat it's existence as being enabled.

python/langsmith/async_client.py
langchain-infra: that would be in this func right? 

```
    async def pull_prompt_commit(
        self,
        prompt_identifier: str,
        *,
        include_model: Optional[bool] = False,
        use_cache: bool = True,
```

python/langsmith/_internal/_cache.py
```diff
@@ -0,0 +1,508 @@
...
+        self._shutdown_event.set()
+        if self._refresh_thread is not None:
+            self._refresh_thread.join(timeout=5.0)
+            self._refresh_thread = None
+
+    def _start_refresh_thread(self) -> None:
+        """Start background thread for refreshing stale entries."""
+        self._refresh_thread = threading.Thread(
+            target=self._refresh_loop,
+            daemon=True,
```
angus-langchain, line 346: probably not a large issue, but this could be a memory leak if I create a bunch of clients in app

python/langsmith/_internal/_cache.py
```diff
@@ -0,0 +1,508 @@
...
+        return self.hits + self.misses
+
+    @property
+    def hit_rate(self) -> float:
+        """Cache hit rate (0.0 to 1.0)."""
+        total = self.total_requests
+        return self.hits / total if total > 0 else 0.0
+
+
+class _BasePromptCache(ABC):
```
angus-langchain, line 62: does this need to be an abstract base class?

python/langsmith/_internal/_cache.py
```diff
@@ -0,0 +1,508 @@
...
+import threading
+import time
+from abc import ABC
+from collections import OrderedDict
+from collections.abc import Awaitable
+from dataclasses import dataclass
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, Callable, Optional, Union
+
+if TYPE_CHECKING:
```
angus-langchain, line 21: ?

python/langsmith/async_client.py
angus-langchain: sorry meant here: https://github.com/langchain-ai/langsmith-sdk/blob/062e5bd53c200c565f03e14fea8ee6acfeafc5a9/python/langsmith/client.py#L8234 https://github.com/langchain-ai/langsmith-sdk/blob/062e5bd53c200c565f03e14fea8ee6acfeafc5a9/python/langsmith/async_client.py#L1863

python/langsmith/_internal/_cache.py
langchain-infra: i guess is there anyway around that? wouldn't that be the same for tracing?

python/langsmith/_internal/_cache.py
langchain-infra: sorry from refactoring, was using ls_schemas before

python/langsmith/_internal/_cache.py
langchain-infra: from refeactor, removed

angus-langchain (approved):

Issue #2029, opened 2025-09-18
pydantic v2 compatibility


darinkishore:
### Feature request

An upgrade to pydantic V2

### Motivation

it is hard to use langsmith's schemas and objects directly in most modern python codebases because most people are on pydantic v2. are you guys open to an upgrade?

angus-langchain:
@darinkishore Will take a look early next week, thanks for the PR

tnorlund:
Any update on this? Really looking forward to moving everything to 3.14.

jacoblee93:
Hey folks, yes this is in progress here: https://github.com/langchain-ai/langsmith-sdk/pull/2241

PR #2248, opened 2025-12-28
feat(python): Bump pydantic to v2

angus-langchain:
## Changes
  - Migrate from Pydantic v1 to v2 syntax
  - Update `class Config` → `model_config = ConfigDict(...)`
  - Update validators: `@validator`/`@root_validator` → `@field_validator`/`@model_validator`
  - Update `.dict()` → `.model_dump()`, `.update_forward_refs()` → `.model_rebuild()`
  - Update Field params: `min_items`/`max_items` → `min_length`/`max_length`
  - Bump minimum pydantic version to `>=2,<3`

  ## Test plan
  - [x] Existing tests pass
  - [x] Manual
  
## Todo

* [x] Update `LangChainTracer` to use Pydantic v2 syntax (with minimum LS version)
  * [x] `run.dict(...)` → `run.model_dump(...)`
  * [x] `Run.construct(...)` → `Run.model_construct(...)`

Resolves https://github.com/langchain-ai/langsmith-sdk/issues/2029

### DIFF
--- a/python/langsmith/async_client.py
+++ b/python/langsmith/async_client.py
@@ -737,7 +737,7 @@ class AsyncClient:
         run_id: Optional[ls_client.ID_TYPE],
         key: str,
         score: Optional[float] = None,
-        value: Optional[Any] = None,
+        value: Union[float, int, bool, str, dict, None] = None,
         comment: Optional[str] = None,
         **kwargs: Any,
     ) -> ls_schemas.Feedback:
@@ -748,7 +748,7 @@ class AsyncClient:
                 Can be None for project-level feedback.
             key (str): The name of the metric or aspect this feedback is about.
             score (Optional[float]): The score to rate this run on the metric or aspect.
-            value (Optional[Any]): The display value or non-numeric value for this feedback.
+            value (Union[float, int, bool, str, dict, None]): The display value or non-numeric value for this feedback.
             comment (Optional[str]): A comment about this feedback.
             **kwargs: Additional keyword arguments to include in the feedback data.
 


python/langsmith/client.py
```diff
@@ -9309,13 +9311,8 @@ def _construct_url(api_url: str, pathname: str) -> str:
...
 def dump_model(model, *, exclude_none: bool = False) -> dict[str, Any]:
-    """Dump model depending on pydantic version."""
-    if hasattr(model, "model_dump"):
-        return model.model_dump(exclude_none=exclude_none)
-    elif hasattr(model, "dict"):
-        return model.dict(exclude_none=exclude_none)
-    else:
-        raise TypeError("Unsupported model type")
+    """Dump model using Pydantic v2."""
+    return model.model_dump(exclude_none=exclude_none)
```
hinthornw: Why not support both ?

python/langsmith/run_trees.py
```diff
@@ -206,7 +197,7 @@ class RunTree(ls_schemas.RunBase):
     parent_dotted_order: Optional[str] = Field(default=None, exclude=True)
     child_runs: list[RunTree] = Field(
         default_factory=list,
-        exclude={"__all__": {"parent_run_id"}},
+        exclude=True,
```
hinthornw, line 200: Note this is a slight behavior change

angus-langchain: yup, we already exclude child_runs when doing `model_dump`, but I am planning on releasing this change in a minor bump

angus-langchain: we also clear this list after posting the child runs, so they aren't guaranteed to be in the tree anyways. For fetching a single run via `read_run` we return a Run and not a RunTree

hinthornw (approved):

python/langsmith/schemas.py
```diff
@@ -787,37 +758,37 @@ class TracerSessionResult(TracerSession):
     Sessions are also referred to as "Projects" in the UI.
     """
 
-    run_count: Optional[int]
+    run_count: Optional[int] = None
```
baskaryan, line 761: ooc why do we have to do this?

angus-langchain: https://docs.pydantic.dev/latest/migration/

_"A field annotated as Optional[T] will be required, and will allow for a value of None. It does not mean that the field has a default value of None. (This is a breaking change from V1.)"_

PR #2249, opened 2025-12-28
feat: Replace UUID5 with deterministic UUID7 for replicas

angus-langchain:
### Description
  Replaces uuid5 with `uuid7_deterministic` function for remapping run ids when tracing with replicas.

### Changes
```
  Original UUID7:  [timestamp][version 7][rand_a][variant][rand_b]
                       │
                       ▼ preserved if input is UUID7
  Derived UUID7:   [timestamp][version 7][hash_a][variant][hash_b]
                                           │               │
                                           └───────────────┘
                                             SHA-256(original:key)
```
  - UUID7 input: timestamp preserved from original
  - Non-UUID7 input: fresh timestamp generated
  - Random bits are derived deterministically from SHA256 hash

  ### Implementation References
  Timestamp byte packing matches official implementations:
  - Python: https://github.com/python/cpython/blob/3.14/Lib/uuid.py - time.time_ns() // 1_000_000 with to_bytes(6, "big")
  - JS: https://github.com/uuidjs/uuid/blob/main/src/v7.ts - Date.now() with division by powers of 256
  
  ### TODO
- [ ] Update LangGraph distributed runtime

### DIFF
--- a/js/src/utils/_uuid.ts
+++ b/js/src/utils/_uuid.ts
@@ -103,9 +103,8 @@ function bytesToUuid(bytes: Buffer): string {
  * - Uses current time if the original is not UUID v7
  * - Uses deterministic "random" bits derived from hashing the original + key
  * - Is valid UUID v7 format
- * - Produces the same output for the same inputs (deterministic)
  *
- * This is useful for creating replica IDs that maintain time-ordering properties
+ * This is used for creating replica IDs that maintain time-ordering properties
  * while being deterministic across distributed systems.
  *
  * @param originalId - The source UUID string (ideally UUID v7 to preserve timestamp)
--- a/js/src/tests/run_trees.uuid.test.ts
+++ b/js/src/tests/run_trees.uuid.test.ts
@@ -93,7 +93,6 @@ test("uuid7Deterministic produces valid, deterministic UUID7s", () => {
   // Valid UUID7
   expect(getUuidVersion(d1)).toBe(7);
 
-  // Deterministic (same input -> same output, critical for distributed tracing)
   expect(d1).toBe(d2);
 
   // Different inputs -> different outputs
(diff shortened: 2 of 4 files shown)


js/src/utils/_uuid.ts
```diff
@@ -70,3 +71,101 @@ export function warnIfNotUuidV7(uuidStr: string, _idType: string): void {
...
+  b[6] = 0x70 | (h[0] & 0x0f);
+
+  // rand_a continued (8 bits from hash)
+  b[7] = h[1];
+
+  // Set variant (10) in high 2 bits + 6 bits from hash
+  b[8] = 0x80 | (h[2] & 0x3f);
+
+  // rand_b (56 bits = 7 bytes from hash)
+  h.copy(b, 9, 3, 10);
```
jacoblee93, line 168: This looks divergent from:

https://github.com/uuidjs/uuid/blob/main/src/v7.ts#L107

Is it?

Also OOC why take these bits specifically instead of e.g. the first 7?

js/src/utils/_uuid.ts
```diff
@@ -70,3 +71,101 @@ export function warnIfNotUuidV7(uuidStr: string, _idType: string): void {
...
+
+  const b = Buffer.alloc(16);
+
+  // Check if original is UUID v7 - if so, preserve its timestamp
+  // If not, use current time to ensure the derived UUID has a valid timestamp
+  const version = getUuidVersion(originalId);
+  if (version === 7) {
+    // Preserve timestamp from original UUID7 (bytes 0-5)
+    const originalBytes = uuidToBytes(originalId);
+    originalBytes.copy(b, 0, 0, 6);
```
jacoblee93, line 144: nit: I assume this syntax is correct but it's pretty opaque - any other way we can do this?

js/src/utils/_uuid.ts
angus-langchain: it's the same structure, just using hash bytes instead of random bytes.

If we reused h[0:7] for b[9:16], we'd have correlation between rand_a and rand_b portions.

js/src/utils/_uuid.ts
angus-langchain: bit manipulation for ya, I could add a comment though.`h.copy(b, 9, 3, 10); // copy h[3..9] to b[9..15]`

Alternative would be doing this 
```
  b[9] = h[3];
  b[10] = h[4];
  b[11] = h[5];
  b[12] = h[6];
  b[13] = h[7];
  b[14] = h[8];
  b[15] = h[9];
```

PR #2251, opened 2025-12-29
feat: add js prompt caching

langchain-infra:
Adds an optional in-memory caching layer for `pullPromptCommit` method in the JavaScript SDK. The cache uses a stale-while-revalidate (SWR) pattern with background refresh, reducing API calls and improving latency for prompt lookups.

  ### Features

  - **In-memory LRU cache** with configurable max size (default: 100 entries)
  - **Background refresh** - Stale entries are refreshed asynchronously while serving cached data
  - **Offline mode** - Set `ttlSeconds=null` for infinite TTL, load from file on init
  - **Cache metrics** - Track hits, misses, refresh counts, and errors
  - **Cross-SDK compatibility** - Cache files are compatible with Python SDK

  ### Usage

  ```typescript

  // Enable caching via constructor
  const client = new Client({
    promptCacheEnabled: true,
    promptCacheMaxSize: 100,
    promptCacheTtlSeconds: 3600, // 1 hour
  });

  // Or via environment variables
  // LANGSMITH_PROMPT_CACHE_ENABLED=true
  // LANGSMITH_PROMPT_CACHE_MAX_SIZE=100
  // LANGSMITH_PROMPT_CACHE_TTL_SECONDS=3600

  // Use normally - caching is transparent
  const prompt = await client.pullPromptCommit("my-prompt", { includeModel: true });

  // Check cache metrics
  console.log(client.promptCache?.metrics);
  console.log(client.promptCache?.hitRate);

  // Save cache for offline use
  client.promptCache?.dump("prompts.json");

  // Later: load from file on init with infinite TTL for offline mode
  const offlineClient = new Client({
    promptCacheEnabled: true,
    promptCacheTtlSeconds: null, // Never expire
    promptCachePath: "prompts.json", // Load on init
  });

  // Cleanup when done
  client.cleanup();
```

Closes INF-1339

### DIFF
--- a/js/src/tests/prompt_cache.test.ts
+++ b/js/src/tests/prompt_cache.test.ts
@@ -274,7 +274,7 @@ describe("PromptCache", () => {
       // Allow the async refresh to complete
       await Promise.resolve();
 
-      expect(cache.metrics.refreshErrors).toBeGreaterThanOrEqual(0);
+      expect(cache.metrics.refreshErrors).toBeGreaterThan(0);
       cache.stop();
     });
 
--- a/js/src/utils/prompts_cache.ts
+++ b/js/src/utils/prompts_cache.ts
@@ -332,9 +332,10 @@ export class PromptCache {
         const newValue = await this.fetchFunc(key);
         this.set(key, newValue);
         this._metrics.refreshes += 1;
-      } catch {
+      } catch (e) {
         // Keep stale data on refresh failure
         this._metrics.refreshErrors += 1;
+        console.warn(`Failed to refresh cache entry ${key}:`, e);
       }
     }
   }

### CODE REVIEW

File: js/src/utils/prompts_cache.ts [hunk 1]
```diff
@@ -0,0 +1,341 @@
...
+      return;
+    }
+
+    for (const key of staleKeys) {
+      try {
+        const newValue = await this.fetchFunc(key);
+        this.set(key, newValue);
+        this._metrics.refreshes += 1;
+      } catch {
+        // Keep stale data on refresh failure
```
angus-langchain, line 336: this is completely skipped, should we at least add a warn
```js
console.warn(`Failed to refresh cache entry ${key}:`, e);
```

js/src/tests/prompt_cache.test.ts
```diff
@@ -0,0 +1,342 @@
...
+
+      cache.set("key1", createMockPromptCommit("test1"));
+
+      // Advance past TTL and refresh interval
+      jest.advanceTimersByTime(2000);
+
+      // Allow the async refresh to complete
+      await Promise.resolve();
+
+      expect(cache.metrics.refreshErrors).toBeGreaterThanOrEqual(0);
```
angus-langchain: this should be `.toBeGreaterThan(0)`

angus-langchain (approved):

js/src/utils/prompts_cache.ts
langchain-infra: added a warning log.

PR #2260, opened 2025-12-31
feat: add prompt cache back and setup environment tests

langchain-infra:
Copy the langgraph-js environments test. Also re-add prompt cache.

### DIFF
--- a/js/src/utils/prompts_cache_fs.browser.ts
+++ b/js/src/utils/prompts_cache_fs.browser.ts
@@ -13,19 +13,13 @@ export function dumpCache(
   _filePath: string,
   _entries: Record<string, unknown>
 ): void {
-  throw new Error(
-    "dump() is not supported in browser environments."
-  );
+  throw new Error("dump() is not supported in browser environments.");
 }
 
 /**
  * Load cache entries from a JSON file.
  * @throws Error - Always throws in browser environments.
  */
-export function loadCache(
-  _filePath: string
-): Record<string, unknown> | null {
-  throw new Error(
-    "load() is not supported in browser environments."
-  );
+export function loadCache(_filePath: string): Record<string, unknown> | null {
+  throw new Error("load() is not supported in browser environments.");
 }
--- a/js/src/utils/prompts_cache_fs.ts
+++ b/js/src/utils/prompts_cache_fs.ts
@@ -41,9 +41,7 @@ export function dumpCache(
  *
  * @returns The entries object, or null if file doesn't exist or is invalid.
  */
-export function loadCache(
-  filePath: string
-): Record<string, unknown> | null {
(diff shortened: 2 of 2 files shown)


js/src/utils/prompts_cache.ts
```diff
@@ -0,0 +1,366 @@
...
+  /**
+   * Load cache contents from a JSON file.
+   *
+   * Loaded entries get a fresh TTL starting from load time.
+   *
+   * Note: This method requires Node.js and will throw in browser/edge environments.
+   *
+   * @returns Number of entries loaded.
+   */
+  async load(filePath: string): Promise<number> {
```
langchain-infra: main difference is the dynamic imports here

js/src/utils/prompts_cache.ts
```diff
@@ -0,0 +1,366 @@
...
+    let fs: typeof import("node:fs");
+    let path: typeof import("node:path");
+    try {
+      fs = await import("node:fs");
+      path = await import("node:path");
+    } catch {
+      throw new Error(
+        "dump() requires Node.js. File operations are not supported in browser/edge environments."
+      );
+    }
```
hntrl: Bundlers will still try to bundle dynamic imports like this into web envs even though it's conditional. 

Think what we have to do is either (1) hoist this util to a new file and have it as its own "node-only" entrypoint (or have some fs "adapter" pattern), or (2) don't have a 'dump' function but still have some way to render the file contents and defer to the user to store it in fs

js/internal/environment_tests/docker-compose.yml
```diff
@@ -0,0 +1,34 @@
+version: "3"
```
hntrl, line 1: Think we'll also want to add a 'vote' test for web env targets. IIRC we have one for both lcjs and lgjs

langchain-infra: added

js/src/utils/prompts_cache.ts
langchain-infra: updated to use a separate entrypoint

js/internal/environment_tests/test-exports-vite/src/index.ts
```diff
@@ -0,0 +1,25 @@
...
+ * This test verifies:
+ * 1. Vite can build langsmith for browser targets
+ * 2. The browser field in package.json correctly swaps Node-specific code
+ * 3. Client export is available (uses fetch, works in browsers)
+ *
+ * Note: traceable and RunTree use AsyncLocalStorage from node:async_hooks
+ * which doesn't have a browser equivalent, so they're not tested here.
+ */
+
+import { Client } from "langsmith";
```
langchain-infra, line 14: can't export runTree here as it fails

hntrl (approved):

PR #2264, opened 2026-01-05
feat: Pass in Cache, rename

langchain-infra:
Simplify initialization and allow passing in Cache. Also enables overriding this for your own impl if you want to bring a different store like Redis

### DIFF
--- a/python/langsmith/async_client.py
+++ b/python/langsmith/async_client.py
@@ -43,7 +43,7 @@ class AsyncClient:
         ] = None,
         retry_config: Optional[Mapping[str, Any]] = None,
         web_url: Optional[str] = None,
-        cache: Union[AsyncCache, bool, None] = None,
+        cache: Union[AsyncCache, bool] = False,
     ):
         """Initialize the async client.
 
@@ -56,7 +56,7 @@ class AsyncClient:
             cache: Configuration for caching. Can be:
                 - True: Enable caching with default settings
                 - AsyncCache instance: Use custom cache configuration
-                - None or False: Disable caching (default)
+                - False: Disable caching (default)
         """
         self._retry_config = retry_config or {"max_retries": 3}
         _headers = {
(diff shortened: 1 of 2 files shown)


python/langsmith/client.py
```diff
@@ -704,11 +704,7 @@ def __init__(
         max_batch_size_bytes: Optional[int] = None,
         headers: Optional[dict[str, str]] = None,
         tracing_error_callback: Optional[Callable[[Exception], None]] = None,
-        prompt_cache_enabled: Optional[bool] = None,
-        prompt_cache_max_size: Optional[int] = None,
-        prompt_cache_ttl_seconds: Optional[float] = None,
-        prompt_cache_refresh_interval_seconds: Optional[float] = None,
-        prompt_cache_path: Optional[str] = None,
+        cache: Union[Cache, bool, None] = None,
```
eyurtsev: Could you remove the `None` option as it's identical to False

```python
cache: Union[Cache, bool] = False
```

python/langsmith/async_client.py
```diff
@@ -44,11 +43,7 @@ def __init__(
         ] = None,
         retry_config: Optional[Mapping[str, Any]] = None,
         web_url: Optional[str] = None,
-        prompt_cache_enabled: Optional[bool] = None,
-        prompt_cache_max_size: Optional[int] = None,
-        prompt_cache_ttl_seconds: Optional[float] = None,
-        prompt_cache_refresh_interval_seconds: Optional[float] = None,
-        prompt_cache_path: Optional[str] = None,
+        cache: Union[AsyncCache, bool, None] = None,
```
eyurtsev: remove None as an option and set default to False

langchain-infra: fixed

python/langsmith/client.py
langchain-infra: fixed

Issue #1461, opened 2025-01-24
DOC: `push_prompt()` seems not to be supported tagging


TakumaKurosawa:
### Issue with current documentation:

- https://docs.smith.langchain.com/prompt_engineering/how_to_guides/prompts/manage_prompts_programatically
- https://docs.smith.langchain.com/prompt_engineering/how_to_guides/prompts/prompt_tags

### Idea or request for content:

I've tried to manage the system prompts programatically as below code.

<details><summary>push_system_prompt.py</summary>
<p>

```py
if __name__ == "__main__":
    version = os.getenv("VERSION")
    commit_sha = os.getenv("COMMIT_SHA")
    tags = [version, commit_sha]

    prompts: dict[str, ChatPromptTemplate] = {
        "multi-agent-primary-assistant-prompt": ChatPromptTemplate.from_messages(
            [
                SystemMessagePromptTemplate.from_template(
                    MULTI_AGENT_PRIMARY_ASSISTANT_PROMPT
                ),
                HumanMessagePromptTemplate.from_template("{user_input}"),
                MessagesPlaceholder("history"),
            ]
        ),
        # other prompts...
    }

    def push_prompt(
        name: str, prompt: ChatPromptTemplate, tags: list[str] | None
    ) -> None:
        try:
            langsmith_client.push_prompt(
                name,
                object=prompt,
                tags=tags,
            )
        except LangSmithConflictError:
            print(f"{name} is not changed")

    with futures.ThreadPoolExecutor() as executor:
        tasks = [
            executor.submit(push_prompt, name, prompt, tags)
            for name, prompt in prompts.items()
        ]
        futures.wait(tasks)
```

</p>
</details> 

The problem is `tags` doesn't  seem to set on new commit.

<img width="683" alt="Image" src="https://github.com/user-attachments/assets/9676cf4c-0344-4623-acfa-52569cbc291e" />

I expected to set tags like below image.

<img width="695" alt="Image" src="https://github.com/user-attachments/assets/e5a86130-6dfd-4ae5-82d8-23a2b62c8145" />

Are you sure to support tagging for each commits by `push_system(tags=tags)`?
If so, could you tell me how to do that? 🙏

TakumaKurosawa:
I wonder `Use Cases` shows my tags 🤔 
<img width="1350" alt="Image" src="https://github.com/user-attachments/assets/926b192f-0dea-407a-8853-44b93b057467" />

[...]

PR #2265, opened 2026-01-06
feat(python sdk): Add support for setting commit tags when pushing a prompt

bees:
Adds support for passing in a list of commit tags when pushing a prompt. 

Related issue: https://github.com/langchain-ai/langsmith-sdk/issues/1461

### DIFF
--- a/python/tests/unit_tests/test_client.py
+++ b/python/tests/unit_tests/test_client.py
@@ -4760,14 +4760,16 @@ def test_prompt_commit_tags(mock_session_cls: mock.Mock) -> None:
     ]
     assert len(post_calls) == 3
 
-    for i, tag in enumerate(tags):
-        call_args = post_calls[i]
+    # Verify all tags were created (order not guaranteed due to threading)
+    tag_names_created = {call[1]["json"]["tag_name"] for call in post_calls}
+    assert tag_names_created == set(tags)
+
+    # Verify all calls have correct structure
+    for call_args in post_calls:
         assert call_args[0][0] == "POST"
         assert call_args[0][1].endswith("/repos/test-owner/test-repo/tags")
-        assert call_args[1]["json"] == {
-            "tag_name": tag,
-            "commit_id": commit_id,
-        }
+        assert call_args[1]["json"]["commit_id"] == commit_id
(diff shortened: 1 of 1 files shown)


python/langsmith/async_client.py
```diff
@@ -1355,6 +1355,33 @@ async def _get_latest_commit_hash(
         commits = response.json()["commits"]
         return commits[0]["commit_hash"] if commits else None
 
+    async def _create_commit_tags(
+        self, prompt_owner_and_name: str, commit_id: str, tags: Sequence[str]
```
baskaryan: nit, might want to type as List[str]. any string literal is already a Sequence[str]. also might want to handle someone passing in a string directly instead of a list, otherwise folks might get confused when each character gets turned into its own tag

python/langsmith/async_client.py
```diff
@@ -1615,7 +1645,11 @@ async def create_commit(
             "POST", f"/commits/{prompt_owner_and_name}", json=request_dict
         )
 
-        commit_hash = response.json()["commit"]["commit_hash"]
+        commit_json = response.json()["commit"]
+        commit_hash = commit_json["commit_hash"]
+        commit_id = commit_json["id"]
+        if tags:
+            await self._create_commit_tags(prompt_owner_and_name, commit_id, tags)
```
baskaryan, line 1654: not for this pr but wonder if we shoudl update the create commit endpoint to accept a list of tags, save us # tags extra requests

python/langsmith/client.py
```diff
@@ -7745,6 +7745,29 @@ def _get_latest_commit_hash(
...
+
+        Raises:
+            requests.exceptions.HTTPError: If the request fails.
+        """
+        for tag in tags:
+            payload = {
+                "tag_name": tag,
+                "commit_id": commit_id,
+            }
+            response = self.request_with_retries(
```
baskaryan, line 7770: nit, maybe more worthwhile to just refactor the endpoint, but could parallelize with threads

python/langsmith/async_client.py
bees: great catch, thank you! Still getting used to python's typing module

PR #2270, opened 2026-01-07
feat(js): respect disable run compression

angus-langchain:


js/src/client.ts
```diff
@@ -1201,7 +1235,9 @@ export class Client implements LangSmithTracingClientInterface {
           !this._multipartDisabled &&
           (serverInfo?.batch_ingest_config?.use_multipart_endpoint ?? true);
         if (useMultipart) {
-          const useGzip = serverInfo?.instance_flags?.gzip_body_enabled;
+          const useGzip =
+            getLangSmithEnvironmentVariable("DISABLE_RUN_COMPRESSION") !==
```
hinthornw: Should this be done (env check) once in the global context vs. checking on every batch? Seems superfluous

PR #2271, opened 2026-01-07
feat: add support for tracing Claude Agent SDK JS

dqbd:


js/src/experimental/anthropic/index.ts
```diff
@@ -0,0 +1,1129 @@
...
+    unstable_v2_createSession?: (...args: unknown[]) => unknown;
+    unstable_v2_prompt?: (...args: unknown[]) => unknown;
+    unstable_v2_resumeSession?: (...args: unknown[]) => unknown;
+  };
+
+  const inputSdk = sdk as TypedSdk;
+  const wrappedSdk = { ...sdk } as TypedSdk;
+
+  // Wrap the query method if it exists
+  if ("query" in inputSdk && typeof inputSdk.query === "function") {
```
angus-langchain, line 1185: Each call to sdk.query or sdk.tool re-wraps the function. Should we cache wrapped functions

js/src/experimental/anthropic/index.ts
```diff
@@ -0,0 +1,1129 @@
...
+  }
+  context.activeToolRuns.clear();
+}
+
+/**
+ * Wraps the Claude Agent SDK's query function to add LangSmith tracing.
+ * Traces the entire agent interaction including all streaming messages.
+ * Internal use only - use wrapClaudeAgentSDK instead.
+ */
+function wrapClaudeAgentQuery<
```
angus-langchain, line 391: This wrapped generator doesn't forward the interrupt() [method](https://platform.claude.com/docs/en/agent-sdk/typescript#methods) or custom properties from the original Query object. This breaks the ability to call query.interrupt() to cancel mid-stream.

angus-langchain: The generator is created inside the async generator, will `interrupt()` be available after query() returns?

js/src/experimental/anthropic/index.ts
```diff
@@ -0,0 +1,1129 @@
...
+              pendingMessages.set(messageId, {
+                message,
+                messageHistory: finalResults.slice(0),
+                startTime: existing.startTime,
+              });
+            }
+
+            // We need to push the assistant message as well
+            if (message.message.content) {
+              finalResults.push({
```
angus-langchain, line 564: Should we only push this message has a  stop_reason and is when it is finalized

angus-langchain: Otherwise could we get duplicate entries

js/src/experimental/anthropic/index.ts
```diff
@@ -0,0 +1,1129 @@
...
+      const { outputs, isError } = formatOutputs(toolResponse);
+      await clientRun.end({
+        outputs,
+        error: isError ? outputs.output?.toString() : undefined,
+      });
+      await clientRun.patchRun();
+
+      // If this is NOT a subagent session (it's a tool within a subagent), remove it
+      if (!isSubagentSession) {
+        // This was a tool within a subagent, safe to remove
```
angus-langchain: can we remove?

js/src/experimental/anthropic/index.ts
dqbd: We do need to append intermediate AI messages to chat history as well (say for tool request -> tool response)

aaron-seq:
## CI/CD Failure Analysis

The integration tests are failing because the `ANTHROPIC_API_KEY` environment variable is missing in CI. Tests on lines 22-23 of `claude_agent_sdk.int.test.ts` throw an error when the key is absent.

**Issue:** Integration test requires API key not configured in GitHub Actions secrets.

**Suggestion:** Either skip these tests in CI (already noted in test comments lines 8-13) or add the API key to repository secrets.

## Code Logic Review

@angus-langchain raised a valid concern about the wrapped generator not forwarding the `interrupt()` method. This breaks the ability to cancel mid-stream execution.

**Recommendation:** Ensure proxy returns an object with all original Query methods including `interrupt()` and custom properties to maintain full API compatibility.

angus-langchain (approved):

js/src/experimental/anthropic/index.ts
```diff
@@ -0,0 +1,1230 @@
...
+        if (typeof prompt === "string") {
+          result.push({ content: prompt, role: "user" });
+        } else {
+          for await (const { message } of prompt) {
+            if (!message) continue;
+            result.push({
+              content: flattenContentBlocks(message.content),
+              role: message.role,
+            });
+          }
```
microstudent: It appears that if prompt is an AsyncIterable, the for await loop in the code will consume the stream directly. Since most streams can only be iterated once, this causes the subsequent actual queryFn invocation to be unable to access the prompt content, leading to a runtime error.
Has the case where prompt is an AsyncIterable already been tested?

dqbd: Thank you for catching! Will defer serializing inputs to prevent runtime errors

PR #2275, opened 2026-01-08
chore: set config in traceable

hinthornw:


python/langsmith/run_helpers.py
```diff
@@ -516,6 +520,7 @@ def manual_extra_function(x):
         dangerously_allow_filesystem=kwargs.pop("dangerously_allow_filesystem", False),
     )
     outputs_processor = kwargs.pop("process_outputs", None)
+    enabled = kwargs.pop("enabled", True)
```
angus-langchain: Does setting enabled=false correctly disable tracing if I'm just using langsmith and not langgraph/chain?

hinthornw: lol added testing

PR #2276, opened 2026-01-08
chore: consolidate workflow files

hinthornw:
Will block merge on the success job, which fails if any of the jobs fail


python/tests/unit_tests/test_client.py
```diff
@@ -1073,7 +1073,7 @@ def test_client_gc_after_autoscale() -> None:
         for call in session.request.mock_calls
         if call.args and call.args[0] == "POST"
     ]
-    assert len(request_calls) >= 500 and len(request_calls) <= 550
+    assert len(request_calls) >= 400 and len(request_calls) <= 1000
```
hinthornw, line 1076: flakey test

PR #2293, opened 2026-01-16
chore: Support jwt auth in replicas

hinthornw:
By default, we use API keys since tracing tends to be a backend concern.

There are some scenarios where you'd want to conditionally trace using alternate credentials.

Importantly, if you provide a different cred, we do not pass the env/client-configured API-key in the headers.

Tested locally to ensure both:
a) you can send to multiple targets using api key for one and auth for another
b) we don't also trace to the default destination

### DIFF
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -1830,11 +1830,6 @@ class Client:
             )
             ```
         """
-        # This is an alternative to API key auth. API key auth takes precedence.
-        # service_key: Optional[str] = None,
-        # tenant_id: Optional[str] = None,
-        # authorization: Optional[str] = None,
-        # cookie: Optional[str] = None,
         service_key: str | None = kwargs.pop("service_key", None)
         tenant_id: str | None = kwargs.pop("tenant_id", None)
         authorization: str | None = kwargs.pop("authorization", None)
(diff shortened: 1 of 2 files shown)


python/langsmith/client.py
```diff
@@ -1970,8 +2048,16 @@ def _create_run_non_batch(
         else:
             # Use all configured write API URLs
             for write_api_url, write_api_key in self._write_api_urls.items():
-                headers = {**self._headers, X_API_KEY: write_api_key}
                 try:
+                    headers = _apply_auth_overrides(
+                        self._headers,
+                        api_key=write_api_key,
+                        service_key=None,
+                        tenant_id=tenant_id,
```
angus-langchain: why are we passing tenant_id here?

python/langsmith/run_trees.py
```diff
@@ -1160,3 +1199,21 @@ def _create_current_dotted_order(
     st = start_time or datetime.now(timezone.utc)
     id_ = run_id or uuid7_from_datetime(st)
     return st.strftime("%Y%m%dT%H%M%S%fZ") + str(id_)
+
+
+def _extract_replica_auth(
```
angus-langchain, line 1213: this is a beefy return value, do we want to make this a bit less unweidly with a NamedTuple?

python/langsmith/client.py
```diff
@@ -2776,6 +2965,8 @@ def update_run(
             client.update_run(run["id"], **run)
             ```
         """
+        authorization = kwargs.pop("authorization", authorization)
```
angus-langchain: why is this in update_run but not create_run

angus-langchain (approved):

python/langsmith/run_trees.py
hinthornw: yee

PR #2300, opened 2026-01-20
Handle streaming inputs in claude agent sdk

angus-langchain:


python/langsmith/integrations/claude_agent_sdk/_client.py
```diff
@@ -162,6 +162,36 @@ def _inject_tracing_hooks(options: Any) -> None:
...
+    messages: list[dict[str, Any]],
+) -> list[dict[str, Any]]:
+    """Unwrap streaming input messages for trace display."""
+    if not messages:
+        return []
+
+    formatted = []
+    for msg in messages:
+        if not isinstance(msg, dict):
+            formatted.append(msg)
```
hinthornw, line 175: what else would it be?

angus-langchain: text/tool blocks, normalized objects

PR #2301, opened 2026-01-20
feat: add NonRecordingRunTree for unconditional tracing API usage

angus-langchain:
### Description
Add a NonRecordingRunTree class that implements the RunTree interface but performs no actual tracing. This enables users to call tracing methods unconditionally without checking if tracing is enabled.

### Changes
- Add `NonRecordingRunTree` class with all methods as no-ops
- Add `NON_RECORDING_RUN` singleton instance
- Update `get_current_run_tree()` to return NON_RECORDING_RUN instead of None
- Update `set_run_metadata()` to be a no-op outside traces
- Export `NonRecordingRunTree` and `NON_RECORDING_RUN` publically

### Usage
```python
run = get_current_run_tree()
run.add_metadata({"key": "value"})  # Works whether tracing or not
if run:  # False when not tracing
   print("Inside trace")
```

### DIFF
--- a/python/langsmith/_expect.py
+++ b/python/langsmith/_expect.py
@@ -104,7 +104,7 @@ class _Matcher:
         self.value = value
         self._executor = _executor or ls_utils.ContextThreadPoolExecutor(max_workers=3)
         self._rt = rh.get_current_run_tree()
-        self._run_id = rt.trace_id if rt.is_recording() else run_id
+        self._run_id = self._rt.trace_id if self._rt.is_recording() else run_id
 
     def _submit_feedback(self, score: int, message: Optional[str] = None) -> None:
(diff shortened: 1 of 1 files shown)


python/langsmith/run_helpers.py
```diff
@@ -69,21 +69,26 @@
 _OTEL_AVAILABLE: Optional[bool] = None
 
 
-def get_current_run_tree() -> Optional[run_trees.RunTree]:
-    """Get the current run tree."""
-    return _context._PARENT_RUN_TREE.get()
+def get_current_run_tree() -> Union[run_trees.RunTree, run_trees.NonRecordingRunTree]:
```
jacoblee93: Can you avoid changing the type here? I think this may break typing in LangChain?

PR #2302, opened 2026-01-21
feat(python): add enabled param tracing context

angus-langchain:


python/langsmith/run_helpers.py
```diff
@@ -1084,7 +1101,7 @@ def _setup(self) -> run_trees.RunTree:
                 attachments=self.attachments or {},  # type: ignore
             )
 
-        if enabled is True:
+        if self.enabled is True or context_enabled is True:
```
hinthornw: ooc: why this instead of `enabled is True`?

PR #2303, opened 2026-01-21
feat: add sandbox sdk

DanielKneipp:
## Add Sandbox Module to LangSmith Python SDK

This PR integrates the sandbox client SDK into the LangSmith Python library as `langsmith.sandbox`.

### Features

- **SandboxClient / AsyncSandboxClient** - Manage sandboxes, templates, volumes, and pools
- **Sandbox / AsyncSandbox** - Execute commands, read/write files in the sandboxes 
- **LangSmith integration** - Automatically uses `LANGSMITH_ENDPOINT` and `LANGSMITH_API_KEY` environment variables

### Usage

```py
from langsmith import sandbox

client = sandbox.SandboxClient()

# Create a template first
client.create_template(name="python-sandbox", image="python:3.12-slim")

# Run code in an isolated sandbox
with client.sandbox(template_name="python-sandbox") as sb:
    result = sb.run("python -c 'print(2 + 2)'")
    print(result.stdout)  # "4\n"
```

### DIFF
--- a/python/langsmith/sandbox/_async_client.py
+++ b/python/langsmith/sandbox/_async_client.py
@@ -220,9 +220,7 @@ class AsyncSandboxClient:
                 ) from e
             if e.response.status_code == 409:
                 data = parse_error_response(e)
-                raise ResourceInUseError(
-                    data["message"], resource_type="volume"
-                ) from e
+                raise ResourceInUseError(data["message"], resource_type="volume") from e
             handle_client_http_error(e)
 
     async def update_volume(
@@ -276,9 +274,7 @@ class AsyncSandboxClient:
                 ) from e
             if e.response.status_code == 400:
                 data = parse_error_response(e)
-                raise ValidationError(
-                    data["message"], error_type="VolumeResize"
-                ) from e
+                raise ValidationError(data["message"], error_type="VolumeResize") from e
             if e.response.status_code == 409:
                 data = parse_error_response(e)
                 raise ResourceNameConflictError(
--- a/python/langsmith/sandbox/_client.py
+++ b/python/langsmith/sandbox/_client.py
@@ -227,9 +227,7 @@ class SandboxClient:
                 ) from e
             if e.response.status_code == 409:
                 data = parse_error_response(e)
-                raise ResourceInUseError(
-                    data["message"], resource_type="volume"
-                ) from e
+                raise ResourceInUseError(data["message"], resource_type="volume") from e
             handle_client_http_error(e)
 
     def update_volume(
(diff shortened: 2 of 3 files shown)


python/langsmith/sandbox/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+    "PoolNotFoundError",
+    "PoolAlreadyExistsError",
+    "PoolValidationError",
+    "PoolTimeoutError",
+    # Name conflict errors (409 on update)
+    "ResourceNameConflictError",
+]
+
+# Emit alpha warning on import
+import logging
```
hinthornw: ooc: why not use `warnings`?

python/langsmith/sandbox/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+    async with sandbox.AsyncSandboxClient() as client:
+        async with await client.sandbox(template_name="python-sandbox") as sb:
+            result = await sb.run("python --version")
+            print(result.stdout)
+"""
+
+from langsmith.sandbox._async_client import AsyncSandboxClient
+from langsmith.sandbox._async_sandbox import AsyncSandbox
+from langsmith.sandbox._client import SandboxClient
+from langsmith.sandbox._exceptions import (
```
hinthornw, line 28: Do we need ALL these things exported on the top level? How concerned are we about import times?

python/langsmith/sandbox/_async_client.py
```diff
@@ -0,0 +1,849 @@
...
+            handle_client_http_error(e)
+            raise  # pragma: no cover
+
+    # ========================================================================
+    # Template Operations
+    # ========================================================================
+
+    async def create_template(
+        self,
+        name: str,
```
hinthornw, line 292: Nice if we include `*` so it's keyword-ony

python/langsmith/sandbox/_async_client.py
```diff
@@ -0,0 +1,849 @@
...
+            return SandboxTemplate.from_dict(response.json())
+        except httpx.ConnectError as e:
+            raise SandboxConnectionError(f"Failed to connect to server: {e}") from e
+        except httpx.HTTPStatusError as e:
+            if e.response.status_code == 404:
+                raise TemplateNotFoundError(f"Template '{name}' not found") from e
+            handle_client_http_error(e)
+            raise  # pragma: no cover
+
+    async def list_templates(self) -> list[SandboxTemplate]:
```
hinthornw, line 376: Pagination?

python/langsmith/sandbox/__init__.py
DanielKneipp: Used a generic `FutureWarning`, let me know if I should use `LangSmithBetaWarning` or create a `LangSmithAlphaWarning` 🙏

python/langsmith/sandbox/__init__.py
DanielKneipp: good one, I actually grouped the exceptions to reduce this

python/langsmith/sandbox/_async_client.py
DanielKneipp: Currently it is not supported by the API, but we will add in the near future

PR #2308, opened 2026-01-23
chore: add as_of to create and update examples response

GowriH-1:
Follow up to https://github.com/langchain-ai/langchainplus/pull/16243

adding as_of to sdk create + update examples response

```
  JavaScript/TypeScript ✓                                                                                                                                                                               
                                                                                                                                                                                                        
  - Create: uploadExamplesMultipart() returns UploadExamplesResponse which now has as_of: string                                                                                                        
  - Update: updateExamplesMultipart() returns UpdateExamplesResponse which extends UploadExamplesResponse, so it inherits the as_of field                                                               
                                                                                                                                                                                                        
  Python ✓                                                                                                                                                                                              
                                                                                                                                                                                                        
  - Create: upsert_examples_multipart() returns UpsertExamplesResponse which now has as_of: str                                                                                                         
  - Update: update_examples_multipart() returns UpsertExamplesResponse which now has as_of: str                                                                                                         

```

Testing:
- Unit tests
- Hit local host:
Create: 
```
{
  "example_ids": [
    "b61b543c-b6de-4256-a18a-a6acc95a30e8",
    "59d5b70f-8c6a-415e-998e-fe17c06cbe74"
  ],
  "count": 2,
  "as_of": "2026-01-23T21:17:28.525513Z"
}
```

Update:
```
{
  "example_ids": [
    "b61b543c-b6de-4256-a18a-a6acc95a30e8"
  ],
  "count": 1,
  "as_of": "2026-01-23T21:17:28.651502Z"
}
```

### DIFF
--- a/js/src/tests/client.int.test.ts
+++ b/js/src/tests/client.int.test.ts
@@ -1485,9 +1485,11 @@ test("upload examples multipart", async () => {
   expect(createdExamples.count).toBe(2);
   expect(createdExamples.as_of).toBeDefined();
   expect(typeof createdExamples.as_of).toBe("string");
-  expect(new Date(createdExamples.as_of).getTime()).toBeLessThanOrEqual(
-    Date.now()
-  );
+  if (createdExamples.as_of) {
+    expect(new Date(createdExamples.as_of).getTime()).toBeLessThanOrEqual(
+      Date.now()
+    );
+  }
 
   const createdExample1 = await client.readExample(exampleId);
   expect(createdExample1.inputs["text"]).toBe("hello world");
@@ -1587,9 +1589,11 @@ test("update examples multipart", async () => {
   ]);
   expect(updateResponse.as_of).toBeDefined();
   expect(typeof updateResponse.as_of).toBe("string");
-  expect(new Date(updateResponse.as_of).getTime()).toBeLessThanOrEqual(
-    Date.now()
-  );
+  if (updateResponse.as_of) {
+    expect(new Date(updateResponse.as_of).getTime()).toBeLessThanOrEqual(
+      Date.now()
+    );
+  }
 
   let updatedExample = await client.readExample(exampleId);
   expect(updatedExample.inputs.text).toEqual("hello world2");


python/langsmith/client.py
```diff
@@ -5450,7 +5450,11 @@ def create_examples(
             ]
 
         if not uploads:
-            return ls_schemas.UpsertExamplesResponse(example_ids=[], count=0)
+            from datetime import datetime, timezone
```
angus-langchain: nit: don't add imports in functions

python/langsmith/client.py
```diff
@@ -5482,9 +5487,13 @@ def _upload_examples_batches_parallel(
                 response = future.result()
                 all_examples_ids.extend(response.get("example_ids", []))
                 total_count += response.get("count", 0)
+                # Track the latest as_of timestamp across all batches
+                # Each batch gets its own timestamp when processed
+                if latest_as_of is None or response["as_of"] > latest_as_of:
```
angus-langchain: response.get("as_of"), this will crash for users who haven't updated langsmith in self hosted

js/src/schemas.ts
```diff
@@ -294,6 +294,7 @@ export interface ExampleUpdateWithAttachments extends ExampleUpdate {}
 export interface UploadExamplesResponse {
   count: number;
   example_ids: string[];
+  as_of: string;
```
baskaryan: not super familiar with JS do we need to add a ? here for self hosted folks

python/langsmith/client.py
```diff
@@ -5450,7 +5450,11 @@ def create_examples(
             ]
 
         if not uploads:
-            return ls_schemas.UpsertExamplesResponse(example_ids=[], count=0)
+            return ls_schemas.UpsertExamplesResponse(
+                example_ids=[],
+                count=0,
+                as_of=datetime.datetime.now(datetime.timezone.utc).isoformat(),
```
baskaryan: can we make this None in the case that no examples were uploaded

python/langsmith/client.py
```diff
@@ -5529,6 +5540,7 @@ def _upload_single_batch(self, batch, dataset_id, dangerously_allow_filesystem):
             return {
                 "example_ids": [data["id"] for data in response_data],
                 "count": len(response_data),
+                "as_of": datetime.datetime.now(datetime.timezone.utc).isoformat(),
```
baskaryan: wonder if it would be better to return None here as well

PR #2310, opened 2026-01-26
feat(py): Allow setting background tracing threads as daemons

jacoblee93:


python/langsmith/_internal/_background_thread.py
```diff
@@ -612,6 +612,7 @@ def tracing_control_thread_func(client_ref: weakref.ref[Client]) -> None:
             threading.Thread(
                 target=tracing_control_thread_func_compress_parallel,
                 args=(weakref.ref(client),),
+                daemon=client._use_daemon_threads,
```
angus-langchain, line 615: we should update the compression thread as well

jacoblee93: This does doesn't it?

PR #2315, opened 2026-01-28
feat(js,python): Use xxhash for deterministic UUID7 for replicas

jacoblee93:
Ran into perf/timing issues with web subtle crypto APIs and given project names are user defined and consequences are not dire if a collision is found anyway, I don't think we need strong guarantees here

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -1885,7 +1885,6 @@ export class Client implements LangSmithTracingClientInterface {
               ...this.fetchOptions,
             } as RequestInit
           );
-          console.log(response);
 
           await raiseForStatus(
             response,


js/src/tests/run_trees.uuid.test.ts
```diff
@@ -81,3 +85,64 @@ test("RunTree default/regular behavior uses UUIDv7 and start_time matches id; po
...
+  // UUID4 input: gets fresh timestamp
+  const beforeMs = Date.now();
+  const derivedV4 = nonCryptographicUuid7Deterministic(uuidv4(), "key");
+  const afterMs = Date.now();
+
+  expect(getUuidVersion(derivedV4)).toBe(7);
+  expect(uuidV7Ms(derivedV4)).toBeGreaterThanOrEqual(beforeMs);
+  expect(uuidV7Ms(derivedV4)).toBeLessThanOrEqual(afterMs);
+});
+
```
dqbd, line 124: Worth either (a) mocking the Date to control or (b) extending `nonCryptographicUuid7Deterministic` to accept custom `Date.now()` to avoid potential flakiness?

jacoblee93: I'll just add a wait to the test

angus-langchain (commented):
would check that this matches the langgraph api impl, so that replicas work between distributed systems

angus-langchain (approved):

PR #2316, opened 2026-01-28
export google gen ai wrapper

angus-langchain:


js/src/wrappers/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from "./openai.js";
+export * from "./gemini.js";
```
jacoblee93: We shouldn't do this - just export from the entrypoint

PR #2317, opened 2026-01-28
pass metadata on feedback creation

angus-langchain:
Forward `session_id` and `start_time` to feedback objects to avoid redis and clickhouse lookups for evaluations and feedback creation in the SDK.

### DIFF
--- a/js/src/tests/client.int.test.ts
+++ b/js/src/tests/client.int.test.ts
@@ -24,7 +24,6 @@ import {
   deleteProject,
   toArray,
   waitUntil,
-  waitUntilRunFound,
   skipIfTransientError,
 } from "./utils.js";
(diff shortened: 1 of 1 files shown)


js/src/client.ts
```diff
@@ -4501,6 +4509,9 @@ export class Client implements LangSmithTracingClientInterface {
           sourceRunId: res.sourceRunId,
           feedbackConfig: res.feedbackConfig as FeedbackConfig | undefined,
           feedbackSourceType: "model",
+          sessionId: run?.session_id,
+          startTime:
+            run?.start_time != null ? String(run.start_time) : undefined,
```
jacoblee93: Why not just accept exactly what runs accept?

Also not sure what `String(new Date())` does - I assume we tolerate both numbers and ISO strings

PR #2329, opened 2026-02-02
fix(evaluation): close ThreadPoolExecutor in async evaluation

baskaryan:
## Description

In `Client.aevaluate()`, when using evaluators with the async evaluation pipeline, a `ThreadPoolExecutor` is created at line 794 in `awith_predictions_and_evaluators()`:

```python
if not hasattr(self, "_evaluation_feedback_executor"):
    self._evaluation_feedback_executor = cf.ThreadPoolExecutor(max_workers=4)
```

This executor is stored as an instance attribute but never explicitly closed, causing a resource leak. In long-running processes or when running many evaluations, this can lead to thread exhaustion.

## Fix

Shutdown the executor in `_aend()` which is called after all examples have been processed:

```python
async def _aend(self) -> None:
    if hasattr(self, "_evaluation_feedback_executor"):
        self._evaluation_feedback_executor.shutdown(wait=True)
    # ... rest of method
```

## Note

The sync version (`Client.evaluate`) does not have this issue because it consistently uses context managers (`with ContextThreadPoolExecutor()`), which automatically handle cleanup.

### DIFF
--- a/python/langsmith/evaluation/_arunner.py
+++ b/python/langsmith/evaluation/_arunner.py
@@ -827,8 +827,6 @@ class _AsyncExperimentManager(_ExperimentManagerMixin):
             async for example in await self.aget_examples():
                 yield process_example(example)
 
-            await self._aend()
-
         # Run the per-example tasks with max-concurrency
         # This guarantees that max_concurrency is the upper limit
         # for the number of target/evaluators that can be run in parallel
@@ -844,6 +842,7 @@ class _AsyncExperimentManager(_ExperimentManagerMixin):
             (result["example"] async for result in r1),
             runs=(result["run"] async for result in r2),
             evaluation_results=(result["evaluation_results"] async for result in r3),
+            _evaluation_feedback_executor=self._evaluation_feedback_executor,
         )
 
     async def awith_predictions(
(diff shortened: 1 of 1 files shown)


aaron-seq (commented):
**Disclosure:** I have reviewed this PR with assistance from an AI coding assistant and have thoroughly examined the changes.

**Code Review:**

This fix correctly addresses the resource leak in async evaluation where a ThreadPoolExecutor was created but never explicitly closed.

**Changes Analysis:**

1. **Problem identification**: The `_evaluation_feedback_executor` ThreadPoolExecutor is created at line 794 in `awith_predictions_and_evaluators()` but was never shutdown, leading to thread exhaustion in long-running processes.

2. **Solution implementation**: The fix properly shuts down the executor in `_aend()` method:
   - Uses `hasattr()` check to verify the executor exists before attempting shutdown
   - Calls `shutdown(wait=True)` to ensure all pending tasks complete before cleanup
   - Placed at the beginning of `_aend()` method, which is called after all examples have been processed

3. **Code correctness**: The logic is sound and mirrors proper cleanup patterns. Using `wait=True` ensures graceful shutdown rather than forceful termination.

4. **Consistency note**: As noted in the PR description, the sync version (`Client.evaluate`) doesn't have this issue because it uses context managers which handle cleanup automatically. This fix brings the async version to parity.

**CI/CD Status:**
2 failing checks, 21 successful checks. The failing test (`test_aevaluate_with_attachments - IndexError: list index out of range`) appears to be an existing flaky test unrelated to this resource cleanup change.

**Conclusion:**
The fix is correct and addresses a real resource leak issue. The implementation follows best practices for executor cleanup.

PR #2330, opened 2026-02-02
feat(python): Non-otel google adk wrapper

angus-langchain:


python/langsmith/integrations/google_adk/_config.py
```diff
@@ -0,0 +1,31 @@
...
+
+_tracing_config: dict[str, Any] = {
+    "name": None,
+    "project_name": None,
+    "metadata": None,
+    "tags": None,
+}
+
+
+def set_tracing_config(
```
ericdong-langchain, line 15: this function will always replace the entire tracing config even if only one parameter is provided.

so if someone updates configure_google_adk(project_name="my_project"), then later configure_google_adk(name="other") it will replace prev updated project_name with None.

```
global _tracing_config
    updates = {}
    if name is not None:
        updates["name"] = name
    if project_name is not None:
        updates["project_name"] = project_name
    if metadata is not None:
        updates["metadata"] = metadata
    if tags is not None:
        updates["tags"] = tags
    _tracing_config = {**_tracing_config, **updates}
```

above should allow for single parameter updates.

python/langsmith/integrations/google_adk/_messages.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+def _safe_serialize(obj: Any) -> Any:
+    """Safely serialize an object to JSON-compatible format."""
+    if obj is None or isinstance(obj, (str, int, float, bool)):
+        return obj
+    if isinstance(obj, bytes):
+        return base64.b64encode(obj).decode("utf-8")
+    if isinstance(obj, dict):
+        return {k: _safe_serialize(v) for k, v in obj.items()}
```
ericdong-langchain, line 94: Circular reference here can have unbounded recursion for cyclical graphs.
Not sure if we will ever hit this edge case though.

python/langsmith/integrations/google_adk/_config.py
angus-langchain: The idea is that this is meant to be called once only, other edits can be done via `tracing_context`, or env vars

python/langsmith/integrations/google_adk/_messages.py
angus-langchain: I don't think tools will return circular json

PR #2335, opened 2026-02-03
feat(sandbox): add js lib

DanielKneipp:
## Description
Add js implementation to interact with sandboxes following the reference [python implementation](https://github.com/langchain-ai/langsmith-sdk/tree/main/python/langsmith/sandbox)

## Usage
```typescript
import { SandboxClient } from "langsmith/experimental/sandbox";

// Client uses LANGSMITH_ENDPOINT and LANGSMITH_API_KEY from environment
const client = new SandboxClient();

// First, create a template (defines the container image)
await client.createTemplate("python-sandbox", { image: "python:3.12-slim" });

// Now create a sandbox from the template and run code
const sandbox = await client.createSandbox("python-sandbox");
try {
  const result = await sandbox.run("python -c 'print(2 + 2)'");
  console.log(result.stdout);    // "4\n"
  console.log(result.exit_code); // 0
} finally {
  await sandbox.delete();
}

// Or use an existing sandbox by name
const existingSb = await client.getSandbox("your-sandbox");
const res = await existingSb.run("python -c 'print(2 + 2)'");
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -52,3 +52,11 @@ repos:
         language: system
         files: ^js/src/.*\.(ts|js)$
         pass_filenames: false
+
+      # JS - type check with TypeScript (cannot auto-fix)
+      - id: js-typecheck
+        name: JS TypeScript Type Check
+        entry: bash -c 'cd js && yarn run check:types'
+        language: system
+        files: ^js/src/.*\.(ts|tsx|mts)$
+        pass_filenames: false
--- a/js/package.json
+++ b/js/package.json
@@ -124,6 +124,7 @@
     "lint:fix": "yarn lint --fix",
     "format": "prettier --write 'src/**/*.{ts,tsx,mts}'",
     "format:check": "prettier --check 'src/**/*.{ts,tsx,mts}'",
+    "check:types": "tsc --noEmit",
     "precommit": "lint-staged",
     "prepublish": "yarn run build"
   },
--- a/js/src/tests/sandbox.test.ts
+++ b/js/src/tests/sandbox.test.ts
@@ -3,12 +3,12 @@ import { jest, describe, it, expect } from "@jest/globals";
 import { SandboxClient } from "../experimental/sandbox/client.js";
 import { Sandbox } from "../experimental/sandbox/sandbox.js";
 import {
-  ResourceNotFoundError,
-  DataplaneNotConfiguredError,
-  QuotaExceededError,
-  ValidationError,
-  ResourceTimeoutError,
-  SandboxCreationError,
+  LangSmithResourceNotFoundError,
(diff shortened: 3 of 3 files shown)


js/src/experimental/sandbox/README.md
jacoblee93, line 1: Should be in docs somewhere as well - maybe let's just make this shorter and link to avoid content duplication

js/src/experimental/sandbox/README.md
```diff
@@ -0,0 +1,363 @@
...
+## Templates
+
+Templates define the container image and resources for sandboxes. **You must create a template before you can create sandboxes.**
+
+```typescript
+// Create a template (required before creating sandboxes)
+const template = await client.createTemplate("my-python-env", {
+  image: "python:3.12-slim",  // Any Docker image
+  cpu: "1",        // CPU limit (default: "500m")
+  memory: "1Gi",   // Memory limit (default: "512Mi")
```
jacoblee93, line 114: Why is this not just G

js/src/experimental/sandbox/helpers.ts
```diff
@@ -0,0 +1,434 @@
...
+
+// =============================================================================
+// Network Error Detection
+// =============================================================================
+
+const objectToString = Object.prototype.toString;
+const isError = (value: unknown): value is Error =>
+  objectToString.call(value) === "[object Error]";
+
+const networkErrorMessages = new Set([
```
jacoblee93: Let's maybe use the existing `AsyncCaller` class?

js/src/experimental/sandbox/sandbox.ts
```diff
@@ -0,0 +1,264 @@
...
+    this.created_at = data.created_at;
+    this.updated_at = data.updated_at;
+    this._client = client;
+  }
+
+  /**
+   * Create a Sandbox from API response data.
+   * @internal
+   */
+  static fromData(data: SandboxData, client: SandboxClient): Sandbox {
```
jacoblee93: Why not just use the constructor?

js/src/experimental/sandbox/errors.ts
jacoblee93, line 1: We prefix all errors with `LangSmith` in the main client, can you add a note to do that here when we combine?

js/src/experimental/sandbox/README.md
DanielKneipp: Agreed, the thing is that for now since this feature is super experimental, I'm trying to keep it a bit under the radar for now since breaking changes are expected. Once things are stable, we will move this to the official docs

js/src/experimental/sandbox/README.md
DanielKneipp: could be as well, in the backend we support both

PR #2349, opened 2026-02-05
feat(js): run evaluations as examples complete

bees:
Currently `evaluate()` buffers all example runs before dispatching evaluation. This PR tweaks control flow to use an async mapping operation so evaluations proceed as soon as runs complete.

### DIFF
--- a/js/src/tests/evaluate.int.test.ts
+++ b/js/src/tests/evaluate.int.test.ts
@@ -643,7 +643,7 @@ test("concurrent evaluate restores dataset order before summary", async () => {
     data: orderedExamples,
     evaluators: [customEvaluator],
     summaryEvaluators: [inputOrderSummaryEvaluator],
-    predictionConcurrency: 2,
+    targetConcurrency: 2,
     evaluationConcurrency: 2,
     description:
       "concurrent evaluate restores dataset order before summary integration test",
--- a/js/src/tests/evaluate_runner.test.ts
+++ b/js/src/tests/evaluate_runner.test.ts
@@ -1,8 +1,10 @@
 import {
   _mapWithConcurrency,
   _reorderResultRowsByExampleIndex,
+  evaluate,
 } from "../evaluation/_runner.js";
 import { PQueue } from "../utils/p-queue.js";
+import { Example } from "../schemas.js";
 
 function sleep(ms: number): Promise<void> {
   return new Promise((resolve) => setTimeout(resolve, ms));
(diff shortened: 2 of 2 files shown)


js/src/evaluation/_runner.ts
```diff
@@ -1242,6 +1268,80 @@ async function _resolveExperiment(
   return [undefined, undefined];
 }
 
+/**
+ * Map over an async iterable with bounded concurrency.
+ * Results are yielded as soon as they resolve (input order is not preserved).
+ * When `maxConcurrency <= 0`, mapping runs sequentially.
+ * Errors from the source iterable or mapper propagate to the consumer.
+ */
+async function* _mapWithConcurrency<TInput, TOutput>(
```
asrira428: Not too familiar with the eval pathways but just to confirm, is there some sort of ordering structure that's in place for if the evals are delivered out of order after async processing?

bees: nope good catch, will add a sorting post-pass

js/src/evaluation/_runner.ts
```diff
@@ -1242,6 +1268,80 @@ async function _resolveExperiment(
...
+  }
+
+  type PendingTask = Promise<{ value: TOutput; self: PendingTask }>;
+
+  const iterator = iterable[Symbol.asyncIterator]();
+  let nextInput: Promise<IteratorResult<TInput>> | null = iterator.next();
+  const pending = new Set<PendingTask>();
+
+  const schedule = (input: TInput) => {
+    const task: PendingTask = mapper(input).then((value) => ({
```
jacoblee93: Can tasks fail? Do we need a catch?

jacoblee93 (approved):
Broadly LGTM, would love a unit test or two for timing though

And yes let's add the post-order sort

js/src/evaluation/_runner.ts
```diff
@@ -1242,6 +1353,85 @@ async function _resolveExperiment(
...
+        nextInput = iterator.next();
+      }
+    } else {
+      pending.delete(winner.res.self);
+      yield winner.res.value;
+    }
+  }
+}
+
+export const __private = {
```
jacoblee93: This isn't necessary, we don't expose this file as an entrypoint anyway. You can just import it and use it in tests

js/src/evaluation/_runner.ts
```diff
@@ -737,16 +837,15 @@ export class _ExperimentManager {
         maxConcurrency,
         debug: this.client.debug,
       });
-      const futures: Promise<ExperimentResultRow>[] = [];
-      for await (const currentResults of this.getResults()) {
-        futures.push(
+
+      for await (const result of _mapWithConcurrency(
+        this.getResults(),
+        maxConcurrency,
```
jacoblee93: I suppose technically this should share concurrency with the `_mapWithConcurrency` for the target function right?

jacoblee93: I think I'm fine with this though

jacoblee93: Not fine, updating

jacoblee93 (approved):

PR #2384, opened 2026-02-09
rfc: improved validation for Client.list_feedback

baskaryan:
Add `project_id=` arg support for Client.list_feedback(). also add validation that one of run_ids/project_id/feedback_key is specified, and handle user providing the raw API params run=, session=, key=, source=.

still needs js support

will be a client minor version bump

### DIFF
--- a/python/langsmith/async_client.py
+++ b/python/langsmith/async_client.py
@@ -826,6 +826,7 @@ class AsyncClient:
         self,
         *,
         run_ids: Optional[Sequence[ls_client.ID_TYPE]] = None,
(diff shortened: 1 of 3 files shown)


yumuzi234:
I see, overall I think this is a good to have, it is better to error out in client, but SDK error and server error is not a big different, as long as we do not break things, it should be fine.
Also, if server behavior change, but client still has to check, which requires user to update SDK to use a new function, not necessary.

yumuzi234 (commented):
commented, to me, the change is fine, but not necessary. @baskaryan can decide, maybe @jacoblee93 knows it better as he worked on SDK more

PR #2394, opened 2026-02-10
Google ADK LangSmith tracing enhancements

j-broekhuizen:
**Enable Nesting:**
- Previously: By default, all spans (agent, LLM calls, tool calls) were siblings under the session root because they all used the same parent context.
- Now:
  1. Save the current context
  2. Create a new context with the agent as the parent
  3. Set the new context so subsequent operations (LLM calls, tool calls) use the agent as their parent
  4. Store the old context in the tuple for later restoration

**Add Agent Instructions as System message:**
- Previously: Agent system instruction wasn't included in the trace message list.
- Now: We retrieve instruction from storage and prepend as system message

**Show tool calls/definitions:**
- Previously: No tool definitions in trace view
- Now: Add logic to convert Google tool schemas to OpenAI format and add to invocation_params

Old:
image

New:
image


python/langsmith/integrations/google_adk/_hooks.py
```diff
@@ -2,33 +2,66 @@
...
+async def before_agent_callback(
+    callback_context: Any, *args: Any, **kwargs: Any
+) -> None:
     """Create chain run for agent invocation."""
     invocation_id = getattr(callback_context, "invocation_id", None)
     if not invocation_id:
         return
 
     agent_name = getattr(callback_context, "agent_name", None) or "google_adk_agent"
+
```
angus-langchain, line 34: can we simplify this to the code I shared, we want to trace what is actually sent to the llm and is less hacky than this implementation

python/langsmith/integrations/google_adk/_hooks.py
```diff
@@ -2,33 +2,66 @@
...
 _active_tool_runs: dict[str, tuple[RunTree, float]] = {}
+_agent_instructions: dict[str, str] = {}
 
 
 def _get_parent_run() -> Optional[RunTree]:
     return get_current_run_tree()
 
 
-def before_agent_callback(callback_context: Any, *args: Any, **kwargs: Any) -> None:
+async def before_agent_callback(
```
angus-langchain: why did we change this to async?

python/langsmith/integrations/google_adk/_hooks.py
```diff
@@ -50,7 +83,18 @@ def before_agent_callback(callback_context: Any, *args: Any, **kwargs: Any) -> N
             start_time=datetime.fromtimestamp(start_time, tz=timezone.utc),
         )
         agent_run.post()
-        _active_agent_runs[invocation_id] = (agent_run, start_time)
+
+        # Store instruction for LLM spans to access
+        if instruction:
+            _agent_instructions[str(agent_run.id)] = instruction
+
+        # Set agent span as current parent for nested operations
```
angus-langchain, line 57: can we do this for context management if we want to child these runs
```
 def before_agent_callback(...):
      agent_run = parent.create_child(...)
      agent_run.post()
      token = _context._PARENT_RUN_TREE.set(agent_run)
      _active_agent_runs[invocation_id] = (agent_run, start_time, token)

  def after_agent_callback(...):
      agent_run, _, token = _active_agent_runs.pop(invocation_id)
      _context._PARENT_RUN_TREE.reset(token)
      agent_run.end(...)
      agent_run.patch()
```

python/langsmith/integrations/google_adk/_tools.py
```diff
@@ -0,0 +1,166 @@
+"""Tool extraction and conversion utilities for Google ADK integration."""
```
angus-langchain: it's great we're tracing tools, but this parsing might be unnecessary if FunctionDeclaration is a pydantic model or if it has `to_json_dict()`

angus-langchain: It's pydantic, I think you can just do this
```py
  def extract_tools_from_llm_request(llm_request: Any) -> list[dict[str, Any]]:
      config = getattr(llm_request, "config", None)
      if not config:
          return []
      tools_list = getattr(config, "tools", None)
      if not tools_list:
          return []
      result = []
      for tool in tools_list:
          for func_decl in getattr(tool, "function_declarations", None) or []:
              try:
                  dumped = func_decl.model_dump(exclude_none=True)
                  result.append({
                      "type": "function",
                      "function": dumped,
                  })
              except Exception:
                  pass
      return result
```

python/langsmith/integrations/google_adk/_hooks.py
j-broekhuizen: Was to support old approach - changed it back

python/langsmith/integrations/google_adk/_hooks.py
j-broekhuizen: Done

python/langsmith/integrations/google_adk/_tools.py
j-broekhuizen: Agree

j-broekhuizen: Done

angus-langchain: do we need this file anymore?

python/langsmith/integrations/google_adk/_client.py
```diff
@@ -9,7 +9,7 @@
 from datetime import datetime, timezone
 from typing import Any, Optional
 
-from langsmith.run_helpers import get_current_run_tree, trace
+from langsmith.run_helpers import get_current_run_tree, get_tracing_context, trace
```
angus-langchain: get_tracing_context unused

python/langsmith/integrations/google_adk/_hooks.py
```diff
@@ -152,10 +164,11 @@ def clear_active_runs() -> None:
     """Clear all active runs (call when session ends)."""
     global _active_agent_runs, _active_tool_runs
 
-    for _, (run, _) in _active_agent_runs.items():
+    for _, (run, _, token) in _active_agent_runs.items():
         try:
             run.end(error="Session ended")
             run.patch()
+            _context._PARENT_RUN_TREE.reset(token)
```
angus-langchain: I think we can remove this line in clear_active_runs

PR #2419, opened 2026-02-12
feat(python): add feedback config CRUD and annotation queue rubric items (closes LSPE-67)

bentanny:
## Summary
- Add SDK methods to create, list, update, and delete feedback configurations (the org-wide definitions for annotation feedback tags like thumbs up/down, rating scales, freeform notes)
- Add `rubric_items` parameter to `create_annotation_queue` and `update_annotation_queue` so feedback configs can be programmatically assigned to queues
- Make `name` optional on `update_annotation_queue` to match backend schema (previously required even for non-name updates)
- New `FeedbackConfigSchema` and `AnnotationQueueRubricItem` types in `schemas.py`
- Both sync (`Client`) and async (`AsyncClient`) implementations
- Integration tests covering full CRUD lifecycle

Follows existing SDK conventions: iterator-based list via `_get_paginated_list`, `request_with_retries` for all HTTP calls, `TypedDict` for lightweight config objects, `BaseModel` for response schemas. The `/feedback-configs` backend API already exists — this just exposes it in the SDK.

Rubric items were added to the existing `create/update_annotation_queue` methods rather than creating separate endpoints, since the backend accepts them as part of the queue create/update payload.

## Test Plan
- [x] `test_feedback_config_crud` — sync integration test (create, list, update, upsert, delete, verify)
- [x] `test_feedback_config_crud` — async integration test (same lifecycle)
- [x] `test_annotation_queue_crud` — existing tests still pass with new optional params
- [x] Syntax validation passes (`ast.parse` on all modified files)

### DIFF
--- a/python/langsmith/async_client.py
+++ b/python/langsmith/async_client.py
@@ -944,6 +944,8 @@ class AsyncClient:
         name: str,
         description: Optional[str] = None,
         queue_id: Optional[ID_TYPE] = None,
+        rubric_instructions: Optional[str] = None,
+        rubric_items: Optional[list[ls_schemas.AnnotationQueueRubricItem]] = None,
     ) -> ls_schemas.AnnotationQueue:
         """Create an annotation queue on the LangSmith API.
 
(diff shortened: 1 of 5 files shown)


python/.gitignore
```diff
@@ -1,3 +1,4 @@
 out
 profiles
 notebooks
+.env
```
jacoblee93: Hmm think the main `.gitignore` should cover this

jacoblee93 (approved):

PR #2427, opened 2026-02-13
fix: propagate tracingEnabled to nested traceables

hntrl:
## Summary

Fixes a bug where nested `traceable` calls would still produce traces even when an ancestor `traceable` had `tracingEnabled: false`, because the `ContextPlaceholder` stored in `AsyncLocalStorage` did not carry the disabled tracing signal.

This is the langsmith-sdk side of a two-part fix: https://github.com/langchain-ai/langchainjs/pull/10044

### `js/src/traceable.ts`

1. **`getTracingRunTree`**: When tracing is disabled, the returned `ContextPlaceholder` now carries `{ tracingEnabled: runTree.tracingEnabled }` instead of an empty `{}`. This preserves the disabled signal for downstream consumers.

2. **Child config inheritance**: When a child `traceable` runs inside a parent that stored a `ContextPlaceholder` with `tracingEnabled: false`, and the child didn't explicitly set `tracingEnabled`, the child's config now inherits `tracingEnabled: false` from the parent context.

### `js/src/singletons/types.ts`

- Added optional `tracingEnabled?: boolean` property to `ContextPlaceholder` type so it can carry the tracing signal.

### DIFF
--- a/js/src/traceable.ts
+++ b/js/src/traceable.ts
@@ -350,7 +350,7 @@ const getTracingRunTree = <Args extends unknown[]>(
     | undefined
 ): RunTree | ContextPlaceholder => {
   if (!isTracingEnabled(runTree.tracingEnabled)) {
-    return {};
+    return { tracingEnabled: runTree.tracingEnabled };
   }
 
   const [attached, args] = handleRunAttachments(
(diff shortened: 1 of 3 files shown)


js/src/traceable.ts
```diff
@@ -904,8 +904,21 @@ export function traceable<Func extends (...args: any[]) => any>(
         return [currentRunTree, processedArgs as Inputs];
       }
 
+      // If the parent context explicitly disabled tracing and the child
+      // didn't override it, propagate the tracingEnabled setting so that
+      // child runs inside a traceable({ tracingEnabled: false }) wrapper
+      // also have tracing disabled.
+      const childConfig =
+        ensuredConfig.tracingEnabled === undefined &&
```
jacoblee93, line 912: Just checking - if you have a child run with `tracingEnabled: true` inside a parent run with `tracingEnabled: false` it will still trace the child after this right? Because `ensuredConfig.tracingEnabled` would be `true`

jacoblee93: This does seem a little bit brittle...

PR #2442, opened 2026-02-18
feat(python): add sandbox exception types and client plumbing

DanielKneipp:
## Summary

- Adds `CommandTimeoutError` and `SandboxServerReloadError` exception types for sandbox command execution.
- Stores the API key on `SandboxClient` / `AsyncSandboxClient` for downstream use.
- Adds `__del__` cleanup methods to both clients for best-effort resource cleanup on garbage collection.

Part 1 of sandbox WebSocket execution support (split from #2437).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/python/langsmith/sandbox/_async_client.py
+++ b/python/langsmith/sandbox/_async_client.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 from typing import Any, Optional
 
 import httpx
@@ -91,8 +92,6 @@ class AsyncSandboxClient:
         """
         try:
             if not self._http.is_closed:
-                import asyncio
-
                 try:
                     loop = asyncio.get_running_loop()
                     if not loop.is_closed():


python/langsmith/sandbox/_client.py
```diff
@@ -88,6 +89,14 @@ def close(self) -> None:
         """Close the HTTP client."""
         self._http.close()
 
+    def __del__(self) -> None:
```
DanielKneipp, line 92: This is to prevent `sys:1: ResourceWarning: unclosed <socket.socket...` if the user uses the client without a context and doesn't call `close()`

python/langsmith/sandbox/_client.py
```diff
@@ -79,6 +79,7 @@ def __init__(
         """
         self._base_url = (api_endpoint or _get_default_api_endpoint()).rstrip("/")
         resolved_api_key = api_key or _get_default_api_key()
+        self._api_key = resolved_api_key
```
DanielKneipp, line 82: This will later be used by WS connections

langchain-infra (approved):

python/langsmith/sandbox/_async_client.py
```diff
@@ -81,6 +82,26 @@ async def aclose(self) -> None:
...
+    def __del__(self) -> None:
+        """Best-effort cleanup of the async HTTP client on garbage collection.
+
+        If an event loop is running, schedules ``aclose()`` as a task.
+        Otherwise the underlying sockets will be closed by the GC.
+        For deterministic cleanup, use ``async with`` or ``await aclose()``.
+        """
+        try:
+            if not self._http.is_closed:
+                import asyncio
```
QuentinBrosse: Move to the top?

QuentinBrosse (approved):

PR #2455, opened 2026-02-19
feat: add native thread querying support

ericdong-langchain:
Added native support for querying threads and their runs in the Python and JS SDKs, using the existing /runs/group API and thread-scoped run listing.

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -489,7 +489,7 @@ type Thread = {
   latency_p50: number;
   latency_p99: number;
   feedback_stats: any | null;
-  group_key: string;
+  thread_id: string;
   first_inputs: string;
   last_outputs: string;
   last_error: string | null;
@@ -2558,7 +2558,7 @@ export class Client implements LangSmithTracingClientInterface {
     }
 
     const result: ListThreadsItem[] = [];
-    for (const [groupKey, runs] of threadsMap.entries()) {
+    for (const [threadId, runs] of threadsMap.entries()) {
       runs.sort((a, b) => {
         const aRun = a as unknown as Record<string, unknown>;
         const bRun = b as unknown as Record<string, unknown>;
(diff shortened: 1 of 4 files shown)


js/src/client.ts
```diff
@@ -2423,6 +2443,166 @@ export class Client implements LangSmithTracingClientInterface {
...
+    if (projectId && projectName) {
+      throw new Error("Provide exactly one of projectId or projectName");
+    }
+
+    const sessionId =
+      projectId ?? (await this.readProject({ projectName: projectName! })).id;
+    const startTimeResolved =
+      startTime ?? new Date(Date.now() - 1 * 24 * 60 * 60 * 1000);
+
+    const runSelect = [
```
QuentinBrosse, line 2506: Should we have a const default value that we reuse?

python/langsmith/client.py
```diff
@@ -1607,6 +1607,44 @@ def _get_cursor_paginated_list(
                 break
             params_["cursor"] = cursors["next"]
 
+    def _get_offset_paginated_list_post(
```
QuentinBrosse: dead code?

python/langsmith/client.py
```diff
@@ -3649,6 +3743,134 @@ def list_runs(
...
+    def list_threads(
+        self,
+        *,
+        project_id: Optional[ID_TYPE] = None,
+        project_name: Optional[str] = None,
+        limit: Optional[int] = None,
+        offset: int = 0,
+        filter: Optional[str] = None,
+        start_time: Optional[datetime.datetime] = None,
+    ) -> list[dict]:
```
QuentinBrosse: Should we use a proper type?

js/src/client.ts
ericdong-langchain: run_select is just a list of keys to return no need for values

QuentinBrosse: Yeah my point is that we have this list in multiple place, so maybe we should have a const and reuse it?

ericdong-langchain: I think this is only used once in js and once in py

PR #2476, opened 2026-02-24
feat: persist failed trace payloads to disk for later replay

angus-langchain:
If LangSmith trace uploads fail due to network errors, the trace data is dropped with a warning. We should add an opt-in mechanism to failed payloads to disk so they can be replayed later.

Set LANGSMITH_FAILED_TRACES_DIR=/path/to/dir to enable. On any upload failure the client serializes the multipart payload into a self-contained JSON file:

Files can be replayed later via langsmith traces replay or by replaying the raw HTTP request. traces over a max size are dropped

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -816,7 +816,7 @@ export class Client implements LangSmithTracingClientInterface {
       "FAILED_TRACES_MAX_MB"
     );
     if (failedTracesMb) {
-      const n = Number(failedTracesMb);
+      const n = parseInt(failedTracesMb, 10);
       if (Number.isFinite(n) && n > 0) {
         this.failedTracesMaxBytes = n * 1024 * 1024;
       }
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -1163,7 +1163,7 @@ class Client:
         )
         _max_mb_str = ls_utils.get_env_var("FAILED_TRACES_MAX_MB")
         try:
-            _max_mb = float(_max_mb_str) if _max_mb_str else 0.0
+            _max_mb = int(_max_mb_str) if _max_mb_str else 0
             self._failed_traces_max_bytes: int = (
                 int(_max_mb * 1024 * 1024) if _max_mb > 0 else 100 * 1024 * 1024
             )


python/langsmith/client.py
```diff
@@ -1153,6 +1158,121 @@ def __init__(
...
+            "FAILED_TRACES_DIR"
+        )
+        _max_mb_str = ls_utils.get_env_var("FAILED_TRACES_MAX_MB")
+        try:
+            _max_mb = float(_max_mb_str) if _max_mb_str else 0.0
+            self._failed_traces_max_bytes: int = (
+                int(_max_mb * 1024 * 1024) if _max_mb > 0 else 100 * 1024 * 1024
+            )
+        except (ValueError, OverflowError):
+            self._failed_traces_max_bytes = 100 * 1024 * 1024
```
QuentinBrosse, line 1176: Log something?

python/langsmith/client.py
```diff
@@ -1153,6 +1158,121 @@ def __init__(
...
+        *,
+        endpoint: str,
+        headers: dict,
+        max_bytes: Optional[int] = None,
+    ) -> None:
+        """Persist a failed trace payload to a local fallback directory.
+
+        Saves a self-contained JSON file with the endpoint, the HTTP headers
+        required for replay, and the base64-encoded request body.  Can be
+        replayed later with ``langsmith traces replay`` or a simple POST:
```
QuentinBrosse: `langsmith traces replay` doesn't exists right?

python/langsmith/client.py
```diff
@@ -1153,6 +1158,121 @@ def __init__(
...
+
+            POST /<endpoint>
+            Content-Type: <value from saved headers>
+            [Content-Encoding: <value from saved headers>]
+            <decoded body>
+
+        If *max_bytes* is set, the oldest trace files are deleted (FIFO) after
+        writing so the directory stays within the budget.
```
QuentinBrosse: For simplicity, should we just drop new traces and have a very explicit error log like: `could not write trace to fallback dir %s as it's already over size limit. Increase `FAILED_TRACES_MAX_MB` if possible.

python/langsmith/client.py
```diff
@@ -2907,6 +3039,16 @@ def _send_multipart_req(
                         logger.warning(f"Failed to multipart ingest runs: {exc_desc}")
                     except Exception:
                         logger.warning(f"Failed to multipart ingest runs: {repr(e)}")
+                    self._dump_failed_trace(
+                        lambda: (
```
QuentinBrosse: nit: should we fix code duplication?

js/src/utils/trace_dump/fs.browser.ts
jacoblee93, line 1: We should refactor all `fs` references into a util that presents the fs interface as an abstraction

js/src/utils/trace_dump/fs.browser.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+export const path = {
+  join: (...parts: string[]) => parts.join("/"),
+  dirname: (p: string) => p.split("/").slice(0, -1).join("/"),
+};
+
+export function randomUUID(): string {
+  // Web Crypto API is available in all modern browsers, Node 19+, Bun, Deno
+  return globalThis.crypto.randomUUID();
```
jacoblee93: I think this is a bad idea for perf, should just use the `uuid` package

js/src/client.ts
```diff
@@ -769,6 +770,28 @@ export class Client implements LangSmithTracingClientInterface {
   private _runCompressionDisabled =
     getLangSmithEnvironmentVariable("DISABLE_RUN_COMPRESSION") === "true";
 
+  private failedTracesDir: string | undefined = (() => {
+    const dir = getLangSmithEnvironmentVariable("FAILED_TRACES_DIR");
```
jacoblee93: Just cache outside?

js/src/client.ts
```diff
@@ -769,6 +770,28 @@ export class Client implements LangSmithTracingClientInterface {
...
+        `LANGSMITH_FAILED_TRACES_DIR is set but file system access is not ` +
+          `available in this environment (${env}). Failed trace data will not be written to disk.`
+      );
+      return undefined;
+    }
+    return dir;
+  })();
+
+  private failedTracesMaxBytes: number = (() => {
+    const mb = getLangSmithEnvironmentVariable("FAILED_TRACES_MAX_MB");
```
jacoblee93: See above why the closure?

js/package.json
```diff
@@ -449,6 +449,7 @@
     "rimraf/**/glob": "10.5.0"
   },
   "browser": {
+    "./dist/utils/fs.js": "./dist/utils/fs.browser.js",
     "./dist/utils/prompt_cache/fs.js": "./dist/utils/prompt_cache/fs.browser.js"
```
jacoblee93: I think the hope would be we could just only ever have one browser export

js/src/utils/fs.ts
```diff
@@ -0,0 +1,72 @@
...
+export async function mkdir(dir: string): Promise<void> {
+  await nodeFsPromises.mkdir(dir, { recursive: true });
+}
+
+export async function writeFileAtomic(
+  filePath: string,
+  content: string
+): Promise<void> {
+  const tempPath = `${filePath}.tmp`;
+  await nodeFsPromises.writeFile(tempPath, content, {
```
jacoblee93, line 27: Wait what is this for? Why not just write file?

js/src/client.ts
```diff
@@ -802,6 +809,19 @@ export class Client implements LangSmithTracingClientInterface {
...
     this.fetchImplementation = config.fetchImplementation;
 
+    // Failed trace dump configuration
+    this.failedTracesDir =
+      getLangSmithEnvironmentVariable("FAILED_TRACES_DIR") || undefined;
+    const failedTracesMb = getLangSmithEnvironmentVariable(
+      "FAILED_TRACES_MAX_MB"
+    );
+    if (failedTracesMb) {
+      const n = Number(failedTracesMb);
```
jacoblee93: Should probably be `parseInt`?

PR #2478, opened 2026-02-24
fix(python): generate example_id during test case creation

xuro-langchain:
Generate `example_id` early during test case creation so it's available in tracing context metadata for both sync and async tests.                                                         
                                                                                                   
**Changes:**                                                                                    
- Generate `example_id` using legacy method if not already set during `_create_test_case()`     
- Add `reference_example_id` to sync test tracing context metadata (parity with async)          
                                                                                                   
**Use case:** Users can access the example ID via                                               
`get_tracing_context()["metadata"]["reference_example_id"]` during test execution. This is useful for passing to subprocesses or external services that need to link their traces back to the experiment.                                                                              
                                                                                                   
## Test plan                                                                                    
                                                                                                   
- [x] Verified sync tests have `reference_example_id` in tracing context                        
- [x] Verified async tests have `reference_example_id` in tracing context                       
- [x] Verified each parametrized test case gets a unique example ID                             
- [x] Verified works with pytest-xdist parallel execution

### DIFF
--- a/python/langsmith/testing/_internal.py
+++ b/python/langsmith/testing/_internal.py
@@ -945,6 +945,14 @@ def _create_test_case(
             func, inputs, test_suite.id
         )
         example_id = example_id or legacy_example_id
+    # Ensure example_id is always generated during test case creation.
+    # This allows reference_example_id to be available in tracing context
+    # metadata for use cases like passing to subprocess traces.
+    if not example_id:
+        example_id = _get_example_id(
+            dataset_id=str(test_suite.id),
+            inputs=inputs or {},
+        )
     pytest_plugin = (
         pytest_request.config.pluginmanager.get_plugin("langsmith_output_plugin")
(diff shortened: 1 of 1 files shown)


python/langsmith/testing/_internal.py
```diff
@@ -945,6 +945,14 @@ def _create_test_case(
             func, inputs, test_suite.id
         )
         example_id = example_id or legacy_example_id
+    # Ensure example_id is always generated during test case creation.
+    # This allows reference_example_id to be available in tracing context
+    # metadata for use cases like passing to subprocess traces.
+    if not example_id:
+        example_id = _get_example_id(
```
jacoblee93, line 952: Shouldn't do this early because the generated `example_id` can be dependent on logged inputs or outputs during the test

If you need a static example id just pass it in the decorator

PR #2481, opened 2026-02-25
fix: add integration and version num

samecrowder:
We want to know which tracing integration traces are coming from and which package version as well. This will help us:
1. Do analytics in Hex to figure out typical trace shape based on integration
2. Eventually auto-infer a list of messages or other higher level constructs within the LangSmith UI based on the integration that is being used.

Adds ls_integration and ls_integration_version to metadata for root trace for:
- Python --> claude agent sdk, oai agents sdk, google adk
- JS --> claude agent sdk, vercel ai sdk

I tested these by tracing to LangSmith:
Screenshot 2026-02-24 at 8 25 23 PM
Screenshot 2026-02-24 at 9 03 49 PM
Screenshot 2026-02-24 at 9 04 28 PM
sing my local changes; here are some screenshots that show the metadata it adds:

Close LSDK-93

### DIFF
--- a/js/src/tests/claude_agent_sdk.test.ts
+++ b/js/src/tests/claude_agent_sdk.test.ts
@@ -352,6 +352,7 @@ describe("wrapClaudeAgentSDK", () => {
           extra: {
             metadata: {
               ls_integration: "claude-agent-sdk",
+              ls_integration_version: expect.any(String),
             },
           },
           inputs: {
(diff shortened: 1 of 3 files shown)


python/langsmith/integrations/claude_agent_sdk/_client.py
```diff
@@ -24,6 +24,17 @@
 logger = logging.getLogger(__name__)
 
 TRACE_CHAIN_NAME = "claude.conversation"
+
+
+def _get_package_version(package_name: str) -> str | None:
```
hinthornw, line 31: we should definitely cache this response with an lru_cache

samecrowder: added, if you'd like to confirm if I did it correctly

PR #2488, opened 2026-02-26
fix: sample before transform in ingest

ericdong-langchain:
Widens `_filter_for_sampling` to take in more input types, reordered sampling before transformation in batch and multipart ingest.

### DIFF
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -2510,8 +2510,6 @@ class Client:
         update: Optional[
             Sequence[Union[ls_schemas.Run, ls_schemas.RunLikeDict, dict]]
         ] = None,
-        *,
-        pre_sampled: bool = False,
     ) -> None:
         """Batch ingest/upsert multiple runs in the Langsmith system.
 
@@ -2522,8 +2520,6 @@ class Client:
             update (Optional[Sequence[Union[Run, RunLikeDict]]]):
                 A sequence of `Run` objects or equivalent dictionaries representing
                 runs that have already been created and should be updated / patched.
-            pre_sampled (bool, default=False): Whether the runs have already been subject
-                to sampling, and therefore should not be sampled again.
 
         Raises:
             LangsmithAPIError: If there is an error in the API request.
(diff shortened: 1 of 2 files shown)


python/langsmith/client.py
```diff
@@ -2589,15 +2601,21 @@ def batch_ingest_runs(
             client.batch_ingest_runs(create=runs_to_create, update=runs_to_update)
             ```
         """
+        if not create and not update:
+            return
+        # filter out runs that are not sampled
+        if not pre_sampled:
```
angus-langchain: don't think we need another configuration here, it's expected that we'd filter first then do anonymization.

python/langsmith/client.py
```diff
@@ -2589,15 +2601,21 @@ def batch_ingest_runs(
...
+            create = self._filter_for_sampling(create or EMPTY_SEQ)
+            update = self._filter_for_sampling(update or EMPTY_SEQ, patch=True)
         if not create and not update:
             return
         # transform and convert to dicts
         create_dicts = [
-            self._run_transform(run, copy=False) for run in create or EMPTY_SEQ
```
angus-langchain, line 2596: This safety handled NoneType, if you remove  EMPTY_SEQ it will cause a type error

python/langsmith/client.py
```diff
@@ -2589,15 +2601,21 @@ def batch_ingest_runs(
...
+            self._run_transform(run, copy=False) for run in create
```
angus-langchain: Should we do the run transform first before sampling, would test this heavily

python/langsmith/client.py
angus-langchain, line 1: This PR needs tests

python/langsmith/client.py
```diff
@@ -1966,31 +1967,42 @@ def _should_sample(self) -> bool:
         return random.random() < self.tracing_sample_rate
 
     def _filter_for_sampling(
-        self, runs: Iterable[dict], *, patch: bool = False
-    ) -> list[dict]:
+        self,
+        runs: Iterable[Union[dict, ls_schemas.Run, ls_schemas.RunLikeDict]],
```
angus-langchain, line 1971: why did this type change?

python/langsmith/client.py
```diff
@@ -1966,31 +1967,42 @@ def _should_sample(self) -> bool:
...
-                trace_id = run.get("trace_id") or run["id"]
+                trace_id = _val(run, "trace_id", None) or _val(run, "id")
 
                 # If we've already made a decision about this trace, follow it
                 if trace_id in self._filtered_post_uuids:
                     continue
 
                 # For new traces, apply sampling
-                if run["id"] == trace_id:
+                if _val(run, "id") == trace_id:
```
angus-langchain, line 2005: what are the changes here for?

python/langsmith/client.py
ericdong-langchain: Since we are sampling before transforming now, I just changed this to match the _run_transform function which was what was called before.

```
def _run_transform(
        run: Union[ls_schemas.Run, dict, ls_schemas.RunLikeDict],
        ...
)
```

python/langsmith/client.py
ericdong-langchain: these changes are needed since we now allow non dict types as well

angus-langchain (approved):

PR #2513, opened 2026-03-02
feat(sdk): use GET /datasets/:dataset/examples endpoint in list examp…

baskaryan:
…les methods

Update Python (sync + async) and JS clients to call the new `/datasets/{dataset_id}/examples` path instead of `/examples?dataset=...`.


jacoblee93 (commented):
I think this will break people on old versions of the API

jacoblee93 (approved):

gowri100 (approved):

PR #2517, opened 2026-03-03
add get insights reports with runs

Palashio:
allows you to fetch insights report through sdk, example usage:

`report = client.get_insights_report(id=report_id, project_id=project_id)`
```
    if report.clusters:
        first_cluster_name = report.clusters[0].name
        print(f"Clusters: {[c.name for c in report.clusters]}")
        cluster = report.clusters[first_cluster_name]
```

### DIFF
--- a/python/langsmith/schemas.py
+++ b/python/langsmith/schemas.py
@@ -1422,7 +1422,7 @@ class _ClusterWithTraces:
         return getattr(self._cluster, name)
 
     def __dir__(self) -> list[str]:
-        return list(set(dir(self.__class__) | set(dir(self._cluster))))
+        return list(set(dir(self.__class__)) | set(dir(self._cluster)))
 
     def __repr__(self) -> str:
         return f"_ClusterWithTraces(name={self._cluster.name!r})"


insights_report.json
ericdong-langchain, line 1: Should this be removed

examples/get_insights_report_example.py
```diff
@@ -0,0 +1,43 @@
+import json
+import os
+
+from langsmith import Client
+
+
+def main() -> None:
+    api_url = "https://dev.api.smith.langchain.com"
```
QuentinBrosse: use prod?

examples/get_insights_report_example.py
```diff
@@ -0,0 +1,43 @@
...
+    api_key = os.environ.get("LANGCHAIN_API_KEY") or os.environ.get("LANGSMITH_API_KEY")
+    if not api_key:
+        raise SystemExit("Set LANGCHAIN_API_KEY (or LANGSMITH_API_KEY) in your environment")
+    client = Client(api_url=api_url, api_key=api_key)
+
+    # From URL like: dev.smith.langchain.com/.../projects/p/{project_id}?clusterJobId={report_id}
+    # project_id = session/tracing project (path: .../projects/p/71a13ede-...)
+    # report_id = insights clustering job id (query: clusterJobId=8fbd954b-...), NOT insightsConfigId
```
QuentinBrosse: rm?

python/langsmith/schemas.py
```diff
@@ -1339,6 +1339,190 @@ def _repr_html_(self) -> str:
...
+    def __getattr__(self, name: str) -> Any:
+        return getattr(self._cluster, name)
+
+    def __dir__(self) -> list[str]:
+        return list(set(dir(self.__class__) | set(dir(self._cluster))))
+
+    def __repr__(self) -> str:
+        return f"_ClusterWithTraces(name={self._cluster.name!r})"
+
+    def load_traces(self) -> list[dict[str, Any]]:
```
QuentinBrosse, line 1430: Do we need that? It's duplicated with what we have in the client, right?

pyproject.toml
QuentinBrosse, line 1: Do we need this?

examples/get_insights_report_example.py
Palashio: oops this was wrongly commited along w/ the json

python/langsmith/schemas.py
Palashio: `cluster.load_traces()` is per cluster so it's not duplicating, but the pagination logic can be pulled out

ericdong-langchain (approved):

PR #2524, opened 2026-03-04
feat(py): add experiment-level metadata to pytest integration

baskaryan:
Adds first-class support for attaching custom metadata to the experiment (project) object created by the pytest integration, not just to individual test runs.

**Developer experience:**

Primary — define a session-scoped fixture in conftest.py:

```python
# conftest.py
@pytest.fixture(scope="session")
def langsmith_experiment_metadata():
    return {
        "model": "gpt-4o",
        "prompt_version": "v2.3",
        "environment": os.environ.get("ENV", "local"),
    }
```

The fixture is dynamic (can read env vars, call functions) and follows pytest's existing conftest.py hierarchy so it can be scoped per directory.

Fallback — set the LANGSMITH_EXPERIMENT_METADATA env var (JSON string), useful for CI/CD pipelines that don't want to touch code:

```bash
LANGSMITH_EXPERIMENT_METADATA='{"model":"gpt-4o","env":"staging"}' pytest
```

**Implementation:**
- `pytest_plugin.py`: calls `item.getfixturevalue("langsmith_experiment_metadata")` in `pytest_runtest_call`; silently ignores missing fixture
- `testing/_internal.py`: threads `experiment_metadata` through `_UTExtra`, `_create_test_case` (with env-var fallback), `_LangSmithTestSuite.from_test`, `_start_experiment`, and `_end_tests` so metadata appears on both the initial project creation and the final update_project call
- System-managed keys (`__ls_runner`, `revision_id`, git info) always take precedence over user-supplied keys
- 8 new unit tests covering the full flow, precedence rules, and env-var fallback/invalid-JSON handling

### DIFF
--- a/python/langsmith/testing/_internal.py
+++ b/python/langsmith/testing/_internal.py
@@ -938,10 +938,7 @@ def _create_test_case(
             try:
                 experiment_metadata = _orjson.loads(env_val)
             except Exception as e:
-                msg = (
-                    "LANGSMITH_EXPERIMENT_METADATA env var is not valid JSON: "
-                    f"{e}"
-                )
+                msg = f"LANGSMITH_EXPERIMENT_METADATA env var is not valid JSON: {e}"
                 raise ValueError(msg) from e
     signature = inspect.signature(func)
     inputs = rh._get_inputs_safe(signature, *args, **kwargs) or None
--- a/python/tests/unit_tests/test_experiment_metadata.py
+++ b/python/tests/unit_tests/test_experiment_metadata.py
@@ -284,9 +284,7 @@ def test_env_var_invalid_json_raises(monkeypatch):
 
     with (
         patch("langsmith.testing._internal._get_test_suite", return_value=dataset),
-        patch(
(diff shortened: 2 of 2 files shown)


python/langsmith/testing/_internal.py
```diff
@@ -916,6 +930,17 @@ def _create_test_case(
...
     metadata = langtest_extra["metadata"]
     split = langtest_extra["split"]
+    # Resolve experiment_metadata: explicit kwarg > env var
+    experiment_metadata = langtest_extra.get("experiment_metadata")
+    if experiment_metadata is None:
+        env_val = os.environ.get("LANGSMITH_EXPERIMENT_METADATA")
+        if env_val:
+            try:
+                experiment_metadata = _orjson.loads(env_val)
+            except Exception:
```
eyurtsev: Any reason not to raise loudly? Does this code run **before** we start the testing or after the testing is done? If we can run the validation prior to any runs. i'd prefer to have it fail loudly since I probably want the metadata to work if I'm setting it

PR #2526, opened 2026-03-04
fix: initialize otel exporter before background thread

ericdong-langchain:
Initialize the OTEL exporter (set self.otel_exporter) before starting the background tracing thread, so the thread always sees the correct value and doesn’t incorrectly log the “Run compression is not enabled” warning when OTEL is enabled.


jacoblee93 (approved):
Thanks!

PR #2532, opened 2026-03-06
fix: gemini wrapper inputs

ericdong-langchain:
## Issue
If you call generate_content with the **typed** Gemini API (`types.Content` and `types.Part.from_text()`), the traced run in LangSmith had no messages in the input. Only dict-style contents were turned into the messages format we use for tracing.

`_process_gemini_inputs` was only handling dict-shaped content and parts, so when people used the typed API (`types.Content` and `types.Part.from_text()`), we weren’t turning those into messages and the trace input stayed empty.

## Fix
`_content_to_dict(obj)`
Turns a single content item into a dict: pass-through for dicts, model_dump/to_dict for SDK types, or role/parts via getattr as fallback.

`_part_to_dict(obj)`
Turns a single part into a dict or string: same for dict/str, then model_dump/to_dict, then fallbacks for .text, .inline_data, function_call, function_response.

`_process_gemini_inputs()`
For list-style contents, each item is normalized with `_content_to_dict()` and each part with `_part_to_dict()` before the existing dict-based logic. So both dict-style and object-style inputs produce a proper messages list.

### DIFF
--- a/python/langsmith/wrappers/_gemini.py
+++ b/python/langsmith/wrappers/_gemini.py
@@ -140,9 +140,7 @@ def _process_gemini_inputs(inputs: dict) -> dict:
                         }
                     )
                 elif "function_call" in part or "functionCall" in part:
-                    fc = _to_dict(
-                        part.get("function_call") or part.get("functionCall")
-                    )
+                    fc = _to_dict(part.get("function_call") or part.get("functionCall"))
                     if isinstance(fc, dict):
                         content_parts.append(
                             {
@@ -292,8 +290,7 @@ def _process_generate_content_response(response: Any) -> dict:
                             )
                         elif "function_call" in part or "functionCall" in part:
                             fc = _to_dict(
-                                part.get("function_call")
-                                or part.get("functionCall")
+                                part.get("function_call") or part.get("functionCall")
                             )
                             if isinstance(fc, dict):
                                 content_parts.append(


jacoblee93 (approved):

python/langsmith/wrappers/_gemini.py
```diff
@@ -39,6 +39,76 @@ def _convert_config_for_tracing(kwargs: dict) -> None:
         kwargs["config"] = vars(kwargs["config"])
 
 
+def _content_to_dict(obj: Any) -> dict:
```
angus-langchain: can we rethink the api here? We have `_content_to_dict`, `_to_dict_safe`, and `_part_to_dict`. I don't think we need all three, they all have similar flows isinstance(obj, dict) -> model_dump() -> to_dict()

python/langsmith/wrappers/_gemini.py
```diff
@@ -39,6 +39,76 @@ def _convert_config_for_tracing(kwargs: dict) -> None:
...
+    if isinstance(obj, dict):
+        return obj
+    if hasattr(obj, "model_dump"):
+        return obj.model_dump()
+    if hasattr(obj, "to_dict"):
+        return obj.to_dict()
+    return {}
+
+
+def _part_to_dict(obj: Any) -> dict | str:
```
angus-langchain: can we just have a simple serialization method and then separate logic to handle the dict

angus-langchain: also this functionhamdles function_call normalization but then _process_gemini_inputs normalizes it again

angus-langchain: I think all three helpers can just be 
```
  def _to_dict(obj: Any) -> Any:
      if isinstance(obj, (dict, str)):
          return obj
      if hasattr(obj, "model_dump"):
          return obj.model_dump()
      if hasattr(obj, "to_dict"):
          return obj.to_dict()
      return obj
```

angus-langchain: then in `_process_gemini_inputs` jus do
```
content = _to_dict(content) if not isinstance(content, dict) else content
part = _to_dict(part) if not isinstance(part, (dict, str)) else part
```

angus-langchain: can also use `_GenerateContentParameters_to_mldev` `_GenerateContentParameters_to_vertex` serializers from genai lib

python/langsmith/wrappers/_gemini.py
angus-langchain, line 1: PR needs tests :)

PR #2534, opened 2026-03-06
feat: auto populate trace id child runs

ericdong-langchain:
The SDK now always propagates the parent’s thread_id to every child in RunTree.create_child().

Performance cost is O(1) per child shallow copy of extra/metadata and a single key copy.

### DIFF
--- a/js/src/tests/utils.ts
+++ b/js/src/tests/utils.ts
@@ -168,6 +168,7 @@ export function sanitizePresignedUrls(payload: unknown) {
     if (key === "presigned_url") {
       try {
         const url = new URL(value);
+        url.searchParams.set("jwt", "[JWT]");
         url.searchParams.set("Signature", "[SIGNATURE]");
         url.searchParams.set("Expires", "[EXPIRES]");
         return url.toString();


langchain-infra (approved):

python/langsmith/run_trees.py
```diff
@@ -532,6 +532,16 @@ def create_child(
             start_time = max(start_time, self.start_time)
 
         serialized_ = serialized or {"name": name}
+        if extra:
```
angus-langchain, line 535: other tracing methods populate all parent metadata automatically, we should probably do the same here, no need to single out `thread_id`

angus-langchain: because we also support ` session_id/conversation_id` for some reason 😒

js/src/run_trees.ts
```diff
@@ -423,6 +423,21 @@ export class RunTree implements BaseRun {
       child_execution_order: child_execution_order,
     });
 
+    // Propagate parent's thread_id so LangSmith threads work when set only on root.
+    const parentThreadId = this.extra?.metadata?.thread_id;
```
angus-langchain: propagate metadata here as well?

python/langsmith/run_trees.py
ericdong-langchain: propagating all metadata now

PR #2537, opened 2026-03-06
fix: tool calls fix for anthropic wrapper

ericdong-langchain:
Tool call inputs were being silently dropped during streaming, added handler for input_json_delta streams.

Converts anthropic messages to dict and extracts tool_use from content blocks.

### DIFF
--- a/js/src/tests/utils.ts
+++ b/js/src/tests/utils.ts
@@ -168,6 +168,7 @@ export function sanitizePresignedUrls(payload: unknown) {
     if (key === "presigned_url") {
       try {
         const url = new URL(value);
+        url.searchParams.set("jwt", "[JWT]");
         url.searchParams.set("Signature", "[SIGNATURE]");
         url.searchParams.set("Expires", "[EXPIRES]");
         return url.toString();


python/langsmith/wrappers/_anthropic.py
```diff
@@ -173,24 +182,56 @@ def _create_usage_metadata(anthropic_token_usage: dict) -> UsageMetadata:
...
+    ]
+    if tool_use_blocks:
+        outputs["tool_calls"] = [
+            {
+                "id": block.get("id", f"call_{i}"),
+                "type": "function",
+                "index": i,
+                "function": {
+                    "name": block.get("name", ""),
+                    "arguments": json.dumps(block.get("input", {})),
```
angus-langchain: please use orjson

jacoblee93: Can we make this a lint rule?

ericdong-langchain: hm gemini wrapper also has a json.dumps, should i update?

python/langsmith/wrappers/_anthropic.py
```diff
@@ -96,7 +97,10 @@ def _infer_ls_params(prepopulated_invocation_params: dict, kwargs: dict):
 
 
 def _accumulate_event(
```
angus-langchain: you should use `accumulate_event` from `anthropic.lib.streaming._messages`

python/langsmith/wrappers/_anthropic.py
```diff
@@ -173,24 +182,56 @@ def _create_usage_metadata(anthropic_token_usage: dict) -> UsageMetadata:
...
-    return rdict
+        outputs["usage_metadata"] = _create_usage_metadata(anthropic_token_usage)
+    outputs.pop("type", None)
+
+    content = outputs.get("content") or []
+    tool_use_blocks = [
+        b for b in content if isinstance(b, dict) and b.get("type") == "tool_use"
+    ]
+    if tool_use_blocks:
+        outputs["tool_calls"] = [
```
angus-langchain, line 166: do we need to include in invocation params as well?  and if its it `outputs["tool_calls"]`, we likely don't need it in `outputs["content"]`

python/langsmith/wrappers/_anthropic.py
```diff
@@ -173,24 +182,56 @@ def _create_usage_metadata(anthropic_token_usage: dict) -> UsageMetadata:
...
             return {"output": all_chunks}
     if full_message is None:
         return {"output": all_chunks}
+
+    for idx, parts in raw_tool_inputs.items():
+        raw = "".join(parts)
+        if raw:
+            try:
+                full_message.content[idx].input = json.loads(raw)
+            except json.JSONDecodeError:
```
angus-langchain: can we add a warn log here

python/langsmith/wrappers/_anthropic.py
angus-langchain, line 1: lets add tests for these changes

python/langsmith/wrappers/_anthropic.py
angus-langchain: yes please

python/langsmith/wrappers/_anthropic.py
ericdong-langchain: when tool_calls is present, content is already replaced with just the text parts or none

python/langsmith/wrappers/_anthropic.py
```diff
@@ -95,42 +96,14 @@ def _infer_ls_params(prepopulated_invocation_params: dict, kwargs: dict):
     }
 
 
-def _accumulate_event(
-    *, event: MessageStreamEvent, current_snapshot: Message | None
-) -> Message | None:
+@functools.lru_cache
```
angus-langchain, line 99: why cache this instead of importing at the top of the file?

ericdong-langchain: anthropic is optional dependency so import will break for users that dont have it installed even if they never call any functions related to anthropic

ericdong-langchain: same pattern is used elsewhere for optional deps

PR #2569, opened 2026-03-12
fix: fix js missing tz gs

ericdong-langchain:
Fix tz in JS

### DIFF
--- a/js/src/tests/run_trees.int.test.ts
+++ b/js/src/tests/run_trees.int.test.ts
@@ -254,7 +254,7 @@ test("dotted order matches start_time", async () => {
   const runs = await toArray(langchainClient.listRuns({ projectName }));
   expect(runs.length).toEqual(1);
   expect(runs[0].dotted_order).toEqual(parentRun.dotted_order);
-  expect((runs[0].start_time as string).replace(/[-:.]/g, "")).toEqual(
+  expect((runs[0].start_time as string).replace(/[-:.Z]/g, "")).toEqual(
     parentRun.dotted_order.split(".")[0].split("Z")[0]
   );
 });


js/src/client.ts
```diff
@@ -5236,7 +5267,8 @@ export class Client implements LangSmithTracingClientInterface {
       await raiseForStatus(res, "get run from annotation queue");
       return res;
     });
-    return response.json();
+    const run = await response.json();
+    return _normalizeRunTimestamps(run) as RunWithAnnotationQueueInfo;
```
jacoblee93: If you want to avoid this you could update `_normalizeRunTimestamps` to be something like

```ts
function _normalizeRunTimestamps<T extends Run>(): T {
}
```

js/src/client.ts
```diff
@@ -85,6 +85,34 @@ import {
...
+    !ts.includes("Z") &&
+    !ts.includes("+") &&
+    !ts.includes("-", 10)
+  ) {
+    return ts + "Z";
+  }
+  return ts;
+}
+
+function _normalizeRunTimestamps(run: Run): Run {
```
jacoblee93: nit: I don't love impure functions, but if we're going to do it here for perf(?), I would make this not return anything and just accept an array of runs

If you want it to be pure just do:

```ts
const runCopy = {...run};
runCopy.start_time = ...
```

js/src/client.ts
```diff
@@ -2202,7 +2228,7 @@ export class Client implements LangSmithTracingClientInterface {
     { loadChildRuns }: { loadChildRuns: boolean } = { loadChildRuns: false }
   ): Promise<Run> {
     assertUuid(runId);
-    let run = await this._get<Run>(`/runs/${runId}`);
+    let run = _normalizeRunTimestamps(await this._get<Run>(`/runs/${runId}`));
     if (loadChildRuns) {
       run = await this._loadChildRuns(run);
```
jacoblee93, line 2233: Do we need to normalize these too? I guess not right?

jacoblee93 (approved):

js/src/client.ts
ericdong-langchain: still needed for parent run

PR #2572, opened 2026-03-13
fix(python): propagate exceptions from AsyncExperimentResults._proces…

baskaryan:
…s_data

Previously, if _process_data raised, __anext__ would see task.done() and silently raise StopAsyncIteration, swallowing the real error. Now we check task.exception() and re-raise it.

### DIFF
--- a/python/langsmith/evaluation/_arunner.py
+++ b/python/langsmith/evaluation/_arunner.py
@@ -1174,7 +1174,7 @@ class AsyncExperimentResults:
     ):
         self._manager = experiment_manager
         self._results: list[ExperimentResultRow] = []
-        self._lock = asyncio.Lock()
+        self._condition = asyncio.Condition()
         self._task = asyncio.create_task(self._process_data(self._manager))
         self._processed_count = 0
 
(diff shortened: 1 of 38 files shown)


python/langsmith/evaluation/_arunner.py
```diff
@@ -1197,6 +1197,8 @@ async def _wait_until_index(index: int) -> None:
                     self._processed_count += 1
                     return result
                 elif self._task.done():
+                    if self._task.exception():
+                        raise self._task.exception()
```
gowri100: but the task is done right? even if there's an exception? 

do we catch this / do anything with this upstream?

baskaryan: i think we want to raise this error. this is saying that we failed to evaluate an example. currently you can't distinguish b/n when we failed to evaluate an example and when we've finished evaluating all the examples

gowri100: would raising this error signify that the entire batch failed or just 1 example? 
in my head, as long as the evaluation completes (even with failures) - we should still show that the process is done - done != success

baskaryan: currently raising error (whether the actual error or the stopiterationerror) would stop the process, but thats behavior thats already the case. its not actually clear to me whats the right behavior here, if im evaluating 10k examples and something is wrong with my target function or evaluators i may want to fail fast. potentially something we'd want to give user the ability to configure, can look at it in sep pr

baskaryan: taking a closer look at process_data, looks like we already handle most errors as well

```
_run_single_evaluator already has a try/except
  that catches evaluator errors and converts them to error
  EvaluationResults. And _aforward handles target function
  errors via _error_handling. So in practice, errors from
  individual examples are already caught — the exception
  that would bubble up to _process_data would be something
  more systemic (e.g., network failure, serialization bug,
  etc.).
```

PR #2574, opened 2026-03-13
fix(python): propagate exceptions from ExperimentResults._process_data

baskaryan:
Previously, if _process_data raised in the background thread, __iter__ would spin until timeout and silently stop, and wait() would return without error. Now _process_data stores the exception and always sets _processing_complete via try/finally, and both __iter__ and wait() re-raise it.

Sync analogue of #2572 and #2573

### DIFF
--- a/python/langsmith/evaluation/_runner.py
+++ b/python/langsmith/evaluation/_runner.py
@@ -550,6 +550,7 @@ class ExperimentResults:
         self._results: list[ExperimentResultRow] = []
         self._queue: queue.Queue[ExperimentResultRow] = queue.Queue()
         self._processing_complete = threading.Event()
+        self._processing_error: Optional[BaseException] = None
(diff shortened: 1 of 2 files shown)


python/langsmith/evaluation/_runner.py
```diff
@@ -577,19 +578,26 @@ def __iter__(self) -> Iterator[ExperimentResultRow]:
...
+        try:
+            results = self._manager.get_results()
+            for item in tqdm(results):
+                self._queue.put(item)
+                self._results.append(item)
+
+            summary_scores = self._manager.get_summary_scores()
+            self._summary_results = summary_scores
+        except BaseException as e:
+            self._processing_error = e
```
jacoblee93, line 598: Where do we actually expect this error to be thrown OOC?

`get_results`?

PR #2599, opened 2026-03-19
chore: add experiment id

hinthornw:


python/langsmith/evaluation/_runner.py
```diff
@@ -564,6 +564,33 @@ def __init__(self, experiment_manager: _ExperimentManager, blocking: bool = True
...
+
+    @property
+    def comparison_url(self) -> Optional[str]:
+        """The URL to the comparison view for this experiment."""
+        experiment = self._manager._get_experiment()
+        if experiment.url:
+            project_url = experiment.url.split("?")[0]
+            base_url = project_url.split("/projects/p/")[0]
+            return (
+                f"{base_url}/datasets/{self._manager.dataset_id}/compare?"
```
baskaryan, line 580: this is the main url for the experiment. not sure what `self._manager._get_experiment().url` but think this one is what should be returned by `url` property (and can drop `comparison_url`)

PR #2630, opened 2026-03-30
fix(python): Fix broken ci integration tests

bracesproul:


python/langsmith/client.py
```diff
@@ -1975,6 +1979,48 @@ def _insert_runtime_env(self, runs: Sequence[dict]) -> None:
                 {k: v for k, v in langchain_metadata.items() if k not in metadata}
             )
 
+    def _copy_multipart_run_dict(self, run: dict[str, Any]) -> dict[str, Any]:
+        return cast(dict[str, Any], ls_utils.deepish_copy(run))
+
+    def _merge_multipart_run_dicts(
```
hinthornw: This is expensive

python/langsmith/client.py
```diff
@@ -990,6 +992,8 @@ def __init__(
         self.otel_exporter: Optional[OTELExporter] = None
         self._max_batch_size_bytes = max_batch_size_bytes
         self._multipart_disabled: bool = False
+        self._multipart_ingest_run_cache: dict[uuid.UUID, dict[str, Any]] = {}
+        self._multipart_ingest_pending_updates: dict[uuid.UUID, dict[str, Any]] = {}
```
hinthornw: hmmm this could make it easy to have a memory leak

PR #2639, opened 2026-03-31
feat(py, js): add description to hub prompts sdk

asrira428:
## Summary
- Add optional `description` param to `create_commit` / `push_prompt` (Python sync + async, JS/TS)
- Add `description` to `PromptCommit` and `ListedPromptCommit` response schemas (Python + TS)
- Mirrors backend changes from langchain-ai/langchainplus#21074 which added `description` (VARCHAR 1000, nullable) to the commit CRUD endpoints

## Changed files
- `python/langsmith/client.py` — `create_commit()`, `push_prompt()`
- `python/langsmith/async_client.py` — `create_commit()`, `push_prompt()`
- `python/langsmith/schemas.py` — `PromptCommit`, `ListedPromptCommit`
- `js/src/client.ts` — `createCommit()`, `pushPrompt()`
- `js/src/schemas.ts` — `PromptCommit`

## Test plan
- [x] tests pass


js/src/client.ts
```diff
@@ -6187,6 +6191,7 @@ export class Client implements LangSmithTracingClientInterface {
       description?: string;
       readme?: string;
       tags?: string[];
+      commitDescription?: string;
```
jacoblee93, line 6083: Why not just `description`

python/langsmith/async_client.py
```diff
@@ -2038,6 +2046,7 @@ async def push_prompt(
         readme: Optional[str] = None,
         tags: Optional[Sequence[str]] = None,
         commit_tags: Optional[str | list[str]] = None,
+        commit_description: Optional[str] = None,
```
jacoblee93, line 1889: Same as above just call `description` for consistency

js/src/client.ts
asrira428: Good point — the reason I used `commitDescription` is that `pushPrompt` already has a `description` option that maps to prompt-level metadata (passed to `updatePrompt`/`createPrompt`). Using the same name would collide.

Should we:
1. Drop `commitDescription` from `pushPrompt` entirely and only expose it on `createCommit`?
2. Or rename the existing prompt-level `description` to something else?

Happy to go with whatever you prefer.

python/langsmith/async_client.py
asrira428: Same situation here — `push_prompt` already has a `description` param for prompt metadata. Used `commit_description` to avoid the collision. See reply on the JS comment for options.

PR #2649, opened 2026-04-02
fix(py): avoid async trace context retention

anceschia:
## Summary
- Avoid retaining LangSmith trace context snapshots in completed async `@traceable` calls on Python 3.11+.

## Technical Reason
- `asyncio.create_task(..., context=copy_context())` can keep the copied context alive after task completion.
- That retained `_PARENT_RUN_TREE` and `RunTree` objects across traced async requests, causing unbounded memory growth.
- Reusing the current task for LangSmith context propagation fixes the leak without changing trace nesting semantics.

## Changes Made
- Replace helper `asyncio.create_task(..., context=...)` calls in `python/langsmith/run_helpers.py` with an in-task `_swap_tracing_context()` helper.
- Apply the same context-swapping approach to async wrappers, async generator setup, and traced async stream iteration.
- Add Python 3.11+ regression tests in `python/tests/unit_tests/test_run_helpers.py` to assert traced async paths no longer create tasks with a `context=` snapshot.

### DIFF
--- a/python/langsmith/run_helpers.py
+++ b/python/langsmith/run_helpers.py
@@ -602,7 +602,6 @@ def traceable(
             )
 
             try:
-                accepts_context = aitertools.asyncio_accepts_context()
                 if func_accepts_parent_run:
                     kwargs["run_tree"] = run_container["new_run"]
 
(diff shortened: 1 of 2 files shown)


jacoblee93:
Hey @anceschia! I think I may have fixed this another way here: https://github.com/langchain-ai/langsmith-sdk/pull/2696

Could you try bumping to latest and letting me know?

PR #2688, opened 2026-04-09
fix(evaluation): add 1ms epsilon to example_version to fix stale snapshot after update_example()

### LINKED ISSUE #2430: BUG: Experiment snapshots always show pre-update example outputs after update_example()
**Bug**
When dataset examples are updated via client.update_example(), all subsequent experiments snapshot the pre-update outputs as the reference. Both the UI's "Expected Output" column and evaluators receiving reference_outputs see the old values rather than the updated ones. 

**Reproduction**
Create a dataset with an example that has outputs={"result": "pass"}
Update it via client.update_example(example_id, outputs={"result": "fail"})
Run an experiment against that dataset using evaluate()
In the experiment comparison view, observe that "Expected Output" still shows {"result": "pass"}
**Root Cause**
The SDK captures example_version = example.modified_at.isoformat() in _runner.py. The server's version resolution uses an exclusive upper bound on the timestamp - resolving at timestamp T returns the example state before T, not at T.

Since update_example() sets modified_at = T and the SDK reads example_version = T, the resolved snapshot is always the state before the update.

T-1:  example has outputs {"result": "pass"}
T:    update_example(outputs={"result": "fail"})  →  modified_at = T
       SDK reads example_version = T
       Version resolution at T → returns state at T-1 → {"result": "pass"} (STALE)
This is deterministic, not a race condition. The snapshot will be stale no matter how long you wait between the update and the experiment.

**Expected Behavior**
An experiment run after update_example() should snapshot the updated outputs, not the pre-update state.

**Workaround**
Performing a second no-op update (re-writing the same outputs) bumps modified_at to T+1. The SDK then captures example_version = T+1, which correctly resolves to the state that includes the content change at T.

T:      update_example(outputs={"result": "fail"})  →  modified_at = T
T+1:  update_example(outputs={"result": "fail"})  →  modified_at = T+1  (no-op touch)
         SDK reads example_version = T+1
         Version resolution at T+1 → returns state at T → {"result": "fail"} (CORRECT)
**Suggested Fix**
Either:

SDK-side: Add a small epsilon to modified_at (e.g. + timedelta(milliseconds=1)) when setting example_version
Server-side: Change version resolution to use an inclusive upper bound (<= instead of <)

Aftabbs:
## Summary

Fixes #2430

- `evaluate()` and `aevaluate()` now capture `example_version` as `modified_at + 1ms` instead of `modified_at` exactly
- The LangSmith server resolves example versions using an exclusive upper-bound (`< T`), so using `modified_at = T` returns the state *before* T — the pre-update snapshot
- Adding 1ms ensures the resolved snapshot is the state *at* T — the correct post-update outputs

## Root cause

In `_runner.py` and `_arunner.py`:

```python
# Before (always returns pre-update state)
example_version = (example.modified_at or example.created_at).isoformat()
```

The server uses `modified_at < T` for version resolution. When `update_example()` sets `modified_at = T` and the SDK reads `example_version = T`, the server looks up the state before T — which is the stale pre-update value.

## Fix

```python
# After (returns post-update state)
_example_ts = example.modified_at or example.created_at
example_version = (_example_ts + datetime.timedelta(milliseconds=1)).isoformat()
```

The 1ms shift makes the exclusive bound `< T+1ms` include T, resolving to the state at T.

## Test plan

- [x] The fix is deterministic — no race condition, no flakiness
- [x] Applied in both sync (`_runner.py`) and async (`_arunner.py`) runners
- [ ] Reproduce using the steps in #2430, verify `expected_output` shows updated value after `update_example()`

### DIFF
--- a/python/langsmith/evaluation/_arunner.py
+++ b/python/langsmith/evaluation/_arunner.py
@@ -3,8 +3,8 @@
 from __future__ import annotations
 
 import asyncio
-import datetime
 import concurrent.futures as cf
+import datetime
 import io
 import logging
 import pathlib
--- a/python/langsmith/evaluation/_arunner.py
+++ b/python/langsmith/evaluation/_arunner.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import datetime
 import concurrent.futures as cf
 import io
 import logging
@@ -1285,7 +1286,12 @@ async def _aforward(
         project_name=experiment_name,
         metadata={
             **metadata,
-            "example_version": (example.modified_at or example.created_at).isoformat(),
+            # Add 1ms so the server's exclusive upper-bound version resolution
+            # returns the state AT modified_at rather than the state before it.
+            "example_version": (
+                (example.modified_at or example.created_at)
+                + datetime.timedelta(milliseconds=1)
+            ).isoformat(),
         },
         client=client,
     )
--- a/python/langsmith/evaluation/_runner.py
+++ b/python/langsmith/evaluation/_runner.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import ast
+import datetime
 import collections
 import concurrent.futures as cf
 import functools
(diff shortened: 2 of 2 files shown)


jacoblee93:
Closing since the original issue is not reproducible

PR #2696, opened 2026-04-09
fix(py): Fix run tree memory leak

jacoblee93:
Fixes #2689

### DIFF
--- a/python/langsmith/_internal/_context.py
+++ b/python/langsmith/_internal/_context.py
@@ -1,6 +1,7 @@
 """Shared context (ContextVars and global defaults) that configure tracing."""
 
 import contextvars
+import weakref
 from typing import TYPE_CHECKING, Any, Literal, Optional, Union
 
 if TYPE_CHECKING:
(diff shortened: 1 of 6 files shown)


python/langsmith/run_helpers.py
```diff
@@ -1678,7 +1697,13 @@ def _setup_run(
...
     )
     context.run(_context._PROJECT_NAME.set, response_container["project_name"])
-    context.run(_PARENT_RUN_TREE.set, response_container["new_run"])
+    # Store only the RunTree ID (not the RunTree itself) in the copied context.
+    # This prevents memory leaks when contexts are captured by asyncio operations
+    # (call_later, create_task, etc.) — the captured context holds only a string,
+    # and the RunTree can be GC'd once no strong references remain.
+    if new_run is not None:
+        run_id = _context.register_run_tree(new_run)
+        context.run(_context._PARENT_RUN_TREE_ID.set, run_id)
```
angus-langchain: do we need to set this to None if `new_run is None`?

probably want
```py
else:
      context.run(_context._PARENT_RUN_TREE_ID.set, None)
```

python/langsmith/_internal/_context.py
```diff
@@ -18,9 +19,63 @@
...
 )
 _CLIENT = contextvars.ContextVar[Optional["Client"]]("_CLIENT", default=None)
-_PARENT_RUN_TREE = contextvars.ContextVar[Optional["RunTree"]](
-    "_PARENT_RUN_TREE", default=None
+
+# Store only the RunTree ID (string) in the context.
+# The actual RunTree is stored in a WeakValueDictionary, allowing it to be
+# garbage collected when no strong references exist. This prevents memory leaks
+# when contexts are captured by asyncio operations.
+_PARENT_RUN_TREE_ID = contextvars.ContextVar[Optional[str]](
```
angus-langchain: why not store weakref.ref(run_tree) directly in the ContextVar instead of separate ID + WeakValueDictionary

```py
_PARENT_RUN_TREE_REF = contextvars.ContextVar[Optional[weakref.ref]](
      "_PARENT_RUN_TREE_REF", default=None
  )
```

python/langsmith/run_helpers.py
```diff
@@ -2235,4 +2269,4 @@ def _maybe_create_otel_context(run_tree: Optional[run_trees.RunTree]):
 _METADATA = _context._METADATA
 _TRACING_ENABLED = _context._TRACING_ENABLED
 _CLIENT = _context._CLIENT
-_PARENT_RUN_TREE = _context._PARENT_RUN_TREE
+_PARENT_RUN_TREE_ID = _context._PARENT_RUN_TREE_ID
```
angus-langchain: I thin fine to break private api if not using in LC, but thinking this will break users using `_PARENT_RUN_TREE`

angus-langchain (approved):

python/langsmith/run_helpers.py
jacoblee93: Who is doing that? They should be using `get_current_run_tree()` right?

python/langsmith/_internal/_context.py
jacoblee93: Yeah think that works too!

angus-langchain (approved):

PR #2703, opened 2026-04-10
fix(py): Flush buffered runs on Client.cleanup()

johanity:
cleanup() stops the background tracing thread without draining the queue, so runs queued right before shutdown get dropped. flush() already handles the drain.

### DIFF
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -9179,6 +9179,7 @@ class Client:
 
     def cleanup(self) -> None:
         """Manually trigger cleanup of background threads."""
+        self.flush()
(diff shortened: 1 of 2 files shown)


angus-langchain:
Hi @johanity, this is directionally correct, and thank you for flagging this. You're right that cleanup() was stopping the background threads without draining, so runs queued right before shutdown got dropped.

However; `flush()` is unconditionally blocking, while `cleanup()` historically was fire-and-forget. Adding flush() unconditionally can be problematic if the langsmith server isn't available. I've opened #2799 with the same underlying fix but also adds a timeout.

PR #2726, opened 2026-04-16
python: pass OTEL config directly to exporter instead of setting env vars

jkennedyvz:
## Summary

- `get_otlp_tracer_provider()` was writing the LangSmith API key into `os.environ["OTEL_EXPORTER_OTLP_HEADERS"]` and the endpoint into `os.environ["OTEL_EXPORTER_OTLP_ENDPOINT"]`, causing credential-scope pollution for other OTLP exporters in the same process
- Pass `endpoint` and `headers` directly to the `OTLPSpanExporter` constructor instead of mutating process-wide env vars
- User-set env var overrides are still honored (read, not written)

## Test plan

- [x] 4 new unit tests covering: no env mutation, default config passthrough, env var override honoring, no-project header case
- [x] Existing unit tests pass (`test_otel_exporter.py` — 8/8)
- [x] CI integration tests for OTEL tracing

Thanks @OneThing4101 for bringing this bug to our attention!

### DIFF
--- a/python/langsmith/_internal/otel/_otel_client.py
+++ b/python/langsmith/_internal/otel/_otel_client.py
@@ -53,7 +53,8 @@ def get_otlp_tracer_provider() -> "TracerProvider":
       Langsmith-Project header if project is configured
 
     These defaults can be overridden by setting the environment variables before
-    calling this function.
+    calling this function. Values are passed directly to the exporter constructor
+    rather than written to os.environ.
 
     Returns:
         TracerProvider: The OTLP tracer provider.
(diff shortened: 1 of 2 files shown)


OneThing4101:
nice

python/langsmith/_internal/otel/_otel_client.py
```diff
@@ -53,7 +53,8 @@ def get_otlp_tracer_provider() -> "TracerProvider":
       Langsmith-Project header if project is configured
 
     These defaults can be overridden by setting the environment variables before
-    calling this function.
+    calling this function. Values are passed directly to the exporter constructor
```
angus-langchain, line 56: no need to call out the previous approach

angus-langchain (approved):

PR #2734, opened 2026-04-18
feat(python): add Context class for agent/skill/file hub repos

vishnu-ssuresh:
## Summary

Adds a new `Context` class (and `AsyncContext` async counterpart) for working with non-prompt Hub repos — **agents, skills, and files**. These are the directory-based repo types that pair with the existing prompt hub.

The existing prompt methods on `Client` (`pull_prompt`, `push_prompt`, etc.) stay untouched. Extending them to cover agents/skills would have required breaking return-type changes. Instead, everything non-prompt lives behind a new namespace:

```python
from langsmith import Client
from langsmith.schemas import FileEntry

client = Client()
client.context.push_agent(
    "-/my-agent",
    files={"AGENTS.md": FileEntry(content="# My Agent\n")},
)
agent = client.context.pull_agent("-/my-agent")
client.context.delete_agent("-/my-agent")
```

Async mirror: `await async_client.context.pull_agent(...)`.

## What's included

- **Schemas** (`schemas.py`): `FileEntry`, `AgentEntry`, `SkillEntry`, `Entry` (Pydantic discriminated union on `type`), `AgentContext`, `SkillContext`, `FileContext`, `MAX_CONTEXT_ENTRIES = 500`.
- **`Context`** (`context.py`): 12 methods — `pull_*`, `push_*`, `delete_*`, `list_*` for each of agent/skill/file.
- **`AsyncContext`**: same surface, all methods `async def`.
- **`client.context` / `async_client.context`** properties — lazy-cached via the `__slots__`-compatible pattern that existing cached attributes like `_info` / `_settings` use.
- **Top-level exports**: `from langsmith import Context, AsyncContext`.

## Notes

- **Endpoint split**: pull/push/delete hit `/api/v1/platform/hub/repos/{owner}/{repo}/directories[/commits]`; list + repo metadata hit `/api/v1/hub/repos/...`. The directories endpoints are gated server-side by `agent_builder_hub_endpoints`.
- **Parallel classes, not shared base**: `Context` and `AsyncContext` are independent, matching the `Client` / `Sandbox` pattern in this repo rather than the `PromptCache` shared-base pattern.
- **`parent_commit` passthrough**: when the caller provides one it's sent for optimistic concurrency; when omitted, the server defaults to latest (no extra round trip).
- **`files={path: None}`** deletes a path, matching the backend's directory-commit semantics.

## Test Plan

- [x] 21 unit tests in `python/tests/unit_tests/test_context.py` — all pass (mock the HTTP layer)
- [x] Lint clean (`uv run ruff check`)
- [ ] 4 sync + 4 async integration tests in `python/tests/integration_tests/test_context{,_async}.py` — require a live backend with `agent_builder_hub_endpoints` enabled; run before merge
- [ ] Manual smoke: push + pull + delete an agent via `client.context` against a real tenant

### DIFF
--- a/python/langsmith/__init__.py
+++ b/python/langsmith/__init__.py
@@ -6,6 +6,7 @@ if TYPE_CHECKING:
     from langsmith._expect import expect
     from langsmith.async_client import AsyncClient
     from langsmith.client import Client
+    from langsmith.context import AsyncContext, Context
     from langsmith.evaluation import (
         aevaluate,
         aevaluate_existing,
@@ -145,6 +146,14 @@ def __getattr__(name: str) -> Any:
         from langsmith.prompt_cache import AsyncCache
 
         return AsyncCache
+    elif name == "Context":
+        from langsmith.context import Context
+
+        return Context
+    elif name == "AsyncContext":
+        from langsmith.context import AsyncContext
+
+        return AsyncContext
 
     elif name == "configure_global_prompt_cache":
         from langsmith.prompt_cache import configure_global_prompt_cache
@@ -162,6 +171,8 @@ def __getattr__(name: str) -> Any:
 __all__ = [
     "Client",
     "AsyncClient",
+    "Context",
+    "AsyncContext",
     "PromptCache",
     "AsyncPromptCache",
     "Cache",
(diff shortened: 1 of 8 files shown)


vishnu-ssuresh:
Closing in favor of a stacked split — same Context code, now in four reviewable layers:

- #2745 — schemas (py + js)
- #2746 — Python impl + unit tests (based on #2745)
- #2747 — JS impl + unit tests (based on #2746)
- #2748 — integration tests py + js (based on #2747)

All Context-related files are byte-identical across the old branch and the new stack; only difference is the stack is rebased onto current main. Branch \`vishnu/sdk-context-class\` is kept for history.

PR #2735, opened 2026-04-18
draft

ayoung19:
Adds optional `process_run` to `WriteReplica`: called with the run payload before each replica's `create_run`/`update_run`. Return a modified dict to send; return `None` to skip the replica.

Today, replicas share the parent RunTree's client, so `hide_inputs`/`hide_outputs` apply uniformly. `process_run` lets each replica redact (or tag, or drop) independently without a second `Client` in the queue.

The callback receives a run-like dict shaped per `RunLikeDict` (name, id, inputs, outputs, trace_id, session_name, etc.) — documented in the field comment.

```python
def redact(r): return {**r, "inputs": {}, "outputs": {}}

tracing_context(replicas=[
    {"api_key": primary, "process_run": redact},
    {"api_key": debug},  # pass-through
])
```

Additive field, no behavior change for existing replicas. Exceptions in `process_run` are caught and logged — bad processors skip their own replica without breaking others.

## Test plan

- [x] `process_run` transforms per replica independently
- [x] Returning `None` skips only that replica
- [x] Runs on both `post()` and `patch()`
- [x] Raising processor is isolated
- [x] Processor receives the documented RunLikeDict shape
- [x] `exclude_inputs` still takes precedence over processor output on patch

### DIFF
--- a/python/langsmith/run_trees.py
+++ b/python/langsmith/run_trees.py
@@ -58,6 +58,9 @@ class WriteReplica(TypedDict, total=False):
     auth: AuthHeaders
     project_name: Optional[str]
     updates: Optional[dict]
+    client: Any  # Optional Client override — if set, uses this client for
+    # create_run/update_run instead of the parent RunTree's client.
+    # Useful when replicas need different hide_inputs/hide_outputs settings.
 
 
 _HEADER_SAFE_REPLICA_FIELDS: frozenset[str] = frozenset({"project_name", "updates"})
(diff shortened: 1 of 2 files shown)


python/langsmith/run_trees.py
```diff
@@ -58,6 +58,14 @@ class WriteReplica(TypedDict, total=False):
     auth: AuthHeaders
     project_name: Optional[str]
     updates: Optional[dict]
+    process_run: NotRequired[Callable[[dict], Optional[dict]]]
```
hinthornw: Can we use a typeddict here at least/ I don't know if this is a direct input/outputs or a dict-like run object

hinthornw (approved):

PR #2750, opened 2026-04-20
feat: runtime selection of tracing destination

emil-lc:
## Problem

The LangSmith SDK supports exporting traces to three destinations: **LangSmith** (the default), **OpenTelemetry** (a third-party OTEL endpoint), or **both** (hybrid). Today, this is configured globally via environment variables (`LANGSMITH_OTEL_ENABLED` + `LANGSMITH_OTEL_ONLY`), which means every trace from a given process goes to the same place.

Users sometimes need to route individual traces to different destinations at runtime - for example, sending most traces to LangSmith but selectively exporting specific workflows to an OTEL collector only. The only way to achieve this before our change was monkey-patching internal SDK functions between calls, which is fragile and breaks across SDK versions.

## Solution

We add a single new key to the existing `langsmith_extra` dict: `tracing_destinations`. It accepts `"langsmith"`, `"otel"`, or `"hybrid"`. When set for a `@traceable` call, it overrides the global tracing mode for that run and all its descendants (child runs created via `create_child`). When not set, behavior is identical to before

## Test plan
See comments

### DIFF
--- a/python/langsmith/_internal/_background_thread.py
+++ b/python/langsmith/_internal/_background_thread.py
@@ -87,6 +87,12 @@ class TracingQueueItem:
     authorization: Optional[str]
     cookie: Optional[str]
     otel_context: Optional[Context]
+    destination: Optional[str]
+    """Per-run routing override: one of ``"langsmith"``, ``"otel"``, ``"hybrid"``.
+
+    ``None`` means "use the global tracing mode" (see :func:`get_tracing_mode`),
+    preserving legacy behaviour for every item not carrying a per-run override.
+    """
 
     __slots__ = (
         "priority",
@@ -98,6 +104,7 @@ class TracingQueueItem:
         "authorization",
         "cookie",
         "otel_context",
+        "destination",
     )
 
     def __init__(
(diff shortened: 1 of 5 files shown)


emil-lc:
Sample usage:
```
#!/usr/bin/env python3

import openai
from langsmith.wrappers import wrap_openai

openai_client = wrap_openai(openai.Client())

@traceable(name="trace-with-dynamic-client")
def joke():
    print(f"Joking... {openai_client}")
    response = openai_client.chat.completions.create(
        model="gpt-4o-mini",
        messages=[{"role": "user", "content": "Tell me a short joke."}],
    )
    return response.choices[0].message.content

joke()
joke(langsmith_extra={"tracing_destinations": "otel"})
joke(langsmith_extra={"tracing_destinations": "hybrid"})
```

Execution with `LANGSMITH_OTEL_EXPORTER_OTLP_ENDPOINT=https://nonexistent LANGSMITH_OTEL_ENABLED=True uv run test.py`

emil-lc:
Closing in favor of https://github.com/langchain-ai/langsmith-sdk/pull/2759

PR #2760, opened 2026-04-21
fix(js/sandbox): clone Response before consuming body in error handlers

ramon-langchain:
## Summary

The sandbox error handlers (`handleClientHttpError`, `handleSandboxCreationError`, `handleVolumeCreationError`, `handlePoolError`) called `parseErrorResponse` first — which consumes the response body — and then attempted `response.clone()` for the 422 validation-error path or for fall-through handling. Cloning a `Response` after its body is consumed throws:

```
TypeError: Response.clone: Body has already been consumed.
```

Fix: clone the response up front and use the clone for any subsequent reads.

Hit in practice by `captureSnapshot` against a non-ok server response.

## Test plan

- [x] New `sandbox_error_handling.test.ts` spins up a local HTTP server returning 422 responses with both pydantic-validation-list and structured-detail bodies; verifies `captureSnapshot` and `createSandbox` throw the correct typed errors instead of crashing.
- [x] `pnpm test` passes.

### DIFF
--- a/js/src/experimental/sandbox/helpers.ts
+++ b/js/src/experimental/sandbox/helpers.ts
@@ -185,6 +185,7 @@ export async function handleSandboxCreationError(
   response: Response
 ): Promise<never> {
   const status = response.status;
+  const clonedResponse = response.clone();
   const data = await parseErrorResponse(response);
 
   if (status === 408) {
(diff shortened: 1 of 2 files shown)


langchain-infra (approved):
not an issue in py?

PR #2761, opened 2026-04-21
perf: defer langsmith.client import and add defer_build to Pydantic schemas

sydney-runkle:
## Summary

Two independent import-time improvements:

**1. Defer `langsmith.client` import in `run_trees.py`**
- `run_trees.py` previously imported `langsmith.client` at module load, pulling in the full HTTP stack (requests, urllib3, ~40ms)
- `Client`, `_ensure_uuid`, `_dumps_json` are now imported lazily at their call sites inside `get_cached_client()`, `create_child()`, and `_Baggage.to_header()`
- `ID_TYPE`, `RUN_TYPE_T`, `Client` moved to `TYPE_CHECKING`

**2. Add `defer_build=True` to all Pydantic models in `schemas.py`**
- 41 Pydantic models were compiling their validators at module import time
- Adding `defer_build=True` defers validator compilation to first instantiation
- Reduces `langsmith.schemas` self-time from ~18ms to ~10ms
- Removes now-redundant `Example.model_rebuild()` call

## Impact

Combined savings: ~45ms off the `from langchain_core.language_models import BaseChatModel` import chain (which goes through `langsmith.run_trees` → `langsmith.schemas`).

## Test plan

- [ ] `uv run --group test pytest python/tests/unit_tests/ -x -q`


jacoblee93 (approved):
Good with me if it passes tests!

PR #2768, opened 2026-04-22
fix(run-trees): coerce Pydantic BaseModel to dict in RunTree inputs/outputs

### LINKED ISSUE #2765: Issue: RunTree raises ValidationError when inputs or outputs is a Pydantic BaseModel
**langsmith version:** 0.7.33
**pydantic version:** 2.x

---

## Summary

Since `0.6.2` (`feat(python): Bump pydantic to v2`, commit `71b71893`), passing a Pydantic `BaseModel` instance as `inputs` or `outputs` to `RunTree` is broken. There are three distinct failure modes depending on which surface is used:

| Surface | `inputs=BaseModel` | `outputs=BaseModel` |
|---|---|---|
| `RunTree(inputs=..., outputs=...)` | `ValidationError` | `ValidationError` |
| `run_tree.set(inputs=..., outputs=...)` | Silently shallow — nested `BaseModel` fields not coerced | Silently shallow — nested `BaseModel` fields not coerced |
| `run_tree.end(outputs=...)` | N/A | Silently stores `BaseModel` instance as-is (not a `dict`) |

The `@traceable` decorator path (`run_helpers.py:_container_end`) handles `BaseModel` outputs correctly via `model_dump()`. This regression is specific to the `RunTree` direct API.

---

## Reproduction

```python
from pydantic import BaseModel
from langsmith.run_trees import RunTree

class MyInput(BaseModel):
    query: str

class MyOutput(BaseModel):
    answer: str

# Case 1: ValidationError on constructor — inputs
RunTree(
    name="test",
    run_type="chain",
    inputs=MyInput(query="hello"),  # raises ValidationError
)

# Case 2: ValidationError on constructor — outputs
RunTree(
    name="test",
    run_type="chain",
    inputs={},
    outputs=MyOutput(answer="hello"),  # raises ValidationError
)

# Case 3: end() silently stores BaseModel instead of dict
rt = RunTree(name="test", run_type="chain", inputs={})
rt.end(outputs=MyOutput(answer="hello"))
print(type(rt.outputs))  # <class '__main__.MyOutput'> — should be dict

# Case 4: set() appears to work but silently breaks for nested models
class Inner(BaseModel):
    val: int

class MyNestedInput(BaseModel):
    query: str
    nested: Inner

rt2 = RunTree(name="test", run_type="chain", inputs={})
rt2.set(inputs=MyNestedInput(query="hello", nested=Inner(val=42)))
print(type(rt2.inputs["nested"]))  # <class '__main__.Inner'> — should be dict
```

**Cases 1 & 2 error:**
```
ValidationError: 1 validation error for RunTree
inputs
  Input should be a valid dictionary [type=dict_type, ...]
```

---

## Root cause

`RunBase.inputs` and `RunBase.outputs` are both typed as `dict` / `Optional[dict]`. In Pydantic v1 (used via the `pydantic.v1` shim in pre-`0.6.2` versions), `arbitrary_types_allowed = True` in `RunBase.Config` was permissive enough to accept non-dict values at construction time. In Pydantic v2, `dict`-typed fields are strictly validated regardless of `arbitrary_types_allowed`, so passing a `BaseModel` instance raises `ValidationError` immediately.

`end()` bypasses the constructor by assigning directly to `self.outputs` (which skips field validation since `validate_assignment` is not set on the model), silently leaving a `BaseModel` instance in a field typed as `dict`. This produces a `PydanticSerializationUnexpectedValue` warning during serialization and breaks for nested models.

`set()` calls `dict(outputs)` / `dict(inputs)`, which in Pydantic v2 only does a shallow conversion — nested `BaseModel` fields remain as `BaseModel` instances rather than plain dicts.

The correct fix already exists in `run_helpers.py:_container_end` (lines 1442–1458) and just needs to be applied consistently:

```python
# run_helpers.py — already handles this correctly
if isinstance(outputs, dict):
    dict_outputs = outputs
elif hasattr(outputs, "model_dump") and callable(outputs.model_dump) and not isinstance(outputs, type):
    try:
        dict_outputs = outputs.model_dump(exclude_none=True, mode="json")
    except Exception:
        dict_outputs = {"output": outputs}
else:
    dict_outputs = {"output": outputs}
```

---

## Expected behavior

All of the following should produce a plain `dict` without error:

```python
RunTree(..., inputs=MyInput(query="hello"))        # {"query": "hello"}
RunTree(..., outputs=MyOutput(answer="hello"))      # {"answer": "hello"}
run_tree.end(outputs=MyOutput(answer="hello"))      # {"answer": "hello"}
run_tree.set(inputs=MyInput(query="hello"))         # {"query": "hello"}
```

Nested models should be fully coerced, not just the top level.

Aftabbs:
## Summary

Fixes #2765

`RunTree` raised `ValidationError` when a Pydantic `BaseModel` instance was passed as `inputs` or `outputs`, because those fields are typed as `dict`. The `@traceable` decorator already handled this case via `model_dump()` in `_container_end`, but the `RunTree` direct API did not.

**Before:**
```python
RunTree(name="test", run_type="chain", inputs=MyInput(query="hello"))
# ValidationError: inputs - Input should be a valid dictionary
```

**After:**
```python
rt = RunTree(name="test", run_type="chain", inputs=MyInput(query="hello"))
rt.inputs  # {"query": "hello"}
```

## Changes

- Add a module-level `_coerce_to_dict(value)` helper that calls `.model_dump()` on Pydantic BaseModel instances, mirroring the logic in `_container_end`.
- Apply it in `RunTree.infer_defaults` (mode="before" validator) — covers constructor args
- Apply it in `RunTree.set()` — covers `set(inputs=..., outputs=...)`
- Apply it in `RunTree.end()` — covers `end(outputs=...)`

## Test plan

Added five unit tests in `tests/unit_tests/test_run_trees.py` covering:
- Constructor with `inputs=BaseModel`
- Constructor with `outputs=BaseModel`
- `end(outputs=BaseModel)` — verifies no silent storage of raw BaseModel
- `set(inputs=NestedBaseModel)` — verifies full recursive conversion via `model_dump()`
- `set(outputs=BaseModel)`

### DIFF
--- a/python/langsmith/run_trees.py
+++ b/python/langsmith/run_trees.py
@@ -92,6 +92,18 @@ _DISTRIBUTED_PARENT_ID = contextvars.ContextVar[Optional[str]](
 
 _SENTINEL = cast(None, object())
 
+
+def _coerce_to_dict(value):
+    if isinstance(value, dict):
+        return value
+    if (
+        not isinstance(value, type)
+        and hasattr(value, 'model_dump')
+        and callable(value.model_dump)
+    ):
+        return value.model_dump()
+    return dict(value)
+
 TIMESTAMP_LENGTH = 36
 
 
@@ -304,6 +316,10 @@ class RunTree(ls_schemas.RunBase):
             values["tags"] = []
         if values.get("outputs") is None:
             values["outputs"] = {}
+        for _key in ("inputs", "outputs"):
+            _val = values.get(_key)
+            if _val is not None and not isinstance(_val, dict):
+                values[_key] = _coerce_to_dict(_val)
         if values.get("attachments") is None:
             values["attachments"] = {}
         if values.get("replicas") is None:
@@ -392,13 +408,13 @@ class RunTree(ls_schemas.RunBase):
             if inputs is None:
                 self.inputs = {}
             else:
-                self.inputs = dict(inputs)
+                self.inputs = _coerce_to_dict(inputs)
         if outputs is not NOT_PROVIDED:
             self.extra[OVERRIDE_OUTPUTS] = True
             if outputs is None:
                 self.outputs = {}
             else:
-                self.outputs = dict(outputs)
+                self.outputs = _coerce_to_dict(outputs)
         if usage_metadata is not NOT_PROVIDED:
             self.extra.setdefault("metadata", {})["usage_metadata"] = (
                 validate_extracted_usage_metadata(usage_metadata)
@@ -491,10 +507,11 @@ class RunTree(ls_schemas.RunBase):
         # the ones that are automatically included
         if not self.extra.get(OVERRIDE_OUTPUTS):
             if outputs is not None:
+                dict_outputs = _coerce_to_dict(outputs)
                 if not self.outputs:
-                    self.outputs = outputs
+                    self.outputs = dict_outputs
                 else:
-                    self.outputs.update(outputs)
+                    self.outputs.update(dict_outputs)
         if error is not None:
             self.error = error
         if events is not None:
--- a/python/tests/unit_tests/test_run_trees.py
+++ b/python/tests/unit_tests/test_run_trees.py
@@ -550,3 +550,62 @@ def test_from_headers_filters_replica_credentials():
     assert "api_url" not in replica
     assert replica.get("project_name") == "legit-project"
     assert replica.get("updates") == {"reroot": True}
+
+
+# Tests for regression: RunTree should accept Pydantic BaseModel for inputs/outputs
+# https://github.com/langchain-ai/langsmith-sdk/issues/2765
+def test_runtree_accepts_pydantic_basemodel_as_constructor_inputs():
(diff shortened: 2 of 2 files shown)


python/langsmith/run_trees.py
```diff
@@ -304,6 +316,10 @@ def infer_defaults(cls, values: dict[str, Any]) -> dict[str, Any]:
             values["tags"] = []
         if values.get("outputs") is None:
             values["outputs"] = {}
+        for _key in ("inputs", "outputs"):
+            _val = values.get(_key)
+            if _val is not None and not isinstance(_val, dict):
```
QuentinBrosse, line 321: NIT: we already test if it's a dict in the function

QuentinBrosse (approved):

QuentinBrosse:
Merged in https://github.com/langchain-ai/langsmith-sdk/pull/2820
Thanks @Aftabbs!

PR #2770, opened 2026-04-22
feat(js-sandbox): support snapshotName and list filters

DanielKneipp:
## Summary

Follow-up to #2758 — extends the snapshot-only sandbox surface so callers can boot from a snapshot **name** (not just an ID) and so `listSnapshots` supports server-side filtering and pagination.

Sibling to the Python SDK change: #2769.

> Stacked on #2758. GitHub will auto-retarget the base to `main` once that PR merges.

### `createSandbox` — accept `snapshotName`

`snapshotId` becomes optional (positional) and a new `snapshotName` field is added to `CreateSandboxOptions`. Exactly one must be provided. To avoid confusion, `snapshotId` is intentionally not duplicated on the options object.

```ts
// By ID (unchanged):
const sandbox = await client.createSandbox(snapshot.id);

// By name — server resolves the snapshot against the caller's tenant:
const sandbox = await client.createSandbox(undefined, {
  snapshotName: "my-snap",
});
```

Validation throws a `LangSmithValidationError` if neither (or both) is provided.

### `listSnapshots` — add `ListSnapshotsOptions`

```ts
const all = await client.listSnapshots();
const page = await client.listSnapshots({
  nameContains: "python",
  limit: 10,
  offset: 20,
});
```

All three options are optional and independent. Omitting them keeps today's zero-arg behavior — the request URL has no query string. The new type is re-exported from the sandbox public index.

### Tests (`src/tests/sandbox.test.ts`)

- createSandbox: new \`snapshot_name\` payload test + zero-identifier and both-identifier rejection tests.
- listSnapshots: existing zero-arg test extended to assert an empty query string; new test asserts \`name_contains\`, \`limit\`, \`offset\` appear on the request URL.

All 80 sandbox tests pass locally.

### README

- Added a \`snapshotName\` example in the Snapshots section alongside the existing \`snapshotId\` example.
- Added a \`listSnapshots({ nameContains, limit, offset })\` example.
- Updated the API Reference rows for \`createSandbox\` and \`listSnapshots\`.
- Added a \`snapshotName\` row to the \`CreateSandboxOptions\` table.
- Added a new \`ListSnapshotsOptions\` table.

### Breaking changes

None. \`createSandbox(snapshotId)\` calls keep working unchanged; the positional arg is now optional, but the payload behavior is identical when it is provided.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/js/src/experimental/sandbox/client.ts
+++ b/js/src/experimental/sandbox/client.ts
@@ -697,17 +697,28 @@ export class SandboxClient {
   /**
    * List snapshots, optionally filtered and paginated server-side.
    *
-   * @param options - Optional filter/pagination options. When omitted,
-   *   returns all snapshots visible to the caller's tenant.
-   * @returns List of Snapshots.
+   * The backend always paginates this endpoint. When `limit` is omitted the
+   * server applies a default page size (currently 50), so a single call is
+   * not guaranteed to return every snapshot visible to the tenant. To iterate
+   * through all results, repeat the call with increasing `offset` values (or
+   * an explicit `limit`) until fewer than `limit` snapshots come back.
+   *
+   * @param options - Optional filter/pagination options.
+   *   - `nameContains`: case-insensitive substring match on snapshot name.
+   *   - `limit`: page size; must be between 1 and 500 (inclusive). Defaults
+   *     to 50 server-side when omitted.
+   *   - `offset`: number of snapshots to skip; must be `>= 0`.
+   *
+   *   Values outside those ranges are rejected by the server.
+   * @returns A single page of Snapshots matching the provided filters.
    *
    * @example
    * ```typescript
-   * const all = await client.listSnapshots();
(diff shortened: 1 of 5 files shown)


ramon-langchain (approved):

js/src/experimental/sandbox/client.ts
```diff
@@ -214,10 +227,11 @@ export class SandboxClient {
    * ```
    */
   async createSandbox(
-    snapshotId: string,
+    snapshotId?: string,
```
ramon-langchain, line 230: optional nit: why not just allow this to be the name too like `snapshotIdOrName: string` ?

ramon-langchain (approved):

js/src/experimental/sandbox/client.ts
DanielKneipp: just avoiding the edge case of the name having the format of an uuid

PR #2776, opened 2026-04-23
feat(js): add tracing of opencode

dqbd:


js/src/experimental/opencode/tracer.ts
```diff
@@ -0,0 +1,488 @@
...
+    const parentConfig: RunTreeConfig = {
+      name: "opencode.session",
+      run_type: "chain",
+      start_time: parentStartTime,
+      end_time: parentEndTime,
+      extra: {
+        metadata: {
+          ls_integration: "opencode-js",
+          ls_agent_type: "root",
+          thread_id: sessionID,
```
jacoblee93, line 281: We prefer uuid7-like things here now for thread id

Not required, and may be better UX to just use OpenCode sessions for consistency anyway

js/src/experimental/opencode/tracer.ts
```diff
@@ -0,0 +1,488 @@
...
+
+          // TODO: opencode is not anthropic but LangSmith does support rendering Standard Messages
+          // This forces FE to enable messages view
+          ls_message_format: "anthropic",
```
jacoblee93: Hmm doesn't `ls_agent_type` do this already?

jacoblee93 (approved):

js/src/experimental/opencode/tracer.ts
dqbd: Currently FE gates this (and only permits LangChain objects or this, will need to update FE)

js/src/experimental/opencode/tracer.ts
dqbd: I think better to keep for consistency

PR #2781, opened 2026-04-23
perf(js): Offload serialize to worker thread at flush time

jacoblee93:
When LANGSMITH_PERF_OPTIMIZATION=true, the flush-time serializePayloadForTracing() call moves to a Node worker_threads worker, removing main-thread event-loop blocking for large (base64-image-heavy) payloads. postMessage transfers the payload via V8's structured clone, which refcounts large strings across isolates -- about 3x cheaper than stringifying on the main thread.

Falls back to synchronous serialize when: the flag is off, in manualFlushMode, worker_threads is unavailable (browsers / edge / Deno / etc.), or the payload contains non-cloneable values (DataCloneError).

Tracks in-flight drain promises so awaitPendingTraceBatches waits for the async serialize chain to register with batchIngestCaller.queue before checking idleness -- otherwise tests using the callback-based flush timing can race and see 0 fetches.

```
    === Event loop benchmark ===
    Mode:                        WORKER OFF (main-thread serialize at flush)
    Runs traced:                 100
    Per-run create payload:      2511.2 KB
    Per-run update payload:      5.2 KB
    Wall time (incl. drain):     1140.5 ms
    
                     total       max       p50       p95       p99
    createRun      381.58      5.99      3.67      4.97      5.99
    updateRun        8.67      0.29      0.08      0.16      0.29
    loop lag       792.38     41.18      0.16      5.33     39.31
    (loop lag monitor: 1ms target, 324 samples > 0)
```

```
    === Event loop benchmark ===
    Mode:                        WORKER ON (worker-thread serialize at flush)
    Runs traced:                 100
    Per-run create payload:      2511.2 KB
    Per-run update payload:      5.2 KB
    Wall time (incl. drain):     769.2 ms
    
                     total       max       p50       p95       p99
    createRun      374.57      7.03      3.68      4.78      7.03
    updateRun        8.01      0.30      0.06      0.20      0.30
    loop lag       400.87     17.04      0.16      3.76      5.29
    (loop lag monitor: 1ms target, 333 samples > 0)


```

### DIFF
--- a/js/src/utils/worker_threads.ts
+++ b/js/src/utils/worker_threads.ts
@@ -1,0 +1,18 @@
+/**
+ * worker_threads abstraction (Node.js version).
+ *
+ * This file is swapped with worker_threads.browser.ts for browser / edge
+ * builds via the package.json `browser` field. Node gets the real module;
+ * browsers get a stub that signals unavailability.
+ *
+ * Only the surface actually used by SerializeWorker is re-exported.
+ */
+
+// eslint-disable-next-line import/no-unresolved
+import { Worker as NodeWorker } from "node:worker_threads";
+
+export type WorkerLike = InstanceType<typeof NodeWorker>;
+
+export const Worker: typeof NodeWorker | null = NodeWorker;
+
+export const WORKER_THREADS_AVAILABLE = true;
--- a/.github/workflows/js-perf.yml
+++ b/.github/workflows/js-perf.yml
@@ -1,0 +1,207 @@
+name: js-perf
+
+on:
+  pull_request:
+    paths:
+      - "js/src/**"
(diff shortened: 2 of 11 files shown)


js/src/utils/worker_threads.ts
```diff
@@ -0,0 +1,18 @@
+/**
```
dqbd, line 1: We're breaking compat for certain bundlers here:
https://github.com/langchain-ai/langsmith-sdk/actions/runs/24859038406/job/72779614157?pr=2781

jacoblee93: Bad merge I think

dqbd (approved):

PR #2797, opened 2026-04-24
fix(evaluation): preserve plain dict passed to feedback_config

Jeet-51:
Fixes #31802

**Problem**
When a plain dict with keys outside FeedbackConfig was passed to
EvaluationResult.feedback_config, Pydantic v2 Union validation silently
dropped all the data, returning {} instead of the original dict — with
no error or warning.

**Root Cause**
Pydantic v2 Union validation tries to coerce a dict into FeedbackConfig
(a TypedDict). Unknown keys are stripped silently instead of falling back
to the plain dict branch of Union[FeedbackConfig, dict].

**Fix**
Added a field_validator on feedback_config (mode="before") that detects
dicts containing keys outside FeedbackConfig's known fields (type, min,
max, categories) and returns them untouched.

**Tests Added**
- Dict with unknown keys → preserved as-is
- Dict with multiple unknown keys → preserved as-is
- Valid FeedbackConfig keys → still work normally
- None → still works

### DIFF
--- a/python/langsmith/evaluation/evaluator.py
+++ b/python/langsmith/evaluation/evaluator.py
@@ -18,7 +18,7 @@ from typing import (
     cast,
 )
 
-from pydantic import BaseModel, ConfigDict, Field, ValidationError, model_validator
+from pydantic import BaseModel, ConfigDict, Field, ValidationError, field_validator, model_validator
 from typing_extensions import TypedDict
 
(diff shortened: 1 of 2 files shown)


Jeet-51:
I noticed PR #2718 addresses the same issue. Happy to close this one 
if that fix is being merged, or I can build on top of it if needed.

PR #2834, opened 2026-05-01
Add JS profile loading

langchain-infra:
## Summary

Adds LangSmith profile loading to the TypeScript client for Node.js-like runtimes.

- Loads the active profile from `LANGSMITH_CONFIG_FILE` or `~/.langsmith/config.json`
- Selects profiles with `LANGSMITH_PROFILE`, then `current_profile`, then `default`
- Applies profile `api_url`, `workspace_id`, OAuth access tokens, and API keys when explicit/env auth is not set
- Refreshes expired OAuth profile tokens before requests and writes refreshed token fields back to the profile file
- Uses the selected profile's `api_url` for OAuth refresh, even when the client endpoint is overridden
- Skips profile loading in browser-like runtimes where filesystem APIs are unavailable

## Auth precedence

Explicit constructor auth and environment auth take precedence over profile auth. Profile OAuth is preferred over profile `api_key` when profile auth is used. For new profile-specific configuration, this PR only adds `LANGSMITH_*` variables (`LANGSMITH_CONFIG_FILE` and `LANGSMITH_PROFILE`).

## Verification

- `pnpm run format:check`
- `pnpm run lint` (passes with existing warnings)
- `pnpm run check:types`
- `env -u LANGSMITH_API_KEY -u LANGCHAIN_API_KEY -u LANGSMITH_ENDPOINT -u LANGCHAIN_ENDPOINT -u LANGSMITH_WORKSPACE_ID -u LANGCHAIN_WORKSPACE_ID LANGSMITH_CONFIG_FILE=/nonexistent/path/config.json pnpm test:single -- src/tests/client.test.ts --runInBand`


js/src/client.ts
```diff
@@ -686,6 +711,114 @@ const SERVER_INFO_REQUEST_TIMEOUT_MS = 10000;
...
+  }
+  return undefined;
+}
+
+function loadProfileState(): LangSmithProfileState | undefined {
+  if (isBrowserLikeRuntime()) {
+    return undefined;
+  }
+  const configPath = getProfileConfigPath();
+  if (!configPath || !fsUtils.existsSync(configPath)) {
```
dqbd, line 757: _q_: do we need to add a browser specific entrypoint to preserve support for browsers?

langchain-infra: so last time i ran into this we setup browser compilation in ci. basically this only gets conditionally imported if not compiled for the browser

ramon-langchain (approved):

js/src/utils/env.ts
```diff
@@ -98,11 +98,13 @@ export function getLangSmithEnvVarsMetadata(): Record<string, string> {
     "LANGCHAIN_API_KEY",
     "LANGCHAIN_ENDPOINT",
     "LANGCHAIN_TRACING_V2",
+    "LANGCHAIN_CONFIG_FILE",
```
dqbd: _nit_: would either skip `LANGCHAIN_CONFIG_FILE`, since Python does not support afaict or add in Python SDK https://github.com/langchain-ai/langsmith-sdk/pull/2835

js/src/client.ts
```diff
@@ -880,8 +1027,142 @@ export class Client implements LangSmithTracingClientInterface {
...
+      });
+      const response = await fetchImplementation(
+        `${normalizeConfigUrl(this.apiUrl)}/oauth/token`,
+        {
+          method: "POST",
+          headers: {
+            "Content-Type": "application/x-www-form-urlencoded",
+          },
+          body: body.toString(),
+          signal: AbortSignal.timeout(TOKEN_REFRESH_TIMEOUT_MS),
```
dqbd, line 1143: _q_: If fetch's `init?: RequestInit` contains `AbortSignal`, we should try to merge it (would not use `AbortSignal.any`, since that can cause memory leaks)

dqbd (approved):

js/src/client.ts
langchain-infra: hmm, thie way this works is that if we have multiple concurrent requests they will both queue for this. so i dont know if it makes sense to use the upstream abort signal?

PR #2835, opened 2026-05-01
Add Python profile loading

langchain-infra:
## Summary

Adds LangSmith profile loading to the Python sync and async clients.

- Loads the active profile from `LANGSMITH_CONFIG_FILE` or `~/.langsmith/config.json`
- Selects profiles with `LANGSMITH_PROFILE`, then `current_profile`, then `default`
- Applies profile `api_url`, `workspace_id`, OAuth access tokens, and API keys when explicit/env auth is not set
- Extracts profile parsing and auth refresh logic into `langsmith._internal._profiles`
- Refreshes expired OAuth profile tokens before requests and writes refreshed token fields back to the profile file
- Uses the selected profile's `api_url` for OAuth refresh, even when the client endpoint is overridden
- Reconciles stale profile-managed auth headers after refresh so snapshotted request headers do not send expired OAuth tokens

## Auth precedence

Explicit constructor auth and environment auth take precedence over profile auth. Profile OAuth is preferred over profile `api_key` when profile auth is used. Existing `LANGCHAIN_*` aliases continue to work for existing env vars, but new profile-specific config uses `LANGSMITH_CONFIG_FILE` and `LANGSMITH_PROFILE` only.

## Verification

- `uv run ruff check python/langsmith/client.py python/langsmith/async_client.py python/langsmith/_internal/_profiles.py python/tests/unit_tests/test_client.py python/tests/unit_tests/test_async_client.py`
- `uv run ruff format --check python/langsmith/client.py python/langsmith/async_client.py python/langsmith/_internal/_profiles.py python/tests/unit_tests/test_client.py python/tests/unit_tests/test_async_client.py`
- `PYTHONPATH=python uv run python /private/tmp/profile_auth_regression_check.py`

Not run: targeted pytest, because this local test environment is missing `vcr` (`ModuleNotFoundError: No module named 'vcr'`).

### DIFF
--- a/python/langsmith/async_client.py
+++ b/python/langsmith/async_client.py
@@ -46,10 +46,12 @@ class AsyncClient:
         "_cache",
         "_custom_headers",
         "_api_key",
+        "_oauth_access_token",
     )
 
     _custom_headers: dict[str, str]
     _api_key: Optional[str]
+    _oauth_access_token: Optional[str]
 
     def _compute_headers(self) -> dict[str, str]:
         headers = {**self._custom_headers}
@@ -57,6 +59,8 @@ class AsyncClient:
         headers["Content-Type"] = "application/json"
         if self._api_key:
             headers[ls_client.X_API_KEY] = self._api_key
+        elif self._oauth_access_token:
+            headers["Authorization"] = f"Bearer {self._oauth_access_token}"
         return headers
 
     @property
@@ -124,11 +128,39 @@ class AsyncClient:
         """
         self._retry_config = retry_config or {"max_retries": 3}
         self._custom_headers = headers or {}
-        api_key = ls_utils.get_api_key(api_key)
-        api_url = ls_utils.get_api_url(api_url)
+        env_api_url = ls_client._get_langsmith_env_var_uncached("ENDPOINT")
+        env_api_key = ls_client._get_langsmith_env_var_uncached("API_KEY")
+        profile_config = ls_client._load_profile_client_config(
+            refresh_api_url=api_url if api_url is not None else env_api_url,
+            refresh_oauth=api_key is None,
+        )
+        api_url_ = (
+            api_url if api_url is not None else env_api_url or profile_config.api_url
+        )
+        explicit_or_env_api_key = api_key if api_key is not None else env_api_key
+        profile_auth_enabled = api_key is None and env_api_key is None
+        profile_oauth_access_token = (
+            profile_config.oauth_access_token if profile_auth_enabled else None
+        )
+        api_key_ = (
+            explicit_or_env_api_key
+            if explicit_or_env_api_key is not None
+            else None
+            if profile_oauth_access_token
+            else profile_config.api_key
+        )
+        self._oauth_access_token = (
+            profile_oauth_access_token.strip().strip('"').strip("'")
+            if profile_oauth_access_token
+            else None
+        )
+        api_key = ls_utils.get_api_key(api_key_)
+        api_url = ls_utils.get_api_url(api_url_)
         self._api_key = api_key
         _headers = self._compute_headers()
-        ls_client._validate_api_key_if_hosted(api_url, api_key)
+        ls_client._validate_api_key_if_hosted(
+            api_url, api_key or self._oauth_access_token
+        )
 
         if isinstance(timeout_ms, int):
             timeout_: Union[tuple, float] = (timeout_ms / 1000, None, None, None)
--- a/python/langsmith/async_client.py
+++ b/python/langsmith/async_client.py
@@ -288,6 +288,10 @@ class AsyncClient:
             kwargs["params"] = filtered_params
 
         await self._ensure_profile_auth()
+        if self._profile_auth is not None and "headers" in kwargs:
+            kwargs["headers"] = self._profile_auth.prepare_request_headers(
+                kwargs["headers"]
+            )
 
         for attempt in range(max_retries):
             try:
--- a/python/langsmith/async_client.py
+++ b/python/langsmith/async_client.py
@@ -23,6 +23,7 @@ import httpx
 from langsmith import client as ls_client
 from langsmith import schemas as ls_schemas
 from langsmith import utils as ls_utils
+from langsmith._internal import _profiles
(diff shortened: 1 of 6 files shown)


python/langsmith/async_client.py
```diff
@@ -124,11 +128,39 @@ def __init__(
...
+        explicit_or_env_api_key = api_key if api_key is not None else env_api_key
+        profile_auth_enabled = api_key is None and env_api_key is None
+        profile_oauth_access_token = (
+            profile_config.oauth_access_token if profile_auth_enabled else None
+        )
+        api_key_ = (
+            explicit_or_env_api_key
+            if explicit_or_env_api_key is not None
+            else None
+            if profile_oauth_access_token
```
KiewanVillatel: Is this condition expected? What's the rational?

python/langsmith/async_client.py
KiewanVillatel, line 113: Should we update the doc?

python/langsmith/client.py
```diff
@@ -605,6 +621,172 @@ def close_session(session: requests.Session) -> None:
...
+            profile_name = current_profile
+        elif "default" in profiles:
+            profile_name = "default"
+    if not profile_name or not isinstance(profiles.get(profile_name), dict):
+        return None
+    return path, cast(_ProfileConfigFile, raw), profile_name
+
+
+def _get_langsmith_env_var_uncached(name: str) -> Optional[str]:
+    for namespace in ("LANGSMITH", "LANGCHAIN"):
```
KiewanVillatel, line 596: Not directly related to this PR, but do we have a plan to deprecate one of those? Keeping support for both creates ambiguity (what if both are set?)

python/langsmith/client.py
```diff
@@ -605,6 +621,172 @@ def close_session(session: requests.Session) -> None:
     session.close()
 
 
+def _profile_config_path() -> Optional[Path]:
```
KiewanVillatel: Should we extract a dedicated file / profile service?

python/langsmith/client.py
```diff
@@ -605,6 +621,172 @@ def close_session(session: requests.Session) -> None:
...
+        return False
+    parsed = _parse_profile_expires_at(expires_at)
+    if parsed is None:
+        return False
+    return (
+        parsed <= datetime.datetime.now(datetime.timezone.utc) + _TOKEN_REFRESH_LEEWAY
+    )
+
+
+def _refresh_profile_oauth_token(
```
KiewanVillatel: Should we use a library like `authlib` to manage oauth and token refresh?

python/langsmith/client.py
```diff
@@ -1455,6 +1674,8 @@ def _compute_headers(self) -> dict[str, str]:
         headers.update(self._custom_headers)
         if self.api_key:
```
KiewanVillatel, line 1511: I think we should document the fact that API_KEY takes precedence over oauth

python/langsmith/client.py
```diff
@@ -605,6 +621,172 @@ def close_session(session: requests.Session) -> None:
...
+        temp_path = path.with_name(f"{path.name}.tmp")
+        temp_path.write_text(json.dumps(config, indent=2) + "\n", encoding="utf-8")
+        os.chmod(temp_path, 0o600)
+        os.replace(temp_path, path)
+        os.chmod(path, 0o600)
+    except OSError:
+        return
+
+
+def _load_profile_client_config(
```
KiewanVillatel: This is doing more than loading a config, there is actual business logic inside, I'd disentangle the 2.

python/langsmith/client.py
```diff
@@ -605,6 +621,172 @@ def close_session(session: requests.Session) -> None:
...
+    path, config, profile_name = state
+    profiles = config.get("profiles") or {}
+    profile = profiles[profile_name]
+    env_auth_set = bool(_get_langsmith_env_var_uncached("API_KEY"))
+    refresh_url = (
+        refresh_api_url
+        or _get_langsmith_env_var_uncached("ENDPOINT")
+        or profile.get("api_url")
+    )
+    if refresh_oauth and not env_auth_set and _should_refresh_profile_token(profile):
```
KiewanVillatel: Why do we refresh the token here? Shouldn't we do that before the requests instead?

python/langsmith/client.py
```diff
@@ -605,6 +621,172 @@ def close_session(session: requests.Session) -> None:
...
+        refresh_token = (profile.get("oauth") or {}).get("refresh_token")
+        if refresh_token:
+            token = _refresh_profile_oauth_token(refresh_url, refresh_token)
+            if token is not None:
+                _apply_profile_token_response(profile, token)
+                profiles[profile_name] = profile
+                config["profiles"] = profiles
+                _save_profile_config(path, config)
+    oauth_access_token = (profile.get("oauth") or {}).get("access_token")
+    return _ProfileClientConfig(
```
KiewanVillatel: Related to the previous comment, I think the config should contain everything: `api_key`, `oauth_access_token` and `oauth_refresh_token`.

Then we could extract the auth business logic in a dedicated service. The service could expose a `get_auth_header`, and this is where we would decide if we use the API key or oauth, and for oauth it would refresh the access token if needed. `get_auth_header` would be used before each API call.

WDYT?

ramon-langchain (approved):

ericdong-langchain (approved):

python/langsmith/client.py
langchain-infra: yea agreed, done

python/langsmith/client.py
langchain-infra: refactored

python/langsmith/client.py
langchain-infra: yes i have a docs page for profiles

python/langsmith/client.py
langchain-infra: mm i am somewhat against introducing mroe deps due to all the supply chain attacks right now

python/langsmith/client.py
langchain-infra: i am not sure about plan

PR #2846, opened 2026-05-05
Extract JS profile auth service

langchain-infra:
## Summary
- Extract JS profile loading, auth header selection, and OAuth refresh into a dedicated profile auth service.
- Keep loaded profile config explicit for api_key, oauth_access_token, and oauth_refresh_token; refresh expired OAuth tokens before requests and persist refreshed tokens.
- Preserve explicit per-request auth while replacing stale profile-managed bearer headers, and keep profile file loading disabled in browser-like runtimes.

## Testing
- pnpm exec oxfmt --write src/client.ts src/utils/profiles.ts src/tests/client.test.ts
- pnpm run format:check
- pnpm run lint (exits 0; reports existing warnings)
- pnpm exec oxlint src/client.ts src/utils/profiles.ts src/tests/client.test.ts (exits 0; reports existing client.ts warnings)
- pnpm run check:types
- env -u LANGSMITH_API_KEY -u LANGCHAIN_API_KEY -u LANGSMITH_ENDPOINT -u LANGCHAIN_ENDPOINT -u LANGSMITH_WORKSPACE_ID -u LANGCHAIN_WORKSPACE_ID -u LANGSMITH_PROFILE LANGSMITH_CONFIG_FILE=/nonexistent/path/config.json pnpm test:single -- src/tests/client.test.ts --runInBand
- pnpm run build:esm
- pnpm test (in js/internal/environment_tests/test-exports-esbuild)
- local profile smoke against /info: 200

### DIFF
--- a/js/src/utils/profiles.ts
+++ b/js/src/utils/profiles.ts
@@ -221,7 +221,7 @@ export function loadProfileClientConfig(): ProfileClientConfig {
 export class ProfileAuth {
   private refreshPromise?: Promise<void>;
 
-  private managedAuthorizationValues = new Set<string>();
+  private managedAuthorizationValue?: string;
 
   constructor(private state: LangSmithProfileState) {
     this.rememberProfileAuthHeader(this.currentAuthHeader());
@@ -253,7 +253,7 @@ export class ProfileAuth {
   }
 
   isProfileAuthorizationHeader(value: string): boolean {
-    return this.managedAuthorizationValues.has(value);
+    return value === this.managedAuthorizationValue;
   }
 
   private async refreshOAuthToken(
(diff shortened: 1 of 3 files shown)


js/src/utils/profiles.ts
```diff
@@ -0,0 +1,342 @@
...
+      apiKey || oauthAccessToken || oauthRefreshToken
+        ? new ProfileAuth(state)
+        : undefined,
+  };
+}
+
+export class ProfileAuth {
+  private refreshPromise?: Promise<void>;
+
+  private managedAuthorizationValues = new Set<string>();
```
ericdong-langchain: This grows everytime token is refreshed, should clear old values on refresh or just track the single most recent value instead of a set.

langchain-infra: Fixed in ee6f6bdd. This now tracks only the current profile-managed Authorization value; _fetch captures the pre-refresh managed value for the active request so stale _mergedHeaders snapshots are still replaced after refresh.

ericdong-langchain (approved):

PR #2935, opened 2026-05-26
fix(python): enforce size_limit_bytes in multipart ingest path

KiewanVillatel:
## Summary

- `_multipart_ingest_ops` now splits ops into multiple requests when the accumulated payload would exceed `size_limit_bytes`, mirroring the batch-splitting logic that already existed in `_batch_ingest_run_ops`
- Adds `_multipart_part_size` helper in `_multipart.py` to estimate the byte size of a `MultipartPartsAndContext` (uses `len(data)` for bytes parts, `os.fstat` for file attachments)
- Size limit respects the same hierarchy: `_max_batch_size_bytes` > server `batch_ingest_config.size_limit_bytes` > `_SIZE_LIMIT_BYTES` (20 MB default)
- A single op larger than the limit is still sent (the check requires `batch_size > 0` before splitting, preserving current behavior for oversized individual ops)

## Linear

https://linear.app/langchain/issue/LSDK-207/multipart-ingest-path-has-no-batch-size-enforcement-allowing

## Test Plan

- [x] `test_multipart_ingest_ops_splits_by_size` — calls `_multipart_ingest_ops` with 3 × 500-byte ops against a 600-byte limit; asserts 3 separate `_send_multipart_req` calls
- [x] `test_multipart_ingest_ops_no_split_when_within_limit` — 2 × 100-byte ops against 10 KB limit; asserts 1 `_send_multipart_req` call
- [x] `test_batch_ingest_run_splits_large_batches[True-*]` — updated expected request count to use the same formula as the `False` (non-multipart) parameterized variant, since both paths now enforce the same limit
- [x] Full unit test suite: 1208 passed, 10 pre-existing failures unrelated to this change (sandbox + test_utils env-var tests)

## Release Note

Fixed an issue where the multipart ingest path could send arbitrarily large requests when many runs with large payloads were batched together, causing OOM crashes in high-throughput environments.

### DIFF
--- a/python/langsmith/_internal/_multipart.py
+++ b/python/langsmith/_internal/_multipart.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
 from collections.abc import Iterable
 from io import BufferedReader
 from typing import Union
@@ -18,6 +19,20 @@ class MultipartPartsAndContext:
     def __init__(self, parts: list[MultipartPart], context: str) -> None:
         self.parts = parts
         self.context = context
+
+
+def _multipart_part_size(part: MultipartPartsAndContext) -> int:
+    """Estimate byte size of a MultipartPartsAndContext for batch splitting."""
+    total = 0
+    for _, (_, data, _, _) in part.parts:
+        if isinstance(data, bytes):
+            total += len(data)
+        elif isinstance(data, BufferedReader):
+            try:
+                total += os.fstat(data.fileno()).st_size
+            except Exception:
+                pass
+    return total
 
 
 def join_multipart_parts_and_context(
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -95,6 +95,7 @@ from langsmith._internal._hub import (
 from langsmith._internal._multipart import (
     MultipartPart,
     MultipartPartsAndContext,
+    _multipart_part_size,
     join_multipart_parts_and_context,
 )
 from langsmith._internal._operations import (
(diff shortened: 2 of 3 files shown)


python/langsmith/client.py
```diff
@@ -2974,52 +2975,88 @@ def _multipart_ingest_ops(
...
                         api_key=api_key,
                         service_key=service_key,
                         tenant_id=tenant_id,
                         authorization=authorization,
                         cookie=cookie,
                     )
-            finally:
-                _close_files(list(opened_files_dict.values()))
+                except ls_utils.LangSmithNotFoundError:
+                    # Fallback to batch ingest if multipart endpoint returns 404
```
QuentinBrosse, line 3041: Do we need that?

python/langsmith/_internal/_multipart.py
```diff
@@ -20,6 +21,20 @@ def __init__(self, parts: list[MultipartPart], context: str) -> None:
         self.context = context
 
 
+def _multipart_part_size(part: MultipartPartsAndContext) -> int:
+    """Estimate byte size of a MultipartPartsAndContext for batch splitting."""
+    total = 0
+    for _, (_, data, _, _) in part.parts:
+        if isinstance(data, bytes):
+            total += len(data)
+        elif isinstance(data, BufferedReader):
```
QuentinBrosse, line 30: Is it checking for attachments? I am not sure we should as attachements have a higher size limit.
The multipart endpoint shouldn't error on a small JSON payload with large attachments.

PR #2952, opened 2026-05-28
release(py): 0.8.7

vishnu-ssuresh:
## What

Bump the LangSmith Python SDK version to 0.8.7.

## Why

The Dockerfile sandbox snapshot API has merged to main and needs a package release so downstream deploys can depend on it from PyPI.

## How

Updated the Python SDK version source and bumpversion config from 0.8.6 to 0.8.7.

## Testing

- Verified python/langsmith/__init__.py and python/.bumpversion.cfg both contain 0.8.7
- Confirmed tag v0.8.7 does not exist on origin

### DIFF
--- a/python/.bumpversion.cfg
+++ b/python/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.8.6
+current_version = 0.8.7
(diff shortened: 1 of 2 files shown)


vishnu-ssuresh:
Closing in favor of the other Python SDK version bump PR.

PR #2954, opened 2026-05-29
feat(evaluation): adding examples and repetitions to create tracer sessions calls

shamikkarkhanis:
## Description
LSE-2266. Adds support for experiment loading state tracking by including num_examples and num_repetitions in session creation call to langsmith. 

## Release Notes
Adds num_examples and num_repetitions as optional experiment fields for py and js SDKs.

## Test Plan
- [x] cd python && uv run pytest tests/unit_tests/evaluation/test_runner.py -k "test_resolve_num_examples or test_evaluate_sends_num_examples" -v --PASSES
- [x] cd js && pnpm test src/tests/evaluate_runner.test.ts -t "evaluate forwards num_examples" --PASSES


python/langsmith/client.py
```diff
@@ -4696,16 +4696,25 @@ def create_project(
...
         reference_dataset_id: Optional[ID_TYPE] = None,
+        num_examples: Optional[int] = None,
+        num_repetitions: Optional[int] = None,
     ) -> ls_schemas.TracerSession:
         """Create a project on the LangSmith API.
 
         Args:
             project_name (str): The name of the project.
-            project_extra (Optional[dict]): Additional project information.
+            project_extra (Optional[dit]): Additional project information.
```
QuentinBrosse: typo?

python/langsmith/evaluation/_runner.py
QuentinBrosse, line 1884: Do we need to add the new fields here?

shamikkarkhanis: Currently not breaking as we only use the fields before _copy is ever called but it could lead to inconsistencies where the fields are 'none' after being copied. will add them, nice catch !

python/langsmith/client.py
```diff
@@ -4696,6 +4696,8 @@ def create_project(
         upsert: bool = False,
         project_extra: Optional[dict] = None,
         reference_dataset_id: Optional[ID_TYPE] = None,
+        num_examples: Optional[int] = None,
+        num_repetitions: Optional[int] = None,
```
baskaryan, line 4700: do we need to add these as their own args, vs putting them in metadata? are these standalone request params that create session endpoint now accepts?

python/langsmith/evaluation/_runner.py
```diff
@@ -1983,6 +2008,59 @@ def _is_valid_uuid(value: str) -> bool:
...
+    """
+    try:
+        if isinstance(data, schemas.Dataset):
+            if data.example_count is not None:
+                return data.example_count
+            return client.read_dataset(dataset_id=data.id).example_count
+        if isinstance(data, uuid.UUID):
+            return client.read_dataset(dataset_id=data).example_count
+        if isinstance(data, str):
+            if _is_valid_uuid(data):
```
baskaryan: is this supposed to be `if not _is_valid_uuid(data)`?

shamikkarkhanis: logic was a bit confusing, cleaned it up in the newest commit

baskaryan (approved):

PR #2975, opened 2026-06-03
feat(python): expose pairwise comparison URL on ComparativeExperimentResults [LSE-2270]

harisaiharish:
## Summary
  - Expose the pairwise comparison URL on `ComparativeExperimentResults` via new `.url` and
  `.comparative_experiment` properties, so non-interactive callers (CI, remote boxes) can
  capture it programmatically instead of scraping the stdout line.
  - Factor the URL construction out of `_print_comparative_experiment_start()` into a
  `_build_comparative_url(experiments, comparative_experiment)` helper; `evaluate_comparative()`
  builds the URL once and passes it to both the printer and the result object - no duplicated
  logic.
  - Purely additive: the printed `View the pairwise evaluation results at:` output is unchanged,
  and the two new `ComparativeExperimentResults` constructor args default to `None`, so
  existing callers are unaffected.

  ## Release Note
  `evaluate_comparative()` now returns the pairwise comparison URL and underlying comparative
  experiment on the result object (`results.url`, `results.comparative_experiment`).

  ## Test Plan
  - [x] `env -u LANGCHAIN_PROJECT -u LANGCHAIN_API_KEY -u LANGCHAIN_TRACING_V2 -u
  LANGSMITH_TRACING uv run python -m pytest --disable-socket --allow-unix-socket -n auto
  tests/unit_tests/evaluation/test_runner.py`: 81 passed (includes 3 new tests:
  `test_build_comparative_url`, `test_build_comparative_url_no_host_returns_none`,
  `test_comparative_experiment_results_exposes_url_and_experiment`)
  - [x] `uv run ruff check langsmith/evaluation/_runner.py
  tests/unit_tests/evaluation/test_runner.py`
  - [x] `uv run ruff format --check langsmith/evaluation/_runner.py
  tests/unit_tests/evaluation/test_runner.py`
  - [x] `uv run mypy langsmith/evaluation/_runner.py`
  - [x] Prod Python e2e: created dataset, ran `evaluate()` on two predictors, then
  `evaluate_comparative()`. Confirmed `results.url` resolves to the live compare view
  (comparative experiment `c9218d6a-2dde-44b6-99e0-09730b50f5a7`, dataset
  `d1bb2d43-4fac-41a0-b899-bb092de4b3ef`) and `results.comparative_experiment` exposes the
  matching IDs.

  ## Notes
  - Python SDK only. The JS SDK has an analogous `evaluateComparative`; JS parity could be a
  follow-up if desired.

PR #2981, opened 2026-06-03
feat(js): exposed pairwise comparison URL on ComparativeExperimentResults [LSE-2374]

harisaiharish:
## Summary
  - Expose the pairwise comparison URL on `ComparisonEvaluationResults` via new `url` and
  `comparativeExperiment` fields, so non-interactive callers (CI, remote boxes) can capture it
  programmatically instead of scraping the `console.log` output. JS parity for the Python change
  in #2975.
  - The URL was already computed into a `viewUrl` variable for logging in
  `evaluateComparative()`; this simply returns it (plus the underlying comparative experiment)
  on the result object instead of discarding it — no duplicated logic.
  - Purely additive: the `View results at:` log output is unchanged, and the only constructor of
  `ComparisonEvaluationResults` is `evaluateComparative()` itself (re-returned unchanged by
  `evaluate(...)`'s two-experiment overload), so no other call site is affected.

  ## Release Note
  JS `evaluateComparative()` now returns the pairwise comparison URL and underlying comparative
  experiment on the result object (`results.url`, `results.comparativeExperiment`).

  ## Test Plan
  - [x] `pnpm run check:types`
  - [x] `pnpm exec oxfmt --check src/evaluation/evaluate_comparative.ts
  src/tests/evaluate.int.test.ts`
  - [x] `pnpm exec oxlint src/evaluation/evaluate_comparative.ts src/tests/evaluate.int.test.ts`
  - [x] `pnpm run build`
  - [x] Prod JS e2e: `pnpm run test:single src/tests/evaluate.int.test.ts -t "comparative
  target"` — passed. Ran `evaluate()` on two targets then comparative `evaluate()`; confirmed
  `results.url` resolves to the live compare view (comparative experiment
  `83fdfd1b-1021-46ec-8e77-1c758a584f40`, dataset `5fce1ecd-7e6b-424f-b998-2acb2f8a140c`) and
  `results.comparativeExperiment` exposes the matching IDs. Test now asserts both fields are
  present and that the URL embeds the comparative-experiment ID.

### DIFF
--- a/js/src/evaluation/evaluate_comparative.ts
+++ b/js/src/evaluation/evaluate_comparative.ts
@@ -1,6 +1,7 @@
 import { v4 as uuid4, validate } from "../utils/uuid/src/index.js";
 import { Client } from "../index.js";
 import {
+  ComparativeExperiment,
   ComparisonEvaluationResult as ComparisonEvaluationResultRow,
   Example,
   Run,
@@ -132,6 +133,10 @@ export interface EvaluateComparativeOptions {
 export interface ComparisonEvaluationResults {
   experimentName: string;
   results: ComparisonEvaluationResultRow[];
+  /** URL of the pairwise comparison view in the LangSmith UI, if available. */
+  url: string | null;
+  /** The comparative experiment, exposing its id, dataset, and metadata. */
+  comparativeExperiment: ComparativeExperiment;
 }
 
 /** @deprecated Use `evaluate` and pass two experiments as targets. */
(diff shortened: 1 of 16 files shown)


js/src/evaluation/evaluate_comparative.ts
```diff
@@ -132,6 +133,10 @@ export interface EvaluateComparativeOptions {
 export interface ComparisonEvaluationResults {
   experimentName: string;
   results: ComparisonEvaluationResultRow[];
```
dqbd, line 137: _nit_: would add JSDoc here as well

dqbd (approved):

PR #3008, opened 2026-06-10
fix: avoid duplicate /v1 in hub URLs when endpoint includes /api/v1

langchain-infra:
## Description
When `LANGSMITH_ENDPOINT` ends in `/api/v1`, the hardcoded `/v1/platform/hub/repos` hub path produced `/api/v1/v1/platform/...`, causing `push_agent`, `pull_agent`, and `_hub_repo_exists` to 404. This routes hub paths through the existing platform-path helpers (Python `platform_hub_path`, JS `_getPlatformEndpointPath`) which only add the `/v1` prefix when the API URL doesn't already end in `/v1`. Applies the fix to both the Python and JS SDKs and documents the no-hardcoded-`/v1` convention in `AGENTS.md`.

## Release Note
Fix hub agent/skill methods (push/pull/exists) failing with 404 when `LANGSMITH_ENDPOINT` includes `/api/v1` (Python and JS SDKs).

## Test Plan
- [ ] Configure a client with `LANGSMITH_ENDPOINT=https://.../api/v1` and confirm `pullAgent`/`pushAgent` hit `/api/v1/platform/hub/repos/...` (single `/v1`).

Made by [Open SWE](https://openswe.vercel.app)

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "langsmith",
-  "version": "0.7.5",
+  "version": "0.7.6",
   "description": "Client library to connect to the LangSmith Observability and Evaluation Platform.",
   "packageManager": "pnpm@10.33.0",
   "files": [
--- a/js/package.json
+++ b/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "langsmith",
-  "version": "0.7.6",
+  "version": "0.7.5",
   "description": "Client library to connect to the LangSmith Observability and Evaluation Platform.",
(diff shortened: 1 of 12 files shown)


ericdong-langchain (approved):

js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "langsmith",
```
dqbd, line 2: Can we bump in a separate PR?

langchain-infra: fixed

PR #3014, opened 2026-06-10
fix(experiments): extract keys from wrapped evaluator function

shamikkarkhanis:
## Description
Addresses LSE-2416 which investigates why evaluator progress was missing when kicking an experiment from the sdk. Issue was identified when defining an evaluator using the documented `(inputs, outputs, reference_outputs)` convention which transforms the evaluator method before extraction. Solution is to include the original function as an attribute of the wrapped function and use original function for extraction. Now works for both `(inputs, outputs, reference_outputs)`  and `(run, example)` defined evaluators. 

The JS sdk presents the same issue where evaluator keys are extracted as an incorrect name and are therefore not updated properly. This leads to 0/n runs completed on the LangSmith UI. The cause is the lack of a proper `_extractEvaluatorFeedbackKeys` method which we've implemented in this PR as well. 

## Release Note
Fix 0/n runs evaluated count in LangSmith when kicking off experiments from the python sdk. 

## Test Plan
- [x] make tests TEST="tests/unit_tests/evaluation/test_runner.py tests/unit_tests/evaluation/test_evaluator.py" --PASSES 
- [x] NODE_OPTIONS=--experimental-vm-modules pnpm jest --config jest.config.cjs src/tests/evaluate_runner.test.ts --PASSES

### DIFF
--- a/python/langsmith/evaluation/_runner.py
+++ b/python/langsmith/evaluation/_runner.py
@@ -2249,6 +2249,8 @@ def _extract_feedback_keys(evaluator: RunEvaluator):
 
 
 def _extract_code_evaluator_feedback_keys(func: Callable) -> list[str]:
+    # Unwrap to the user's original function so key extraction reads their source.
+    func = getattr(func, "__wrapped_evaluator_func__", func)
     python_code = inspect.getsource(func)
 
     def extract_dict_keys(node):
--- a/python/langsmith/evaluation/_runner.py
+++ b/python/langsmith/evaluation/_runner.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import ast
 import collections
 import concurrent.futures as cf
 import functools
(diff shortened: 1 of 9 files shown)


python/langsmith/evaluation/_runner.py
```diff
@@ -2249,6 +2249,8 @@ def _extract_feedback_keys(evaluator: RunEvaluator):
 
 
 def _extract_code_evaluator_feedback_keys(func: Callable) -> list[str]:
+    # Unwrap to the user's original function so key extraction reads their source.
+    func = getattr(func, "__wrapped_evaluator_func__", func)
```
baskaryan: should we extract feedback keys before we wrap the evaluator func, to avoid having to do this?

shamikkarkhanis: addressed

baskaryan (approved):

PR #3018, opened 2026-06-11
feat: expose the resources from the generated openapi client in the langsmith client

sineha-mani:
**Description**

- Adds client.online_evaluators in Python, backed by the generated OpenAPI OnlineEvaluatorsResource.
- Adds client.onlineEvaluators in JS with CRUD-style methods for create, retrieve, update, list, delete, bulk delete, and spend.

Python
Screenshot 2026-06-11 at 12 08 04 PM
Screenshot 2026-06-11 at 12 08 11 PM

JS
Screenshot 2026-06-11 at 12 10 27 PM
Screenshot 2026-06-11 at 12 10 39 PM

References [LSE-2426](https://linear.app/langchain/issue/LSE-2426/add-first-class-sdk-methods-for-online-evaluators)

**Test Plan**
[X] Check if SDK can be used for online eval crud ops for users.
[X] Tested Python and JS by running local smoke scripts against the smith-go backend on http://localhost:8080, creating an online code evaluator, retrieving it, listing evaluators, inspecting it manually, and deleting it after confirmation.
[X] Added unit tests.

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -1395,6 +1395,13 @@ export class Client implements LangSmithTracingClientInterface {
       this.apiKey === undefined && this.workspaceId === undefined
         ? { "X-API-Key": null }
         : undefined;
+    const {
+      method: _method,
+      headers: _headers,
+      body: _body,
+      signal: _signal,
+      ...openAPIFetchOptions
+    } = this.fetchOptions;
 
     return new OpenAPILangsmith({
       apiKey: this.apiKey,
@@ -1402,7 +1409,7 @@ export class Client implements LangSmithTracingClientInterface {
       baseURL: this._getOpenAPIBaseUrl(),
       timeout: this.timeout_ms,
       fetch: this._fetch,
-      fetchOptions: this.fetchOptions,
+      fetchOptions: openAPIFetchOptions,
       defaultHeaders,
     });
   }
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -74,6 +74,8 @@ import {
 
 import { EvaluationResult, EvaluationResults } from "./evaluation/evaluator.js";
 import { __version__ } from "./index.js";
+import { Langsmith as OpenAPILangsmith } from "./_openapi_client/index.js";
+import { OnlineEvaluators } from "./_openapi_client/resources/online-evaluators.js";
 import { assertUuid } from "./utils/_uuid.js";
 import { warnOnce } from "./utils/warn.js";
 import { parseHubIdentifier } from "./utils/prompts.js";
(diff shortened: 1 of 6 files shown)


js/src/client.ts
```diff
@@ -1388,6 +1404,77 @@ export class Client implements LangSmithTracingClientInterface {
...
+      await raiseForStatus(res, `Failed to ${method} ${path}`);
+      return res;
+    });
+    if (response.status === 204) {
+      return undefined as T;
+    }
+    return response.json() as Promise<T>;
+  }
+
+  public get onlineEvaluators(): OnlineEvaluatorsClient {
```
KiewanVillatel: We should use the generated open api client in `_openapi_client`  instead.

python/langsmith/client.py
```diff
@@ -1610,6 +1620,24 @@ def info(self) -> ls_schemas.LangSmithInfo:
 
         return self._info
 
+    @property
+    def online_evaluators(self) -> OnlineEvaluatorsResource:
+        """Access generated online evaluator CRUD methods."""
+        from langsmith._openapi_client import Langsmith as OpenAPILangsmith
+
+        self._ensure_profile_auth()
```
KiewanVillatel, line 1628: It should work and we can go like this for a first version. But we ideally we should delegate the OAuth flow to the open api client (should be supported) and instantiate only once the client.

python/tests/unit_tests/test_client.py
```diff
@@ -798,6 +798,30 @@ def test_create_run_unicode() -> None:
     client.update_run(id_, status="completed")
 
 
+def test_online_evaluators_uses_generated_openapi_resource() -> None:
```
KiewanVillatel, line 801: This is the first time we are using the generated open api client, would it be possible to also add an integration test?

sineha-mani: Added

js/src/client.ts
```diff
@@ -1381,13 +1383,28 @@ export class Client implements LangSmithTracingClientInterface {
...
+
   private _getPlatformEndpointPath(path: string): string {
     // Check if apiUrl already ends with /v1 or /v1/ to avoid double /v1/v1/ paths
     const needsV1Prefix =
       this.apiUrl.slice(-3) !== "/v1" && this.apiUrl.slice(-4) !== "/v1/";
     return needsV1Prefix ? `/v1/platform/${path}` : `/platform/${path}`;
   }
 
+  public get onlineEvaluators(): OnlineEvaluators {
+    return new OpenAPILangsmith({
```
KiewanVillatel, line 1406: Should we instantiate only once the client in the constructor?

python/langsmith/client.py
```diff
@@ -1610,6 +1620,24 @@ def info(self) -> ls_schemas.LangSmithInfo:
...
+        headers = self._headers
+        if self._profile_auth is not None:
+            headers = self._profile_auth.prepare_request_headers(headers)
+
+        return OpenAPILangsmith(
```
KiewanVillatel, line 1633: Could we already extract a method to instantiate the client with proper auth headers (eg. `_new_openapi_client`)?
We'll need it when adding more resources.

js/src/client.ts
sineha-mani: resolved to instantiate only once per client in ctr!

python/langsmith/client.py
sineha-mani: Extracted this method!

python/langsmith/client.py
```diff
@@ -147,6 +148,12 @@ def _reset_tracing_drop_log() -> None:
         _tracing_drops_last_log_time = 0.0
 
 
+def _get_openapi_base_url(api_url: str) -> str:
+    """Convert a handwritten client API URL to a generated OpenAPI base URL."""
+    api_url = api_url.rstrip("/")
+    return api_url[:-3] if api_url.endswith("/v1") else api_url
```
KiewanVillatel, line 154: Just wondering, why is this needed? In which case the `api_url` would end with `/v1`?

KiewanVillatel (approved):

PR #3021, opened 2026-06-11
fix(sandbox): deliver output callbacks across stream reconnects

langchain-infra:
## Problem

`sandbox.run(..., { onStdout })` (Python: `on_stdout`) silently drops all output produced after a mid-stream WebSocket reconnect. The callbacks are invoked inside the **per-connection** message pump (`runWsStream` / `run_ws_stream`), but `CommandHandle`'s auto-reconnect swaps in a fresh connection via `reconnectWsStream`, which has no callback support — so after the first transient disconnect (LB idle timeout, flaky WAN), callbacks go quiet while the command result still resolves with `exit_code 0`. Silent tail truncation.

Reported by a user migrating from E2B (follow-up to the server-side ring-buffer fix in langchainplus#27168): their repro using the callback path lost 3,516 of 29,297 records on 2/3 runs over their network, with no error.

## Fix

Move callback invocation into `CommandHandle` / `AsyncCommandHandle`'s reconnecting iterator — the one place that survives reconnects and sees every chunk exactly once — and stop passing callbacks to the per-connection stream functions. Applied to the JS handle and both sync/async Python handles (the JS handle is a port of the Python one; all three had the bug).

Also fixes `RunOptions.commandId` being silently dropped in the JS WS path before reaching the execute payload.

## Validation

- New unit tests in both SDKs: callbacks invoked for every chunk, and specifically for chunks delivered after a simulated mid-stream reconnect (fail before the fix, pass after). JS: 100 sandbox tests pass; Python: 444 sandbox tests pass.
- End-to-end against prod with a deterministic disconnect probe (30 MB JSONL stream via `run({ onStdout })`, hard `ws.terminate()` every 5 MB):
  - **Before:** 1 disconnect → 4,889 of 29,297 records delivered, `exit_code 0` (silent loss).
  - **After:** 6 forced disconnects → all 29,297 records delivered, `exit_code 0`.
- The reporting user's updated repro (`mwcd/langsmith-sandbox-stream-repro@8589c98`) passes with this fix.

### DIFF
--- a/js/src/sandbox/command_handle.ts
+++ b/js/src/sandbox/command_handle.ts
@@ -54,6 +54,8 @@ export class CommandHandle {
   private _lastStdoutOffset: number;
   private _lastStderrOffset: number;
   private _started: boolean;
+  private _onStdout?: (data: string) => void;
+  private _onStderr?: (data: string) => void;
 
   /** @internal */
   constructor(
@@ -64,6 +66,8 @@ export class CommandHandle {
       commandId?: string;
       stdoutOffset?: number;
       stderrOffset?: number;
+      onStdout?: (data: string) => void;
+      onStderr?: (data: string) => void;
     },
   ) {
     this._stream = messageStream;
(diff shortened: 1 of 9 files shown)


langchain-infra:
Split per language for separate release trains: JS → #3022, Python → #3023. Identical content; closing this combined PR.

PR #3026, opened 2026-06-12
Refactor/precompute evaluator feedback keys

baskaryan:
- pre-compute feedback_keys and expose as property on DynamicEvaluator
- move key extraction util into its own file (no functional change) 
- add key extraction unit tests


python/langsmith/evaluation/_runner.py
```diff
@@ -2236,10 +2223,8 @@ def _get_random_name() -> str:
 
 def _extract_feedback_keys(evaluator: RunEvaluator):
```
baskaryan: should maybe just make feedback_keys a property on RunEvaluator

PR #3041, opened 2026-06-16
feat(client): warn when backend version is below minimum required

KiewanVillatel:
## Description

Adds a backend version compatibility check to both the Python sync/async and TypeScript LangSmith clients, and extracts the minimum version into a dedicated constants file for easy CI updates.

### Python

- Extracted `_MIN_BACKEND_VERSION = "0.16.0"` into `python/langsmith/_internal/_constants.py` as the single source of truth
- Added module-level `_check_backend_version(version: str)` in `client.py`, shared between sync and async clients; uses `packaging.version.parse` so pre-release tags like `0.16.4rc1` are handled correctly
- Called in `Client.online_evaluators` property
- `AsyncClient`:
  - Added `async def info()` that fetches and caches `/info`, falling back to an empty `LangSmithInfo()` on error
  - `__aenter__` pre-fetches info and calls `_check_backend_version` once on context-manager entry
  - `online_evaluators` stays a sync `@property` so callers can do `client.online_evaluators.method()`

### TypeScript

- Added `js/src/utils/constants.ts` with `export const _MIN_BACKEND_VERSION = "0.16.0"`
- Added exported `_checkBackendVersion(version, minVersion?)` in `client.ts`; uses `parseInt(s, 10)` (not `Number()`) to correctly parse pre-release suffixes like `rc1`
- Called inside `_ensureServerInfo()` after a successful `/info` fetch

### Tests

- Python: parametrized `test_check_backend_version` (8 cases including pre-release), plus async-specific tests for `info()` caching, error fallback, and the `__aenter__` version warning
- TypeScript: `describe("_checkBackendVersion")` block with 8 `it.each` cases
- Both test suites mock `_MIN_BACKEND_VERSION` to `"0.5.0"` so tests won't break when the constant is bumped by CI

Closes LIN-386

## Self-Hosted Release Note
none

## Test Plan
- [x] Python linter and type checks passing
- [x] TypeScript type checks passing (`tsc --noEmit`)
- [x] Unit tests passing (Python and TypeScript) with mocked version constant


python/langsmith/async_client.py
```diff
@@ -115,6 +120,23 @@ def workspace_id(self, value: Optional[str]) -> None:
         self._workspace_id = ls_utils.get_workspace_id(value)
         self._client.headers = httpx.Headers(self._compute_headers())
 
+    async def info(self) -> ls_schemas.LangSmithInfo:
```
QuentinBrosse, line 124: Should we use `@property` to match the sync client?

python/langsmith/client.py
```diff
@@ -312,6 +313,24 @@ class TracerProvider:  # type: ignore[no-redef]
 _fallback_dirs_created: set[str] = set()
 
 
+def _check_backend_version(version: str) -> None:
```
QuentinBrosse: nit: should we move it to a dedicated file to avoid importing client in async_client?

js/src/client.ts
```diff
@@ -700,6 +701,37 @@ function _formatFeedbackScore(score?: ScoreType): ScoreType | undefined {
   return score;
 }
 
+export function _checkBackendVersion(
+  version: string,
+  minVersion: string = _MIN_BACKEND_VERSION,
+): void {
+  const parse = (v: string) => v.split(".").map((s) => parseInt(s, 10));
```
QuentinBrosse, line 708: nit: we use the `semver` package in `js/scripts/check-npm-version.js`, not sure if we want to add it as a run-time dependency and use it. No strong opinion.

QuentinBrosse (approved):

python/langsmith/async_client.py
KiewanVillatel: Unfortunately the method is async so we cannot use a property.

python/langsmith/client.py
KiewanVillatel: Makes sense, will do.

js/src/client.ts
KiewanVillatel: The code is relatively simple so IMO it's worth avoiding adding another runtime dependency.

python/langsmith/client.py
KiewanVillatel: Done

PR #3043, opened 2026-06-16
fix: update JS SDK type declarations with skipLibCheck disabled

sineha-mani:
The published JS package emitted broken _openapi_client declarations when consumers typechecked dependencies with skipLibCheck: false: NullableHeaders was stripped from .d.ts output, and declarations referenced optional fetch types through invalid relative node_modules paths.

This PR patches the emitted declarations during the JS package build, without modifying the generated _openapi_client source, and adds an ESM environment regression test that typechecks import * as langsmith from "langsmith" with skipLibCheck: false.

Reproduced with a fresh npm project installing the published package, then verified the same repro passes when installing a locally packed fixed langsmith package.

**Simple descr:**
Some TypeScript users couldn’t compile projects that imported langsmith because our published package included broken generated type files.

This PR fixes the package build so those type files are cleaned up before release, without changing the generated source directly.

I also added a regression test that recreates the customer setup, so future releases should catch this before publishing.

The issue happened with these steps to reproduce:

mkdir langsmith-broken-types
cd langsmith-broken-types
npm init -y
npm pkg set type=module
npm i typescript langsmith
npx tsc --init --skipLibCheck false --exactOptionalPropertyTypes false
echo "import * as langsmith from 'langsmith'; console.log(langsmith);" > main.ts
npx tsc --noEmit

To test run this local equivalent against your fixed checkout:

cd /Users/sineha/Desktop/langchain/langsmith-sdk/js
pnpm run build
PACK_FILE=$(npm pack --silent)
mkdir /tmp/langsmith-broken-types-local
cd /tmp/langsmith-broken-types-local
npm init -y
npm pkg set type=module
npm i typescript "/Users/sineha/Desktop/langchain/langsmith-sdk/js/$PACK_FILE"
npx tsc --init --skipLibCheck false --exactOptionalPropertyTypes false
echo "import * as langsmith from 'langsmith'; console.log(langsmith);" > main.ts
npx tsc --noEmit

### DIFF
--- a/js/internal/environment_tests/test-exports-esm/package.json
+++ b/js/internal/environment_tests/test-exports-esm/package.json
@@ -5,13 +5,14 @@
   "type": "module",
   "description": "Tests that langsmith exports work correctly in ESM environments",
   "scripts": {
-    "test": "npx tsx src/index.ts"
+    "test": "tsc --noEmit && tsx src/index.ts"
   },
   "dependencies": {
     "langsmith": "file:../../.."
   },
   "devDependencies": {
+    "@types/node": "^20.0.0",
     "tsx": "^4.7.0",
-    "@types/node": "^20.0.0"
+    "typescript": "^6.0.3"
   }
 }
--- a/js/scripts/fix-openapi-client-declarations.js
+++ b/js/scripts/fix-openapi-client-declarations.js
@@ -1,0 +1,54 @@
+import { readFile, writeFile } from "node:fs/promises";
+import { dirname, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+
+function abs(relativePath) {
+  return resolve(dirname(fileURLToPath(import.meta.url)), relativePath);
+}
+
+async function patchFile(relativePath, replacements) {
+  const filePath = abs(relativePath);
+  let content = await readFile(filePath, "utf8");
+
+  for (const [search, replacement] of replacements) {
+    const nextContent = content.replace(search, replacement);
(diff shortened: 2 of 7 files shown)


js/internal/environment_tests/test-exports-esm/package.json
```diff
@@ -5,13 +5,14 @@
   "type": "module",
   "description": "Tests that langsmith exports work correctly in ESM environments",
   "scripts": {
-    "test": "npx tsx src/index.ts"
+    "test": "tsc --noEmit && tsx src/index.ts"
```
sineha-mani, line 8: That is the test command for the existing ESM environment smoke test.

Before, it only ran the file:

"test": "npx tsx src/index.ts"
Now it first typechecks the test project, then runs it:

"test": "tsc --noEmit && tsx src/index.ts"
So it catches the exact regression: TypeScript users importing langsmith with skipLibCheck: false should not see broken package declaration errors.

dqbd (approved):

js/scripts/fix-openapi-client-declarations.js
```diff
@@ -0,0 +1,54 @@
+import { readFile, writeFile } from "node:fs/promises";
+import { dirname, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+
```
sineha-mani, line 4: We do it this way because js/src/_openapi_client/ is generated and protected, so we cannot edit those source files directly in this repo. This script patches only the built output before publishing.

PR #3047, opened 2026-06-16
feat: updating create feedback to add opt out for trace retention

sineha-mani:
Description

Adds SDK support for the new do_not_extend_trace_retention feedback creation option, so callers can create feedback without extending the associated trace’s retention tier. Python sync/async clients now accept do_not_extend_trace_retention=True, and the JS client accepts doNotExtendTraceRetention: true while sending the backend’s snake_case field. This preserves existing behavior by default and only opts out when explicitly requested.

References [LSE-2442](https://linear.app/langchain/issue/LSE-2442/add-api-control-for-feedback-trace-retention-sdk)

Python Sync
Screenshot 2026-06-16 at 3 26 31 PM

Python Async
Screenshot 2026-06-16 at 3 26 40 PM

JS
Screenshot 2026-06-16 at 3 26 48 PM

Test Plan
Added/updated unit coverage to verify Python feedback creation sends the opt-out field.
Added/updated unit coverage to verify JS feedback creation sends the opt-out field.
Ran targeted linting for the touched Python and JS files.
Tested with scripts for each SDK.

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -475,6 +475,7 @@ interface FeedbackCreate {
   session_id?: string;
   start_time?: number | string;
   comparative_experiment_id?: string;
+  do_not_extend_trace_retention?: boolean;
 }
 
 interface FeedbackUpdate {
@@ -4907,6 +4908,7 @@ export class Client implements LangSmithTracingClientInterface {
       comparativeExperimentId,
       sessionId,
       startTime,
+      doNotExtendTraceRetention,
     }: {
       score?: ScoreType;
       value?: ValueType;
@@ -4924,6 +4926,8 @@ export class Client implements LangSmithTracingClientInterface {
       sessionId?: string;
       /** The start time of the run this feedback is for. Accepts ISO string or epoch ms. */
       startTime?: number | string;
+      /** If true, create feedback without extending the trace's retention tier. */
+      doNotExtendTraceRetention?: boolean;
     },
   ): Promise<Feedback> {
     if (!runId && !projectId) {
@@ -4962,6 +4966,7 @@ export class Client implements LangSmithTracingClientInterface {
       feedbackConfig,
       session_id: sessionId ?? projectId,
       start_time: startTime,
+      do_not_extend_trace_retention: doNotExtendTraceRetention,
     };
     const body = JSON.stringify(feedback);
     const url = `${this.apiUrl}/feedback`;
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -475,7 +475,7 @@ interface FeedbackCreate {
   session_id?: string;
   start_time?: number | string;
   comparative_experiment_id?: string;
-  do_not_extend_trace_retention?: boolean;
+  extend_trace_retention?: boolean;
 }
 
 interface FeedbackUpdate {
@@ -4908,7 +4908,7 @@ export class Client implements LangSmithTracingClientInterface {
       comparativeExperimentId,
       sessionId,
       startTime,
-      doNotExtendTraceRetention,
+      extendTraceRetention,
     }: {
       score?: ScoreType;
       value?: ValueType;
@@ -4926,8 +4926,8 @@ export class Client implements LangSmithTracingClientInterface {
       sessionId?: string;
       /** The start time of the run this feedback is for. Accepts ISO string or epoch ms. */
       startTime?: number | string;
-      /** If true, create feedback without extending the trace's retention tier. */
-      doNotExtendTraceRetention?: boolean;
+      /** If false, create feedback without extending the trace's retention tier. */
+      extendTraceRetention?: boolean;
     },
   ): Promise<Feedback> {
     if (!runId && !projectId) {
@@ -4966,7 +4966,7 @@ export class Client implements LangSmithTracingClientInterface {
       feedbackConfig,
       session_id: sessionId ?? projectId,
       start_time: startTime,
-      do_not_extend_trace_retention: doNotExtendTraceRetention,
+      extend_trace_retention: extendTraceRetention,
     };
     const body = JSON.stringify(feedback);
     const url = `${this.apiUrl}/feedback`;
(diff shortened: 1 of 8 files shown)


python/langsmith/client.py
```diff
@@ -7717,6 +7721,7 @@ def bar(y):
                 feedback_group_id=_ensure_uuid(feedback_group_id, accept_null=True),
                 extra=extra,
                 error=error,
+                do_not_extend_trace_retention=do_not_extend_trace_retention,
```
sineha-mani: Question: is it okay that `do_not_extend_trace_retention=True` bypasses the queued/multipart feedback path and sends directly to `/feedback`?

Context: the direct `/feedback` API correctly honors `do_not_extend_trace_retention`, but the queued `/runs/multipart` feedback path does not preserve the opt-out in local testing. To keep the user-facing behavior correct, this sends opt-out feedback through the direct path while leaving the default/upgrade behavior on the existing queued path.

python/langsmith/_internal/_operations.py
```diff
@@ -170,6 +170,10 @@ def serialize_feedback_dict(
         feedback_create: dict = feedback.model_dump()  # type: ignore
     else:
         feedback_create = cast(dict, feedback)
+    if "do_not_extend_trace_retention" in feedback_create:
```
sineha-mani: Note:
{"do_not_extend_trace_retention": True}
becomes this before queued serialization:
{"_skip_trace_upgrade": True}

we wanted to have do_not_extend_trace_retention public facing and _skip_trace_upgrade internal

js/src/client.ts
```diff
@@ -4924,6 +4926,8 @@ export class Client implements LangSmithTracingClientInterface {
       sessionId?: string;
       /** The start time of the run this feedback is for. Accepts ISO string or epoch ms. */
       startTime?: number | string;
+      /** If true, create feedback without extending the trace's retention tier. */
+      doNotExtendTraceRetention?: boolean;
```
gowri100: ooc why did we change from extend trace retention to do not

does this require any defaults?

python/langsmith/client.py
gowri100: why can /runs/multipart not honor do_not_extend_trace_retention

js/src/client.ts
sineha-mani: We used doNotExtendTraceRetention because this is an opt-out flag. The default behavior is still to extend retention when feedback is created, and setting this to true skips that. Also, extend_evaluator_trace_retention is a separate property for evaluator based trace retention.

python/langsmith/client.py
sineha-mani: The problem is that the batched/multipart path can drop this opt-out flag before the backend applies retention logic, so feedback may still extend trace retention even when the user asked it not to.

To keep the user-facing behavior correct, I send opt-out feedback through the direct /feedback path instead. That path already understands do_not_extend_trace_retention and reliably skips the retention extension. The default case still uses the existing queued path.

**The flow can be** 
1. User calls client.createFeedback(...).
2. SDK sees batching/multipart is enabled.
3. SDK queues the feedback with pending trace/run operations.
4. That later gets flushed through the /runs/multipart ingestion path.
5. Backend processes it as part of run ingestion.

Our change says: if doNotExtendTraceRetention is true, don’t use that batching shortcut. Send it directly to /feedback, because that path definitely maps the opt-out to “do not auto-upgrade retention.

**Alternative**
We could also put this flag through the batched/multipart path so that path honors the opt-out too. I wasn’t sure that was necessary for this scope, since the direct /feedback path already gives correct behavior for the opt-out case while leaving the default batched path unchanged.

**Reasoning for not already doing the alternative**
- This SDK change is the smallest fix that guarantees correct user-facing behavior.
- Changing the backend batching path is broader: it would mean plumbing the opt-out flag through the multipart/run-ingest queue, making sure both Python and Go workers preserve it, and testing that it still behaves correctly for normal batched feedback.

sineha-mani: To clarify: With this branch, a createFeedback(..., doNotExtendTraceRetention: true) call will never upgrade the trace.

js/src/client.ts
gowri100: i wonder if we should call this extend_trace_retention for parity with the other retention names

python/langsmith/client.py
gowri100: the sdk is public - i wonder if the client should worry about the internal methods called by these endpoints (feedback vs multipart)


do you know what is the effort level required to add this feature to multipart?

gowri100: would using feedback instead of batched/multipart cause latency issues on the customer's end?

js/src/client.ts
sineha-mani: ooh i see how this could be good but i went with this because zhen mentioned that ideally, making default value false is better than true - but if this is a strong thought i'm happy to make the edits though, let me know!!

gowri100: @yumuzi234 can you provide more clarity as to why default false is better than true - and if we should deviate from the original extend_trace_retention naming?

yumuzi234: only because of Go, variables have zero default value, like boolean to false.
TS is not really the case, as no default.

sineha-mani: oh okay cool! - i'll rename but heads up i have to make a new backend PR on lcplus first to rename the actual API

python/langsmith/client.py
sineha-mani: Gotcha - it's not a big lift now that I have the serialize_feedback_dict translation that maps opt-out to _skip_trace_upgrade on the queued payload. I originally forced opt-out through direct /feedback because the batched path didn't carry the flag in a form the backend understood, so retention could still get extended even when the user opted out. 

Now that translation is in place and the backend already honors _skip_trace_upgrade on ingest, I removed the routing guard so opt-out can use the same fast path without changing what customers pass in.

So this has been resolved.

GowriH-1 (approved):

PR #3058, opened 2026-06-19
fix(py): declare OpenAPI client runtime dependencies

### LINKED ISSUE #3057: Python 0.8.17 generated OpenAPI client imports undeclared runtime dependencies
langsmith._openapi_client imports anyio, distro, sniffio, and typing_extensions, but these are missing from python/pyproject.toml project.dependencies / wheel Requires-Dist. In a distro build environment this causes importing langsmith._openapi_client to fail; pytest masked it as AttributeError during mock.patch("langsmith._openapi_client.Langsmith").

- _openapi_client/_base_client.py: anyio, distro, typing_extensions
- _openapi_client/_utils/_sync.py: anyio, sniffio, typing_extensions
- other non-generated modules also import typing_extensions.

I'm opening an issue instead of a direct PR because I'm not sure if they should be listed as hard dependencies or into an optional group, or maybe a new group?

gaoflow:
## Summary
- add the generated Python OpenAPI client's direct runtime imports to project dependencies
- update uv.lock so wheel metadata includes anyio, distro, sniffio, and typing-extensions
- leave generated _openapi_client files untouched

## Testing
- make format
- uv run --python 3.12 ruff check . && uv run --python 3.12 mypy langsmith
- make tests
- make build
- uv run --python 3.12 --no-dev python -c "import langsmith._openapi_client"

Fixes #3057

AI assistance was used under my direction.

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -24,11 +24,15 @@ keywords = [
     "platform",
 ]
 dependencies = [
+    "anyio>=3.5.0,<5",
+    "distro>=1.7.0,<2",
     "pydantic>=2,<3",
     "requests>=2.0.0",
     "orjson>=3.9.14; platform_python_implementation != 'PyPy'",
     "httpx>=0.23.0,<1",
     "requests-toolbelt>=1.0.0",
+    "sniffio>=1.3.0",
+    "typing-extensions>=4.11.0,<5",
     "zstandard>=0.23.0",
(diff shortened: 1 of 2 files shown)


jacoblee93:
Thanks for this! Looks like @KiewanVillatel merged this in https://github.com/langchain-ai/langsmith-sdk/pull/3070 so should be good to go

PR #3059, opened 2026-06-19
feat(anonymizer): curated secret-detection preset (createSecretAnonymizer)

paarth-a:
## Summary

Adds a curated, high-precision **secret-detection preset** to the anonymizer module of both the JS and Python SDKs, so applications (and our tracing integrations) can redact common credentials from run inputs/outputs/metadata **client-side, before they are uploaded**.

New public API (shipped via the existing `langsmith/anonymizer` subpath — no new entrypoint):

- `createSecretAnonymizer({ extraRules?, maxDepth? })` / `create_secret_anonymizer(*, extra_rules=None, max_depth=24)`
- `DEFAULT_SECRET_RULES` — the curated rule set
- `SECRET_PLACEHOLDER` — the replacement token (`[SECRET_DETECTED]`)

```ts
import { createSecretAnonymizer } from "langsmith/anonymizer";

const client = new Client({ anonymizer: createSecretAnonymizer() });
```
```python
from langsmith import Client
from langsmith.anonymizer import create_secret_anonymizer

client = Client(anonymizer=create_secret_anonymizer())
```

It reuses the existing `createAnonymizer` / `create_anonymizer` machinery, so redaction runs in the same place the SDK already masks data (before the HTTP POST) and automatically covers replica destinations (they reuse the run's client).

## Motivation

The tracing integrations currently send raw trace data (prompts, tool I/O, file contents, shell commands) with no redaction. A user pasting an API key into a prompt, an `export …_API_KEY=…` in a bash command, or a `.env` read all flow verbatim into traces. The anonymizer hook already exists; what was missing is a ready-made, low-false-positive rule set that any caller can opt into.

## Rule coverage (curated, high-precision)

- **Provider keys (prefix-anchored):** Anthropic, OpenAI (incl. project/service/admin + legacy), LangSmith (`lsv2_pt_…` / `ls__…`), GitHub (PAT + fine-grained), AWS access key id, Google API key + OAuth, Slack tokens + webhooks, Stripe, npm, SendGrid.
- **Structured tokens:** JWT; PEM private-key blocks (RSA/EC/OPENSSH/DSA/plain **and** PGP `…PRIVATE KEY BLOCK-----` armor).
- **Structural / contextual:** sensitive `NAME=value` / `"name": "value"` (component-boundary matched so `token` hits `api_token` but not `tokenizer`), `Authorization` / `x-api-key` / `x-auth-token` headers, bare `Bearer <token>`, and credentials embedded in URLs (`proto://user:PASS@host`).

This is **not** a port of gitleaks/secretlint — those informed the pattern shapes only. The set favors low false positives over exhaustive coverage; a generic high-entropy dragnet was intentionally excluded (it would redact legitimate hashes/UUIDs/code and degrade traces). `extraRules` lets callers add their own patterns.

## Design notes

- **JS ↔ Python parity:** the two rule sets are kept 1:1 (no lookbehind used). The `Authorization`-header rule preserves the `Bearer`/`Token`/`Basic` scheme in both.
- **Component boundary:** a trailing `(?![A-Za-z0-9])` after each keyword constrains only the trailing side, so concatenated names (`mysecret`, `dbpassword`) still match while `tokenizer`/`tokens` do not.
- **No silent partial redaction:** LangSmith v2 keys are multi-segment (`lsv2_pt_<key>_<tail>`); the rule consumes the full underscore-delimited tail so nothing leaks past the placeholder.

## Testing

- JS: `yarn jest src/tests/anonymizer.test.ts` — **51 passing** (each provider, JWT, PEM/PGP, multi-segment key, structural + header + URL + bare-Bearer cases, precision guards, deep-nesting, and an end-to-end "applies through the Client before upload" check).
- Python: `pytest tests/unit_tests/test_anonymizer.py` — **43 passing** (mirrors the JS suite).

## Downstream / rollout

The Claude Code, Codex, and Cursor tracing integrations will opt in via `Client({ anonymizer: createSecretAnonymizer() })` (on by default, with a `*_REDACT` opt-out). Those consumer changes are prepared separately and depend on this landing + publishing.

## Notes for reviewers

- Redaction is **opt-in at the SDK level** (this only adds the preset/helper; no default behavior change for existing SDK users).
- One intentional cosmetic asymmetry: `Authorization:` preserves the `Bearer` scheme word, while a `Bearer`-prefixed value under `X-Api-Key:` is collapsed into a single placeholder by the structural rule. No leak either way.


KiewanVillatel (approved):

jkennedyvz:
I compared `DEFAULT_SECRET_RULES` here against the curated gitleaks rule set our LLM gateway guard redacts server-side. The two intentionally diverge (this preset is broader on structural/contextual detection — `NAME=value`, auth headers, bare `Bearer`, URL creds — none of which the gateway has). But there are a few **provider-key signatures the gateway redacts that this preset would let through** before upload:

| Secret | Gateway rule | This PR | Notes |
|---|---|---|---|
| **GitLab PAT** (`glpat-…`) | `gitlab-pat` | ❌ none | Clean prefix, no FP risk. |
| **Slack app-level token** (`xapp-…`) | `slack-app-token` | ❌ — `xox[baprs]-` only matches `xox*`, not `xapp-` | gitleaks treats `xapp-` as its own rule. |
| **PyPI upload token** (`pypi-AgEI…`) | `pypi-upload-token` | ❌ none | Clean prefix. |
| **Azure AD client secret** (`…Q~…`) | `azure-ad-client-secret` | ⚠️ only via the structural `client_secret=…` rule; a bare value isn't matched | Lower priority (FP-prone shape). |
| **AWS key prefixes** | `aws-access-token` covers `AKIA/ASIA/A3T*/ABIA/ACCA` | ⚠️ only `AKIA\|ASIA` | Minor — the others are uncommon. |

The first three are the cheap wins — prefix-anchored, parity with rules the gateway already trusts, negligible false-positive risk. Suggested additions (keeping JS↔Python 1:1):

- `gitlab-pat`: `glpat-[A-Za-z0-9_-]{20,}`
- Slack app token: extend the class to `x(?:ox[baprs]|app)-…`, or add a dedicated `xapp-\d-[A-Za-z0-9-]+` rule
- `pypi-upload-token`: `pypi-AgEIcHlwaS[A-Za-z0-9_-]{50,}`

Not blocking — flagging so the client-side preset is a superset of what the gateway would otherwise catch in transit. (The reverse gaps — gateway's curated list is missing SendGrid / Google OAuth / JWT / `ls__` legacy LangSmith that this preset adds — I'm tracking separately on the gateway side.)

paarth-a:
Thanks @jkennedyvz — addressed in `bc04c478` (JS + Python, kept 1:1), all three cheap wins plus the AWS note:

- **GitLab PAT** — `glpat-[A-Za-z0-9_-]{20,}`
- **Slack app-level token** — dedicated `xapp-\d-[A-Za-z0-9-]{10,}` rule (the `xox[baprs]-` rule didn't reach `xapp-`)
- **PyPI upload token** — `pypi-AgEIcHlwaS[A-Za-z0-9_-]{50,}`
- **AWS** — broadened to `AKIA|ASIA|ABIA|ACCA|A3T[A-Z0-9]` while in there

Each has a fixture in the parametrized redaction test (incl. an `A3T` variant).

Deferred the **Azure AD client secret** for now: as you noted it's FP-prone as a bare value, and the structural `client_secret=…` rule already covers the assignment shape — I'd rather not add a broad bare-value matcher to a deliberately low-false-positive preset. If the gateway has a sufficiently prefix-/shape-anchored signature for it, happy to bring that over too.

PR #3061, opened 2026-06-19
test(python): fix integration assertions for updated attachment error message

QuentinBrosse:
The `ValueError` raised for filesystem-path attachments when `dangerously_allow_filesystem` is not set was reworded. The two assertions in `test_multipart_ingest_create_with_attachments_error` still matched the old wording and were failing on `main`.


emil-lc (commented):
Looks good

PR #3062, opened 2026-06-19
chore: reconcile bumpversion config and mandate release process for agents

QuentinBrosse:
## What

- **`python/.bumpversion.cfg`**: set `current_version` to `0.8.17` to match `langsmith/__init__.py`. It had drifted (was `0.8.12`) because recent releases hand-edited the version instead of running `bump2version`, which made `bump2version` fail with `VersionNotFoundException`.
- **`python/AGENTS.md` / `js/AGENTS.md`**: add a `Releasing` section requiring agents to follow the `CONTRIBUTING.md` release process (`bump2version` / `bump-version` + a version-bump PR) rather than hand-editing version files or improvising.

## Why

Keeps the documented `bump2version` release flow working and prevents ad-hoc release steps.

### DIFF
--- a/python/.bumpversion.cfg
+++ b/python/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.8.12
+current_version = 0.8.17
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)
 serialize = {major}.{minor}.{patch}
 search = {current_version}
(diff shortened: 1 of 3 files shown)


python/.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.8.12
```
QuentinBrosse, line 2: match https://github.com/langchain-ai/langsmith-sdk/blob/411401f6ca5ff401c29772ed097c9c9ed97f994c/python/langsmith/__init__.py#L32

emil-lc (approved):

PR #3065, opened 2026-06-20
feat(client): make secret redaction the default in both SDKs

jkennedyvz:
## Summary

Makes secret redaction **on by default** in both the Python and JS `Client`, so users have to **opt in to sending us secrets** rather than opt in to redaction.

Builds on the curated secret-detection preset from #3059 (`create_secret_anonymizer` / `createSecretAnonymizer`). The Client constructor now auto-applies it to run **inputs, outputs, and metadata** before upload — no caller action required.

## Changes

**Both SDKs:**
- New constructor parameter: `redact_secrets` (Python) / `redactSecrets` (JS) — defaults to `None` / `undefined`, meaning "on."
- New env var: `LANGSMITH_REDACT_SECRETS` — set to `"false"` to disable.
- When enabled and no custom `anonymizer` is provided, the Client auto-configures `create_secret_anonymizer()` / `createSecretAnonymizer()`.
- **Metadata redaction:** the anonymizer is now also applied to run metadata (`_hide_run_metadata` / `processMetadata`), not just inputs/outputs.

**Precedence rules:**
1. Custom `anonymizer` → takes priority over default secret redaction.
2. Explicit `hide_inputs`/`hide_outputs`/`hide_metadata` → default secret anonymizer is **not** applied, so custom filters are respected.
3. `redact_secrets=False` / `LANGSMITH_REDACT_SECRETS=false` → disables default redaction entirely.

## Usage

```python
# Python — redaction is on by default
from langsmith import Client
client = Client()  # secrets redacted automatically

# Opt out (send raw values)
client = Client(redact_secrets=False)
# or: export LANGSMITH_REDACT_SECRETS=false
```

```ts
// JS — redaction is on by default
const client = new Client();  // secrets redacted automatically

// Opt out (send raw values)
const client = new Client({ redactSecrets: false });
// or: export LANGSMITH_REDACT_SECRETS=false
```

## Testing

- **Python:** `pytest tests/unit_tests/test_anonymizer.py` — 53 passing (includes 7 new default-redaction tests). `pytest tests/unit_tests/test_client.py tests/unit_tests/test_anonymizer.py` — 276 passing.
- **JS:** `pnpm test:single src/tests/anonymizer.test.ts` — 61 passing (includes 6 new default-redaction tests).
- Lint/format clean in both SDKs (`ruff`, `mypy`, `oxlint`, `oxfmt`).

## Notes for reviewers

- This is a **behavioral default change** for existing SDK users: secrets will now be redacted from traces unless they explicitly opt out. This is the intended security posture — users should opt in to sending us secrets, not opt in to redaction.
- The redaction is high-precision (provider key prefixes, JWT, PEM blocks, structural `NAME=value` / `Authorization` / URL-credential rules) and favors low false positives. See #3059 for the full rule set.
- Existing users who pass a custom `anonymizer` or explicit `hide_inputs`/`hide_outputs`/`hide_metadata` are unaffected.

image

### DIFF
--- a/js/src/anonymizer/index.ts
+++ b/js/src/anonymizer/index.ts
@@ -96,12 +96,9 @@ export type ReplacerType =
 // pass over the string instead of N, reducing per-string overhead from N
 // replace() calls to 1.
 //
-// For the secret anonymizer, we go further: instead of parsing the serialized
-// JSON, walking the object tree to extract string nodes, running regex per
-// node, and walking the tree again to apply updates, we run the combined
-// regex directly on the serialized JSON string and parse once at the end.
-// This eliminates extractStringNodes, applyUpdateAtPath, and the intermediate
-// JSON.parse entirely.
+// The secret anonymizer applies this optimization inside the node-based
+// pipeline: each string node is prefiltered (skipping large base64 blobs and
+// strings with no secret indicators), then matched against the combined regex.
 
 type SimpleRule = { source: string; flags: string; replace: string };
 type StructuralRule = { regex: RegExp; replace: string };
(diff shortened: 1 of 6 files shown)


jacoblee93:
This is breaking right?

paarth-a (approved):

paarth-a:
wait maybe this requires more discussion as per slack

jacoblee93 (changes requested):
Yes @paarth-a we need to be careful rolling this out (if we do)

My preference would generally be to do this at the integration level even if we do decide to make breaking changes, keeping the raw client calls relatively pure

PR #3077, opened 2026-06-23
feat: add SmithDB by-key path to add_runs_to_annotation_queue (runs= param) [LSDK-287]

ayoung19:
## Description

Wires the SmithDB-native `POST /annotation-queues/{queue_id}/runs/by-key` endpoint into the hand-written Python + JS SDKs.

Per the settled AQ→SmithDB decision (LSDK-287; Quentin's sync + Jacob Lee's PR comment), this does **not** add a separate `_by_key` method. Instead it **overloads the existing** `add_runs_to_annotation_queue` with a `runs=` param — one method, two shapes:

```python
# preferred (SmithDB): carries the full lookup key, no scan
client.add_runs_to_annotation_queue(queue_id, runs=[
    {"run_id": r.id, "session_id": r.session_id, "start_time": r.start_time}
])

# legacy (unchanged): run IDs only -> v1 /runs
client.add_runs_to_annotation_queue(queue_id, run_ids=[r.id])
```

Exactly one of `runs` / `run_ids` must be provided (else a `LangSmithUserError`). JS mirrors this: `addRunsToAnnotationQueue(queueId, string[] | { runIds?, runs? })`.

**Non-breaking:** existing `run_ids=` callers (and JS bare `string[]`) are unaffected. `run_id`/`run_ids` is **not** deprecation-warned yet — that's a later, comms-synced step (tracked separately), so this PR only adds the `runs=` capability.

Go/Java (Stainless) are intentionally **out of scope** here — that generation is being punted (see langchainplus #28211, held).

### Changes
- **Python (sync + async):** `add_runs_to_annotation_queue(run_ids=… | runs=…)`; `RunToAddByKey` TypedDict; shared `_serialize_run_to_add_by_key` helper (serializes UUIDs + `start_time`).
- **JS:** unified `addRunsToAnnotationQueue` accepting `string[]` or an options object.
- Integration tests for the `runs=` path in both SDKs.

## Test Plan

- [x] Python: `ruff check` + `ruff format` clean; signature + mutual-exclusivity smoke test passes
- [x] JS: `tsc --noEmit` clean, `oxlint` 0 errors, `oxfmt --check` clean
- [x] Integration tests updated (`test_annotation_queue_runs_by_key`, `annotationqueue add runs by key`) — run against a live LangSmith API


jacoblee93:
Hmm instead of doing this can we gate on backend version? I don't want to bloat the client surface area if possible

js/src/client.ts
```diff
@@ -5715,33 +5716,77 @@ export class Client implements LangSmithTracingClientInterface {
...
+   * An empty array is treated as the (legacy) run-IDs form and is a no-op.
+   *
    * @param queueId - The ID of the annotation queue
-   * @param runIds - The IDs of the runs to be added to the annotation queue
+   * @param runs - Either a list of run IDs (deprecated) or a list of run keys.
    */
   public async addRunsToAnnotationQueue(
     queueId: string,
-    runIds: string[],
+    runs: string[] | RunToAddByKey[],
```
jacoblee93: nit: naming could be better, how about just `AddRunsToAnnotationQueueParams`

js/src/client.ts
```diff
@@ -5715,33 +5716,77 @@ export class Client implements LangSmithTracingClientInterface {
...
+              run.sourceProposedExampleId,
+              `runs[${i}].sourceProposedExampleId`,
+            ).toString();
+          }
+          return serialized;
+        }),
+      );
+    } else {
+      url = base;
+      body = JSON.stringify(
```
jacoblee93, line 5772: Do we need to log a warning?

jacoblee93 (approved):

js/src/client.ts
```diff
@@ -5715,33 +5716,77 @@ export class Client implements LangSmithTracingClientInterface {
...
-    const body = JSON.stringify(
-      runIds.map((id, i) => assertUuid(id, `runIds[${i}]`).toString()),
-    );
+    const base = `${this.apiUrl}/annotation-queues/${assertUuid(
+      queueId,
+      "queueId",
+    )}/runs`;
+
+    // Discriminate by element shape: objects → by-key, strings (or empty) → v1.
+    const byKey = runs.length > 0 && typeof runs[0] !== "string";
```
QuentinBrosse: nit: Should we test the list itself?

js/src/client.ts
QuentinBrosse: I think we should wait until the migration guide is ready before starting the deprecation. (in 1 or 2 weeks).

So we can include the migration guide in the deprecation warning

js/src/client.ts
ayoung19: renaming it to `RunKey`

PR #3081, opened 2026-06-23
perf(anonymizer): skip base64 blobs in createSecretAnonymizer

paarth-a:
## Summary

Skips base64 blobs in `createSecretAnonymizer` so the preset doesn't scan binary payloads. Before the rules run, a string is skipped if it's a `data:` URI or **wholly base64** (standard alphabet). Scanning a multi-MB base64 value (e.g. a screenshot in a tool result) with every rule is wasted CPU — ~50 ms (JS) / ~350 ms (Python) to find zero secrets — and risks matching random base64 substrings and corrupting the blob.

## Scope / deliberate tradeoff

- The base64 check uses the **standard alphabet (`A-Za-z0-9+/`) only — not base64url**, so `-`/`_`-bearing provider keys (`sk-ant-`, `lsv2_`, `ghp_`, `glpat-`) and dotted tokens (JWTs) are **not** wholly base64 and still get scanned.
- Consequence: a **standalone** pure-base64 value (e.g. a bare `AKIA…` AWS key) is no longer redacted, but the **same key in context** (`AWS_ACCESS_KEY_ID=AKIA…`) still is — that string isn't wholly base64.

## Changes

- `js/src/anonymizer/index.ts`, `python/langsmith/anonymizer.py` — `createSecretAnonymizer` / `create_secret_anonymizer` short-circuit `data:` URIs and wholly-base64 strings before applying rules.
- Tests for both: base64 blob / bare-key skipped, `data:` skipped, in-context key still redacted, non-base64 strings still scanned.

## Testing

- JS: `yarn jest src/tests/anonymizer.test.ts` — 59 pass.
- Python: `pytest tests/unit_tests/test_anonymizer.py` — 48 pass. (ruff + mypy clean.)

## Note

This rides on top of the already-released preset, so it needs a version bump to publish (JS `0.7.x`, Python `0.9.x`) — handled separately at release time.

### DIFF
--- a/js/src/anonymizer/index.ts
+++ b/js/src/anonymizer/index.ts
@@ -308,6 +308,14 @@ export const DEFAULT_SECRET_RULES: StringNodeRule[] = [
  *   `createAnonymizer`'s default of 10 because traced payloads nest deeply,
  *   e.g. `messages[].content[].args`).
  *
+ * Strings that are wholly base64 (image/PDF/binary blobs) or `data:` URIs are
+ * skipped before the rules run — scanning a multi-MB blob with every rule is
+ * wasted CPU and risks matching random base64 substrings (corrupting it). The
+ * base64 check uses the STANDARD alphabet (A-Za-z0-9+/) only, NOT base64url, so
+ * `-`/`_`-bearing provider keys (sk-ant-, lsv2_, ghp_, glpat-) and dotted JWTs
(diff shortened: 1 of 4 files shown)


jacoblee93:
Don't think this is a good idea - the user can implement this heuristic themselves if they want and we don't want to add arbitrary undocumented logic to skip something this important

PR #3094, opened 2026-06-26
fix(google_adk): capture LLM inputs after before_model_callback runs [LSDK-279]

harisaiharish:
## Summary

Fixes incomplete LLM traces in the Google ADK Python integration when a `before_model_callback` (ADK plugin or agent callback) modifies the request.

The wrapper captured the LLM inputs and called `post()` **before** invoking ADK's `_call_llm_async`. But ADK runs `before_model_callback` *inside* `_call_llm_async`, and those callbacks mutate `llm_request` in place (e.g. `llm_request.append_instructions(...)`). As a result, any dynamic changes - injected system instructions, guardrails, skills blocks - were sent to the model but **never reflected in the trace**, so replaying the LLM run in the Playground produced a different response.

The fix defers input capture and `post()` until the wrapped coroutine has yielded, by which point the before-model callbacks have run and `llm_request` reflects what was actually sent.

## Changes

- `langsmith/integrations/google_adk/_client.py` - `wrap_flow_call_llm_async`:
  - Create the LLM child run up front (to anchor the real start time) but **defer** capturing `messages`/`tools`/`model_name` and calling `post()` until after the wrapped coroutine yields its first event (after `before_model_callback` has mutated `llm_request`).
  - Capture + post is idempotent and also invoked in the no-event and exception paths, so the run is always recorded (even if a callback raises before the model is called).
- `tests/unit_tests/wrappers/test_google_adk.py` - new regression test (sync + async) using a fake `BaseLlm` model and a `before_model_callback` that injects an instruction; asserts the injected text appears in the traced LLM run inputs.

## Self-hosted release note

Fixed a bug in the Google ADK Python integration where LLM run inputs did not reflect modifications made by `before_model_callback` (ADK plugins/agent callbacks), such as dynamically injected system instructions. Traces now capture the final request sent to the model.

## Testing

- **Python:** `pytest tests/unit_tests/wrappers/test_google_adk.py` - 8 passing (includes 2 new sync/async regression tests).
- Verified the new test fails without the fix (system message missing the injected instruction) and passes with it.
- Lint/format clean (`ruff`).

### DIFF
--- a/python/langsmith/integrations/google_adk/_client.py
+++ b/python/langsmith/integrations/google_adk/_client.py
@@ -349,36 +349,45 @@ async def wrap_flow_call_llm_async(
         return
 
     llm_request = args[1] if len(args) > 1 else kwargs.get("llm_request")
-    model_name = extract_model_name(llm_request) if llm_request else None
-    messages = convert_llm_request_to_messages(llm_request) if llm_request else None
-    tools = extract_tools_from_llm_request(llm_request) if llm_request else []
-
-    inputs: dict[str, Any] = {}
-    if messages:
-        inputs["messages"] = messages
-
-    metadata: dict[str, Any] = {"ls_provider": _get_ls_provider()}
-    if model_name:
-        metadata["ls_model_name"] = model_name
-
-    # Build extra dict with invocation_params if tools exist
-    extra: dict[str, Any] = {"metadata": metadata}
-    if tools:
-        extra["invocation_params"] = {"tools": tools}
 
     start_time = time.time()
+    # Anchor start time now, but capture inputs later:
+    # ADK's before_model_callback mutates llm_request in-place in wrapped().
     llm_run = parent.create_child(
-        name=model_name or "google_adk_llm",
+        name="google_adk_llm",
         run_type="llm",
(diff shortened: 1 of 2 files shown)


langchain-infra (approved):

python/langsmith/integrations/google_adk/_client.py
```diff
@@ -349,36 +349,43 @@ async def wrap_flow_call_llm_async(
...
+        if messages:
+            llm_run.inputs = {"messages": messages}
+        if model_name:
+            llm_run.name = model_name
+            llm_run.extra.setdefault("metadata", {})["ls_model_name"] = model_name
+        if tools:
+            llm_run.extra.setdefault("invocation_params", {})["tools"] = tools
+
+        try:
+            llm_run.post()
```
QuentinBrosse, line 386: nit: should we set `posted=True` here instead?

QuentinBrosse (approved):

python/langsmith/integrations/google_adk/_client.py
harisaiharish: good call

PR #3097, opened 2026-06-26
fix: lazily initialize sync OpenAPI client

### LINKED ISSUE #3096: Client initialization can fail when NO_PROXY contains IPv6 CIDR entries
### Summary

`langsmith==0.9.2` can fail during basic sync `Client(...)` initialization when the process environment contains `NO_PROXY` / `no_proxy` entries such as `::/0`.

The failure happens before any LangSmith API request is made.

### Environment

Observed with:

- `langsmith==0.9.2`
- `httpx==0.28.1`
- Python 3.13
- `NO_PROXY` / `no_proxy` containing an IPv6 CIDR-style entry such as `::/0`

### Reproduction

```python
import os
from langsmith import Client

os.environ["NO_PROXY"] = "::/0"
os.environ["no_proxy"] = "::/0"

Client(
    api_url="https://api.smith.langchain.com",
    api_key="test-api-key",
    auto_batch_tracing=False,
)
```

This can raise:

```text
httpx.InvalidURL: Invalid port: ':'
```

### Expected behavior

Constructing the sync `Client` should not fail because of proxy parsing in an async OpenAPI/httpx client that is not yet being used.

### Notes

This appears related to the generated OpenAPI client being initialized eagerly during sync `Client.__init__`. In earlier versions, generated OpenAPI resources were effectively initialized only when those resources were accessed.

sunfayao:
## Summary

This changes the sync `Client` to lazily initialize the generated OpenAPI client.

The recent generated v2 resource wiring exposes resources such as `client.runs` and `client.online_evaluators`, backed by the generated OpenAPI client. That behavior is preserved, but the generated client is now constructed only when those resources are first accessed.

This avoids creating an `httpx.AsyncClient` during basic sync `Client(...)` construction. In environments where `NO_PROXY` / `no_proxy` contains IPv6 CIDR-style entries such as `::/0`, eager construction can fail before any LangSmith request is made with:

```text
httpx.InvalidURL: Invalid port: ':'
```

## Change

- Set `self._langsmith_api` to `None` during sync `Client.__init__`
- Add `_get_langsmith_api()` to construct and cache the generated OpenAPI client on first use
- Keep `client.runs` and `client.online_evaluators` behavior unchanged from the caller perspective
- Add a regression test for sync `Client(...)` initialization with `NO_PROXY=::/0`

## Testing

- `python3 -m py_compile python/langsmith/client.py python/tests/unit_tests/test_client.py`
- Manual reproduction check with `NO_PROXY='::/0' no_proxy='::/0'` verifies `Client(...)` initialization succeeds after this change

I could not complete the full `make tests` run in my local environment because dependency setup failed while resolving the build-system dependency `hatchling`.

Fixes #3096

### DIFF
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -889,6 +889,7 @@ class Client:
     _manual_cleanup: bool
     _profile_auth: Optional[_profiles.ProfileAuth]
     _profile_auth_headers: dict[str, str]
+    _langsmith_api: Optional[LangsmithOpenAPIClient]
 
     def __init__(
         self,
@@ -1427,18 +1428,7 @@ class Client:
             )
             self._failed_traces_max_bytes = 100 * 1024 * 1024
 
-        self._langsmith_api = LangsmithOpenAPIClient(
-            api_key=self._api_key,
-            tenant_id=str(self._workspace_id) if self._workspace_id else None,
-            base_url=self.api_url,
-            timeout=_httpx.Timeout(
-                connect=self._timeout[0],
-                read=self._timeout[1],
-                write=self._timeout[1],
-                pool=self._timeout[0],
-            ),
-            default_headers=self._headers or None,
-        )
+        self._langsmith_api = None
 
     # ------------------------------------------------------------------
     # Stainless v2 resource accessors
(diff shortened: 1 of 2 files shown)


KiewanVillatel:
Fixed in https://github.com/langchain-ai/langsmith-sdk/pull/3099.

PR #3100, opened 2026-06-26
fix(client): apply anonymizer to run error field

paarth-a:
## Problem

The client-side `anonymizer` (including `create_secret_anonymizer`) only covers run **inputs** and **outputs** — not the `error` field. A secret captured in an exception traceback is therefore uploaded **unredacted**.

From a customer report (SDK 0.8.16): an `aiohttp` call failed, the exception's request-object repr carried the full `Authorization` header, and it landed verbatim in the run's `error` field. The error string is built in `run_helpers._format_error_with_exceptions_to_handle` as `repr(exc)` + traceback and was assigned straight into the payload with no redaction pass.

There is also no `hide_error` toggle and no `process_error` hook, so on released versions there's no client-side way to scrub it (only `@traceable(exceptions_to_handle=...)`, which drops the error entirely).

## Fix

Route `error` through the same anonymizer as inputs/outputs via a new `Client._hide_run_error`, wired at the two ingestion chokepoints:

- **`_run_transform`** — covers `create_run`, `batch_ingest_runs`, and `multipart_ingest_runs` (create *and* update funnel through here)
- **`update_run`** — the one standalone patch path that hides inline

The value is wrapped as `{"error": ...}` before calling the anonymizer so that both `create_secret_anonymizer` and a user-supplied anonymizer matching the documented `Callable[[dict], dict]` signature receive a dict — a bare string would break a dict-typed anonymizer. No JSON round-trip (unlike inputs/outputs), since `error` is already a string.

Scoped intentionally minimal — **no** new `hide_error`/`HIDE_ERROR` toggle (happy to add if wanted; left out to avoid a public-API/naming decision in a fix PR).

## Tests

`test_anonymizer_redacts_error_field` verifies redaction on the create path (over-the-wire POST payload), the patch path (`_hide_run_error`), and pass-through when no anonymizer is configured. `ruff check`/`format` and the anonymizer + client suites pass.

## Performance

Negligible, including for the tracing plugins that ship `create_secret_anonymizer`: zero added cost on success spans (guarded on `error is not None`), and on error spans it's one regex pass over a single string — cheaper per-call than the inputs/outputs scrub, which does a full orjson serialize→parse round-trip.

## Follow-up (not in this PR)

`create_secret_anonymizer`'s docstring claims it redacts "inputs, outputs, **and metadata**", but `_hide_run_metadata` never consults `self._anonymizer` (metadata is gated only by the `hide_metadata` bool/callable). Either wire metadata through the anonymizer or fix the docstring — separate change.

### DIFF
--- a/python/tests/unit_tests/test_client.py
+++ b/python/tests/unit_tests/test_client.py
@@ -1493,6 +1493,72 @@ def test_hide_metadata(
         )
 
 
+def test_anonymizer_redacts_error_field() -> None:
+    """The configured anonymizer must scrub ``error``, not just inputs/outputs.
+
+    A credential can land in the error field via an exception traceback (e.g. an
+    HTTP-client error whose request-object repr includes an ``Authorization``
+    header), and it must not be uploaded raw. Covers both the create/post path
+    (via ``_run_transform``) and the patch path (via ``_hide_run_error``).
+    """
+    from langsmith.anonymizer import SECRET_PLACEHOLDER, create_secret_anonymizer
+
+    # Assembled at runtime so no literal secret-shaped string sits in source.
+    fake_key = "sk-ant-api03-" + "A" * 30
+    traceback_str = (
+        "Traceback (most recent call last):\n"
+        '  ClientResponseError: 401, request_info=headers={"Authorization": '
+        f'"Bearer {fake_key}"}}'
+    )
+
+    session = mock.MagicMock(spec=requests.Session)
+    client = Client(
+        api_url="http://localhost:1984",
+        api_key="123",
+        auto_batch_tracing=False,  # Easier to inspect single calls
+        session=session,
+        anonymizer=create_secret_anonymizer(),
+    )
+
+    # ── create/post path goes through _run_transform ────────────────────────
+    client.create_run(
+        "errored_run",
+        inputs={"in": "put"},
+        run_type="llm",
+        id=uuid.uuid4(),
+        error=traceback_str,
+    )
+    post_call = None
+    for call in session.request.mock_calls:
+        if len(call.args) > 1 and call.args[0] == "POST" and "runs" in call.args[1]:
+            post_call = call
+            break
+    assert post_call is not None, "POST request to /runs not found"
+    payload_data = post_call.kwargs.get("data", b"{}")
+    payload = json.loads(
+        payload_data.decode("utf-8")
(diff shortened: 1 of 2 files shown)


python/tests/unit_tests/test_client.py
```diff
@@ -1493,6 +1493,72 @@ def test_hide_metadata(
...
 
+def test_anonymizer_redacts_error_field() -> None:
+    """The configured anonymizer must scrub ``error``, not just inputs/outputs.
+
+    A credential can land in the error field via an exception traceback (e.g. an
+    HTTP-client error whose request-object repr includes an ``Authorization``
+    header), and it must not be uploaded raw. Covers both the create/post path
+    (via ``_run_transform``) and the patch path (via ``_hide_run_error``).
+    """
+    from langsmith.anonymizer import SECRET_PLACEHOLDER, create_secret_anonymizer
```
QuentinBrosse: Can we import at the top of the file?

python/tests/unit_tests/test_client.py
```diff
@@ -1493,6 +1493,72 @@ def test_hide_metadata(
...
+        if isinstance(payload_data, bytes)
+        else payload_data
+    )
+    assert fake_key not in payload["error"]
+    assert SECRET_PLACEHOLDER in payload["error"]
+
+    # ── patch path uses _hide_run_error directly ────────────────────────────
+    scrubbed = client._hide_run_error(traceback_str)
+    assert fake_key not in scrubbed
+    assert SECRET_PLACEHOLDER in scrubbed
```
QuentinBrosse: Should we just create another test that does a patch?

python/tests/unit_tests/test_client.py
paarth-a: addressed in [fd44fc7](https://github.com/langchain-ai/langsmith-sdk/pull/3100/commits/fd44fc78a9a288b5a0d2e2d88f28401fffcd137a)

python/tests/unit_tests/test_client.py
paarth-a: addressed in [fd44fc7](https://github.com/langchain-ai/langsmith-sdk/pull/3100/commits/fd44fc78a9a288b5a0d2e2d88f28401fffcd137a)

QuentinBrosse (approved):

PR #3110, opened 2026-06-29
fix(client): apply anonymizer to run error field (JS)

paarth-a:
## Problem

JS parity for #3100. The client-side `anonymizer` (including `createSecretAnonymizer`) is aliased only to `hideInputs`/`hideOutputs` at construction (`client.ts`), so it never touches the `error` field. A secret captured in an exception message/traceback is therefore uploaded **unredacted**.

The two prep chokepoints both skipped `error`:
- `prepareRunCreateOrUpdateInputs` — processed inputs/outputs/metadata/events (create + batch + multipart paths)
- `updateRun` — processed inputs/outputs/metadata inline (the standalone patch path)

Errors land here raw — e.g. `runTree.end(undefined, String(error))` in `traceable`, or a user-supplied `run.error` / `err.stack`. (Same metadata nuance as Python: the anonymizer isn't aliased to `hideMetadata` either — out of scope here.)

## Fix

Retain the anonymizer reference (`this.anonymizer = config.anonymizer`) and add a `processError` mirroring `processInputs`/`processOutputs`. Since the anonymizer is typed `(KVMap) => KVMap`, the error string is wrapped as `{ error }`, scrubbed, and unwrapped — the same trick as the Python SDK's `Client._hide_run_error` (`{"error": …}`). Applied at both chokepoints:

- `prepareRunCreateOrUpdateInputs` → covers `createRun` + batch + multipart (create *and* update)
- `updateRun` → the inline patch path

No new public `hideError` toggle (matching #3100's minimal scope) — this only makes the existing anonymizer cover `error`.

## Tests

Three tests in `anonymizer.test.ts` (`describe("error field")`), using `mockClient({ anonymizer: createSecretAnonymizer() })` + `getAssumedTreeFromCalls`:
- redaction on the create path (POST)
- redaction on the update path (PATCH via `updateRun`)
- pass-through when no anonymizer is configured

`pnpm test:single` → 58 passed (3 new), `tsc --noEmit` clean, `oxlint` 0 errors.

Companion to the Python fix in #3100.

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -881,6 +881,8 @@ export class Client implements LangSmithTracingClientInterface {
     | boolean
     | ((metadata: KVMap) => KVMap | Promise<KVMap>);
 
+  private anonymizer?: (values: KVMap) => KVMap | Promise<KVMap>;
+
   private omitTracedRuntimeInfo?: boolean;
 
   private tracingSampleRate?: number;
@@ -1266,6 +1268,7 @@ export class Client implements LangSmithTracingClientInterface {
     this.hideOutputs =
       config.hideOutputs ?? config.anonymizer ?? defaultConfig.hideOutputs;
     this.hideMetadata = config.hideMetadata ?? defaultConfig.hideMetadata;
+    this.anonymizer = config.anonymizer;
 
     this.omitTracedRuntimeInfo = config.omitTracedRuntimeInfo ?? false;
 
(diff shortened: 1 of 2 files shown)


js/src/client.ts
```diff
@@ -1508,6 +1511,24 @@ export class Client implements LangSmithTracingClientInterface {
     return metadata;
   }
 
+  /**
+   * Apply the configured anonymizer to a run's error string.
+   *
+   * Unlike inputs/outputs, `error` is a plain string (an exception message or
```
jacoblee93, line 1525: We really need to fix our typing...

jacoblee93 (approved):

js/src/client.ts
jacoblee93: Added a TODO since it will be breaking

jacoblee93:
Thank you!

PR #3111, opened 2026-06-29
fix(google-adk): set tool span as active tracing context in wrap_tool_run_async

navarra-lisandro:
## Summary
Hi SDK team, this is a quick fix spotted from a support case, did a local test and it looked straightforward, but happy to defer to the team's judgment or close if there's a better approach already in flight. 

- `get_current_run_tree()` inside a tool body returned the **parent span** instead of the **tool span** because `set_tracing_parent` was not called after creating the child `tool_run`.
- Fix: wrap `await wrapped(*args, **kwargs)` with `with set_tracing_parent(tool_run):` so the `ContextVar` holds the tool span for the duration of the call (including nested LLM calls, callbacks, and exception paths).
- `set_tracing_parent` was already imported at line 14; this is a one-line change.
- Added a regression test: `test_wrap_tool_run_async_sets_tool_as_active_context` — verifies that `get_current_run_tree()` inside a tool body resolves to the tool span, not the parent.

> Quick fix spotted from a support case — did a local test and it looked straightforward, but happy to defer to the team's judgment or close if there's a better approach already in flight.

## Test plan

- [ ] New test `test_wrap_tool_run_async_sets_tool_as_active_context` passes with the fix and fails without it (verified locally)
- [ ] Full `make tests` suite run locally — all existing tests pass
- [ ] Note: 4 pre-existing sandbox test failures exist on `main` (`test_create_sandbox_merges_custom_headers`, `test_run_with_custom_headers`) due to `LANGSMITH_API_KEY` leaking from the environment — unrelated to this change, confirmed present on `main` before this branch

Closes [#25736 ](https://app.usepylon.com/support/issues/views/my-issues?issueNumber=25736&view=fs)(reported by Gong via support case)

### DIFF
--- a/python/langsmith/integrations/google_adk/_client.py
+++ b/python/langsmith/integrations/google_adk/_client.py
@@ -301,7 +301,8 @@ async def wrap_tool_run_async(
         logger.debug(f"Failed to post tool run: {e}")
 
     try:
-        result = await wrapped(*args, **kwargs)
+        with set_tracing_parent(tool_run):
+            result = await wrapped(*args, **kwargs)
         if isinstance(result, dict):
             outputs = result
         elif isinstance(result, list):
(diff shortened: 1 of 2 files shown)


jacoblee93 (approved):
Nice, thank you!

PR #3107, opened 2026-06-26
feat(integrations): voice tracing integrations (Pipecat, LiveKit, OpenAI Realtime, ADK Live)

carolinedivittorio:
## Status: in development

These voice integrations are in development. Each voice package emits a `LangSmithVoiceBetaWarning` on import, and they are isolated from the core SDK: `import langsmith` loads none of them (verified) — they load only when you explicitly `from langsmith.integrations.<voice> import ...`, which requires the corresponding extra's framework dependency.

## Summary

Adds `langsmith.integrations` packages for tracing voice agents to LangSmith. Two mechanisms, matching how each provider exposes its data:

- **OTel span processors** (frameworks that emit OpenTelemetry):
  - `pipecat` — `PipecatLangSmithSpanProcessor` + `configure_pipecat` (extra: `pipecat`)
  - `livekit` — `LiveKitLangSmithSpanProcessor` + `configure_livekit` (extra: `livekit`)
- **RunTree event-stream tracers** (providers whose streaming paths emit no usable telemetry):
  - `openai_realtime` — both `wrap_realtime` (raw `client.realtime.connect()` loop) and `wrap_realtime_session` (`agents.realtime`), behind the **`openai-realtime`** extra
  - `google_adk_live` — `LangSmithLivePlugin` for `Runner.run_live`, behind the **`google-adk-live`** extra

Shared, private machinery lives in `integrations/_voice/`: an OTel `BaseLangSmithSpanProcessor` (dedupes the Pipecat/LiveKit processors) and a `RunTree` `EventSession` engine (root span, turn grouping, transcript rollup, stereo-WAV audio attachment).

## Design choices

- **One extra per provider concept.** `langsmith[openai-realtime]` provides *both* OpenAI Realtime wrappers from one package. ADK Live lives in its own `google_adk_live` package behind `langsmith[google-adk-live]`, **kept separate from the existing `google_adk` batch integration** so non-streaming ADK users take on no new dependencies. The existing `openai_agents_sdk` and `google_adk` packages are unchanged (batch-only).
- **No numpy dependency.** The stereo conversation WAV is built with the stdlib `array` module, so no extra pulls numpy.
- `google_adk_live` exposes its plugin lazily (`__getattr__`) so importing the package doesn't require `google-adk`.
- The OTel processors mutate `ReadableSpan._attributes` on `on_end` (the standard rewrite-on-export pattern from the reference processors); the RunTree engine builds spans explicitly.

## Validation

[...]

PR #3112, opened 2026-06-29
feat(integrations): add Pipecat voice tracing integration

carolinedivittorio:
## Pipecat voice tracing (plus shared voice foundation)

First of a 4-PR stack adding LangSmith tracing integrations for voice agents. This PR adds:

- **Shared foundation** under `langsmith.integrations._voice` (private): an in-development warning helper; the OTel `BaseLangSmithSpanProcessor` (downstream wrapping, LangSmith OTLP exporter default, thread-id injection, message + audio-attachment helpers); per-conversation thread grouping via `set_thread_id` (a `ContextVar`); and stdlib-only `pcm_to_wav` for encoding the recording (no numpy).
- **Pipecat integration** (`langsmith.integrations.pipecat`): `configure_pipecat` + `PipecatLangSmithSpanProcessor`, translating Pipecat's OTel spans (`conversation` / `turn` / `stt` / `llm` / `tts`) into LangSmith's `gen_ai.*` / `langsmith.*` shape and attaching the `AudioBufferProcessor` recording to the conversation root. Per-conversation state is TTL-bounded (via `cachetools`) so an abandoned call cannot leak it.

Install: `pip install "langsmith[pipecat]"`.

### In development

The voice integrations are in development: each package emits a `LangSmithVoiceBetaWarning` on import and is isolated from the core SDK (`import langsmith` loads none of them; they load only on an explicit `from langsmith.integrations.<voice> import ...`).

### Stack

Subsequent PRs add LiveKit (#3113), OpenAI Realtime (#3114), and Google ADK Live (#3115). This stack replaces #3107.

### Validation

- Unit tests in `tests/unit_tests/wrappers/test_pipecat.py` cover span rewriting, OpenAI→LangChain message normalization, audio attachment, TTL-bounded state, `set_thread_id` injection, and export/exception isolation.
- `ruff check` + `ruff format` + `mypy` clean.
- A `pipecat-ai>=1.0; python_version>='3.11'` marker keeps dependency resolution satisfiable on Python 3.10.

Python only. Authored with the help of an AI agent. :-)

#### PR Dependency Tree

* **PR #3112** 👈
  * **PR #3113**
    * **PR #3114**
      * **PR #3115**

#### PR Dependency Tree

* **PR #3112** 👈
  * **PR #3113**
    * **PR #3114**
      * **PR #3115**

### DIFF
--- a/python/langsmith/_internal/voice/__init__.py
+++ b/python/langsmith/_internal/voice/__init__.py
@@ -1,0 +1,46 @@
+"""Private shared machinery for the voice tracing integrations.
+
+Two independent bases live here, sharing only this namespace (not a common
+class):
+
+* ``base_span_processor`` — Track A: :class:`BaseLangSmithSpanProcessor`, an
+  OpenTelemetry ``SpanProcessor`` shared by the framework integrations that emit
+  their own OTel spans (Pipecat, LiveKit).
+* ``session`` — Track B: :class:`EventSession`, a LangSmith ``RunTree`` builder
+  shared by the integrations that observe a remote event stream and construct
+  the trace themselves (OpenAI Realtime, OpenAI Agents realtime, ADK Live).
+
+Nothing in this package is part of the public API; import from the per-framework
+packages instead.
+"""
+
+from __future__ import annotations
+
+import contextvars
+from typing import Optional
+
+# Per-conversation LangSmith thread id, read by the voice integrations via
+# ``thread_id_from_context``. A ``ContextVar`` — not a module global or a
+# closure — so that concurrent conversations running as separate asyncio tasks
+# each see their own value rather than clobbering a single shared one.
+_VOICE_THREAD_ID: contextvars.ContextVar[Optional[str]] = contextvars.ContextVar(
+    "langsmith_voice_thread_id", default=None
+)
+
+
+def set_thread_id(thread_id: Optional[str]) -> None:
+    """Set the LangSmith thread id for the current (async) context.
+
+    Call this once per conversation, inside that conversation's asyncio task, to
+    group its spans into a LangSmith thread. The voice processors capture it as
+    the conversation's spans start and apply it across the whole trace, so no
+    wiring is needed beyond this call — and it holds even for spans finished on a
+    background task. Because it is stored in a :class:`~contextvars.ContextVar`,
+    concurrent conversations each see their own value; a shared closure would not.
+    """
+    _VOICE_THREAD_ID.set(thread_id)
+
+
+def thread_id_from_context() -> Optional[str]:
+    """Return the thread id set by :func:`set_thread_id` for this context."""
+    return _VOICE_THREAD_ID.get()
--- a/python/langsmith/_internal/voice/base_span_processor.py
+++ b/python/langsmith/_internal/voice/base_span_processor.py
@@ -1,0 +1,340 @@
+"""``BaseLangSmithSpanProcessor`` — Track A's shared OTel span processor.
+
+The framework integrations that emit their own OpenTelemetry spans (Pipecat,
+LiveKit) translate those spans into the ``gen_ai.*`` / ``langsmith.*`` attribute
+namespaces LangSmith's OTLP ingester understands, then forward them to an
+exporter. This base owns everything that translation shares — the downstream
+wrapping, the LangSmith OTLP exporter default, opt-in ``thread_id`` injection,
+the ``gen_ai.*`` message writers, and size-capped audio attachment — so each
+framework subclass implements only
+``_dispatch`` (classify a span by name and rewrite it); the base exports it.
+
+The processor wraps a *downstream* processor rather than being added as a
+sibling: ``on_end`` rewrites attributes and then forwards to the downstream, so
+spans are always translated before export. The default downstream is a
+``BatchSpanProcessor`` around the LangSmith ``OtelExporter`` (see
+:mod:`langsmith.integrations.otel`), which targets LangSmith's
+``/otel/v1/traces`` endpoint with the right auth headers from standard LangSmith
+config — no OTLP env vars required.
+"""
+
+from __future__ import annotations
+
+import base64
+import json
+import logging
+from dataclasses import dataclass
+from typing import Any, Optional
+
+from opentelemetry.sdk.trace import Event, ReadableSpan, SpanProcessor
+from opentelemetry.sdk.trace.export import BatchSpanProcessor
+
+from langsmith._internal.otel._span_utils import rebuild_readable_span
+
+from . import thread_id_from_context
+
+logger = logging.getLogger(__name__)
+
+# Default cap (bytes) on raw audio before base64 encoding. The LangSmith
+# ingester accepts attachments up to ~200MB; base64 inflates by ~1.33x, so a
+# 150MB raw cap encodes to ~200MB — right at that ceiling. Override per
+# processor via ``audio_size_limit_bytes`` (or ``None`` to disable the cap).
+DEFAULT_AUDIO_SIZE_LIMIT = 150_000_000
+
+# Hard ceiling on the per-trace thread-id cache (see ``_remember_thread_id``).
+# Entries are tiny strings and are normally freed at conversation end via
+# ``_forget_thread_id``; this bounds memory for conversations that never reach
+# cleanup (crash, dropped connection), evicting oldest-first.
+_THREAD_ID_CACHE_MAXSIZE = 100_000
+
+
+@dataclass
+class TranslatedSpan:
+    """A span being translated into LangSmith's namespaces before export.
+
+    Wraps the original read-only OTel ``ReadableSpan`` with mutable
+    ``attributes`` and ``events`` seeded from it. Handlers rewrite the copies
+    while translating; :meth:`finalize` builds a fresh ``ReadableSpan`` from them
+    — so OpenTelemetry's private ``span._attributes`` / ``span._events`` are
+    never mutated.
+
+    Created per span in :meth:`BaseLangSmithSpanProcessor.on_end` and threaded
+    through dispatch — no global state. A processor that defers a span (see
+    :meth:`BaseLangSmithSpanProcessor._dispatch`) simply holds the
+    ``TranslatedSpan`` itself until it exports it later, so the in-progress
+    translation outlives the originating ``on_end`` call.
+    """
+
+    span: ReadableSpan
+    attributes: dict[str, Any]
+    events: list[Event]
+
+    @classmethod
+    def of(cls, span: ReadableSpan) -> TranslatedSpan:
+        """Seed a draft from a span's own (read-only) attributes and events."""
+        return cls(span, dict(span.attributes or {}), list(span.events or []))
+
+    def finalize(self) -> ReadableSpan:
+        """Build the export span: the original's fields + our rewritten attrs/events."""
+        return rebuild_readable_span(
+            self.span, attributes=self.attributes, events=self.events
+        )
+
+
+class BaseLangSmithSpanProcessor(SpanProcessor):
+    """Shared base for the OTel→LangSmith framework span processors.
+
+    Subclasses implement :meth:`_dispatch` to classify each ended span and
+    rewrite its attributes via the helpers here; the base exports it. The base
+    handles the downstream/exporter wiring, ``thread_id`` injection, and static
+    metadata stamping.
+    """
+
+    def __init__(
+        self,
+        downstream_processor: Optional[SpanProcessor] = None,
+        *,
+        api_key: Optional[str] = None,
+        project: Optional[str] = None,
+        endpoint: Optional[str] = None,
+        audio_size_limit_bytes: Optional[int] = DEFAULT_AUDIO_SIZE_LIMIT,
+        metadata: Optional[dict[str, Any]] = None,
+    ) -> None:
+        """Create the processor.
+
+        Args:
+            downstream_processor: where rewritten spans are forwarded. Defaults
+                to ``BatchSpanProcessor(OtelExporter(...))`` targeting LangSmith.
+            api_key: LangSmith API key for the default exporter. Defaults to
+                ``LANGSMITH_API_KEY``.
+            project: LangSmith project for the default exporter. Defaults to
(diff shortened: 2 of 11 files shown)


python/pyproject.toml
```diff
@@ -58,6 +58,17 @@ otel = [
...
 
+# Enabled via `pip install langsmith[pipecat]`
+# pipecat-ai requires Python >=3.11; the marker keeps universal resolution
+# satisfiable on 3.10 (the extra is simply unavailable there).
+pipecat = [
+    "pipecat-ai>=1.0; python_version>='3.11'",
+    "opentelemetry-sdk>=1.30.0",
+    "opentelemetry-api>=1.30.0",
+    "opentelemetry-exporter-otlp-proto-http>=1.30.0",
+    "cachetools>=5.0.0",
```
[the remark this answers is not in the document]
carolinedivittorio: @jacoblee93 thoughts on this?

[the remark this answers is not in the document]
jacoblee93: Will defer @sydney-runkle but I think generally we don't upper-bound in Python because it can actively break builds

python/langsmith/_internal/voice/__init__.py
jacoblee93, line 1: I think these should live alongside the other OTEL utils in `langsmith-sdk/python/langsmith/_internal/otel`

Could be `langsmith-sdk/python/langsmith/_internal/otel/voice`

python/langsmith/integrations/_voice/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def warn_in_development(integration: str) -> None:
+    """Emit a :class:`LangSmithVoiceBetaWarning` for a voice integration.
+
+    Called when one of the voice integration packages is imported, to signal
+    that these integrations are in development and their APIs may change.
+    """
+    warnings.warn(
+        f"langsmith.integrations.{integration} is in development; its API may "
```
jacoblee93: Is this what we do for other beta/experimental integrations?

python/langsmith/integrations/_voice/base_span_processor.py
```diff
@@ -0,0 +1,328 @@
...
+
+from . import thread_id_from_context
+
+logger = logging.getLogger(__name__)
+
+# Default audio attachment cap (bytes) before base64 — guards against a long
+# call producing a multi-megabyte span attribute. The LangSmith ingester accepts
+# attachments up to ~200MB, so this conservative default leaves ample headroom
+# even after base64's ~1.33x inflation.
+DEFAULT_AUDIO_SIZE_LIMIT = 20_000_000
```
jacoblee93: This doesn't seem to match what the comment says above

Shouldn't this be >100MB if we support 200MB?

python/langsmith/integrations/_voice/base_span_processor.py
```diff
@@ -0,0 +1,328 @@
...
+
+def _span_attrs(span: ReadableSpan) -> dict[str, Any]:
+    """Return a span's attribute dict for in-place reads and writes.
+
+    OpenTelemetry types ``ReadableSpan._attributes`` as ``Mapping | None``, but
+    at span end it is a populated, mutable dict. These processors rewrite that
+    dict in place (rather than rebuilding the span), so reads and writes go
+    through this accessor, which narrows the type accordingly.
+    """
+    return cast("dict[str, Any]", span._attributes)
```
jacoblee93: Is this an OTEL bug then?

I don't love accessing a private prop like this, at least in JS OTEL is pretty unstable so would expect this to break suddenly

python/langsmith/_internal/voice/base_span_processor.py
```diff
@@ -0,0 +1,328 @@
...
+                ``{LANGSMITH_ENDPOINT}/otel/v1/traces``.
+            audio_size_limit_bytes: skip attaching audio larger than this; set
+                ``None`` to disable the cap.
+            metadata: static ``langsmith.metadata.*`` stamped on every span.
+        """
+        super().__init__()
+        if downstream_processor is None:
+            from langsmith.integrations.otel.processor import OtelExporter
+
+            downstream_processor = BatchSpanProcessor(
```
jacoblee93, line 122: I would check with someone on ingest to make sure that we can handle a batch of spans that each hit the `audio_size_limit_bytes` cap

We can also use a simple span exporter if not

python/langsmith/_internal/voice/base_span_processor.py
```diff
@@ -0,0 +1,328 @@
...
+
+    OpenTelemetry types ``ReadableSpan._attributes`` as ``Mapping | None``, but
+    at span end it is a populated, mutable dict. These processors rewrite that
+    dict in place (rather than rebuilding the span), so reads and writes go
+    through this accessor, which narrows the type accordingly.
+    """
+    return cast("dict[str, Any]", span._attributes)
+
+
+class BaseLangSmithSpanProcessor(SpanProcessor):
```
jacoblee93, line 84: Should we be doing all this voice specific stuff at the exporter level rather than the processor level?

I have often found that easier

python/langsmith/integrations/_voice/base_span_processor.py
```diff
@@ -0,0 +1,328 @@
...
+        # (which would disrupt other processors and the traced app). On error we
+        # log and still export the span untranslated — degraded, not lost.
+        #
+        # _dispatch returns whether the base should export the span now. A
+        # subclass that defers (returns False) owns its own later _export call;
+        # any other return value — including a dispatch failure — leaves
+        # export=True so the span is still forwarded exactly once.
+        export = True
+        try:
+            self._inject_thread_id(span)
```
jacoblee93: Are you sure this gets called in context? I thought OTEL hooks could be backgrounded

In the past what we've done is use a map that stores active conversations by id (assuming each span has some stable attribute tying it back to that conversation) if we need to work across other spans in a conversation

python/langsmith/integrations/_voice/base_span_processor.py
```diff
@@ -0,0 +1,328 @@
...
+
+    # -- shared helpers -------------------------------------------------------
+
+    def _inject_thread_id(self, span: ReadableSpan) -> None:
+        """Stamp ``langsmith.metadata.thread_id`` from the per-context id (opt-in).
+
+        LangSmith needs the thread id on every run for thread-level filtering and
+        token/cost aggregation. Reads the id set via
+        :func:`~langsmith.integrations._voice.set_thread_id` — a ``ContextVar``,
+        so it stays correct across concurrent conversations (``None`` when unset,
```
jacoblee93: See above, I am not sure this will work in all cases

python/langsmith/_internal/voice/base_span_processor.py
```diff
@@ -0,0 +1,328 @@
...
+    def _set_messages(
+        span: ReadableSpan,
+        *,
+        prompt: Optional[list[dict]] = None,
+        completion: Optional[list[dict]] = None,
+    ) -> None:
+        """Write simple role/content turns in the indexed + singular forms.
+
+        For plain conversation turns (STT, TTS, an exchange) where messages have
+        only role + content. Writes the indexed ``gen_ai.prompt.{n}.role/content``
```
jacoblee93, line 278: Does this belong in a base class? Feels more provider specific

python/langsmith/integrations/_voice/base_span_processor.py
carolinedivittorio: Yeah, updated to 150 MB.

python/langsmith/integrations/_voice/base_span_processor.py
carolinedivittorio: It's not a bug. It's because the ReadableSpan is meant to be read only. I'll do what the strands exporter does which is to recreate a ReadableSpan instead of updating the attributes.

python/langsmith/_internal/voice/base_span_processor.py
carolinedivittorio: It would be nice, but we need to hold the root span open so we can attach audio, which would make the exporter really complicated to write. (and goes against what it's meant to do)

python/langsmith/_internal/voice/base_span_processor.py
carolinedivittorio: Only one of the spans should have audio, so we should be fine!

python/langsmith/integrations/_voice/__init__.py
carolinedivittorio: Yeah there's a beta decorator we should use instead - updating. thanks for catching!

python/langsmith/_internal/voice/__init__.py
carolinedivittorio: that feels off given everything that uses it lives under langsmith/integrations. I was hoping that the underscore would signify it being internal & the methods being specific to the integrations/voice specifically. I'm happy to move if you feel strongly.

python/langsmith/_internal/voice/base_span_processor.py
carolinedivittorio: the goal here isn't to rewrite the provider specific messages format but to translate to gen_ai.* so that langsmith can decode role/input/output formats.

python/langsmith/_internal/voice/__init__.py
jacoblee93: Pushback is that `langsmith/integrations._voice` is a bit easier for an end user to find with stuff like autocomplete

Comes down to if there is ever a case where you would want a user to discover and directly import this. That's a bigger concern than the most logical layout for us. Either way, I think underscoring it is a bit of a half measure here, and it just pollutes the `langsmith.integrations` namespace unnecessarily

python/langsmith/integrations/_voice/base_span_processor.py
jacoblee93: Ya that is good

python/langsmith/_internal/voice/base_span_processor.py
jacoblee93: I don't think we can control how `BatchSpanProcessor` batches things? What is this is running multithreaded?

python/langsmith/_internal/voice/base_span_processor.py
jacoblee93: Got it got it

python/langsmith/integrations/pipecat/__init__.py
```diff
@@ -0,0 +1,90 @@
+"""LangSmith integration for Pipecat."""
+
+from __future__ import annotations
+
+import logging
+from typing import Any, Optional
+
+from langsmith._internal._beta_decorator import warn_beta
+
+from .._voice import set_thread_id, thread_id_from_context
```
jacoblee93: nit: I'm not a Python expert and don't know why but I don't think we use relative paths for other integrations

python/langsmith/integrations/_voice/base_span_processor.py
```diff
@@ -0,0 +1,404 @@
...
+    events: list[Event]
+
+    @classmethod
+    def of(cls, span: ReadableSpan) -> TranslatedSpan:
+        """Seed a draft from a span's own (read-only) attributes and events."""
+        return cls(span, dict(span.attributes or {}), list(span.events or []))
+
+    def finalize(self) -> ReadableSpan:
+        """Build the export span: the original's fields + our rewritten attrs/events."""
+        return ReadableSpan(
```
jacoblee93: Mmm I think we're not technically supposed to do this too:

https://opentelemetry-python.readthedocs.io/en/latest/sdk/trace.html#opentelemetry.sdk.trace.Span

I think better than direct accessing a private prop though - let's make it a shared method with Strands too?

python/langsmith/integrations/pipecat/processor.py
```diff
@@ -0,0 +1,348 @@
...
+
+
+class PipecatLangSmithSpanProcessor(BaseLangSmithSpanProcessor):
+    """Enriches Pipecat's OTel spans with LangSmith-compatible attributes."""
+
+    def __init__(
+        self,
+        downstream_processor: Optional[Any] = None,
+        *,
+        llm_span_kind: str = "llm",
```
jacoblee93, line 74: What's the use case for making this configurable?

python/langsmith/integrations/_voice/base_span_processor.py
```diff
@@ -0,0 +1,404 @@
...
+                pass
+        return {
+            "type": "tool_call",
+            "id": raw_call.get("id"),
+            "name": fn.get("name"),
+            "args": args if isinstance(args, dict) else {},
+        }
+
+    @classmethod
+    def _to_langchain_message(cls, msg: dict) -> dict:
```
jacoblee93: I don't actually think we need this? LangSmith should handle OpenAI messages fine

If we do need it, can we double check to make sure there isn't already a util? If there isn't a util, can we just make this a util?

jacoblee93 (approved):

python/langsmith/integrations/pipecat/processor.py
carolinedivittorio: If users overwrite the llm node, we don't want nested llm runs in our trace. So if that's the case, then llm becomes a chain, and the the actual inference becomes the llm run.

python/langsmith/_internal/voice/base_span_processor.py
carolinedivittorio: Let me investigate. I think realistically we shouldn't hit this bc most conversations will be well below the limit, but I can check

python/langsmith/integrations/pipecat/__init__.py
carolinedivittorio: ok let me look. thanks

carolinedivittorio: switched to full export path.

python/langsmith/integrations/_voice/base_span_processor.py
carolinedivittorio: removed it!

python/langsmith/integrations/_voice/base_span_processor.py
carolinedivittorio: made it shared.

PR #3113, opened 2026-06-29
feat(integrations): add LiveKit voice tracing integration

carolinedivittorio:
## LiveKit voice tracing

Adds `langsmith.integrations.livekit`: `configure_livekit` + `LiveKitLangSmithSpanProcessor`, an OTel span processor that rewrites LiveKit's `lk.*` spans into LangSmith's `gen_ai.*` / `langsmith.*` namespaces, defers the conversation root span, and rolls up the per-turn transcript. Works for both the STT/LLM/TTS cascade and speech-to-speech models.

**Audio recording:**
- Development embeds the local `audio.ogg` LiveKit writes under `ctx.session_directory` (`audio_path_provider`).
- Production records with [LiveKit Egress](https://docs.livekit.io/home/egress/overview/) and attaches the file via `expect_recording` / `complete_recording`. Egress finishes uploading after the call, so the processor holds the root span open until the host supplies the downloaded bytes, then embeds them as a real attachment (a metadata URL is not a playable attachment, and OTLP run ids are assigned server-side, so after-the-fact attachment is not possible).

Install: `pip install "langsmith[livekit]"`. Builds on the shared `_voice` foundation from #3112.

Authored with the help of an AI agent.

#### PR Dependency Tree

* **PR #3112**
  * **PR #3113** 👈
    * **PR #3114**
      * **PR #3115**

#### PR Dependency Tree

* **PR #3113** 👈
  * **PR #3114**
    * **PR #3115**

### DIFF
--- a/python/langsmith/integrations/livekit/__init__.py
+++ b/python/langsmith/integrations/livekit/__init__.py
@@ -1,0 +1,83 @@
+"""LangSmith integration for LiveKit Agents."""
+
+from __future__ import annotations
+
+import logging
+from pathlib import Path
+from typing import Any, Callable, Optional
+
+from langsmith._internal._beta_decorator import warn_beta
+from langsmith._internal.voice import set_thread_id, thread_id_from_context
+
+from .processor import LiveKitLangSmithSpanProcessor
+
+logger = logging.getLogger(__name__)
+
+__all__ = [
+    "LiveKitLangSmithSpanProcessor",
+    "configure_livekit",
+    "set_thread_id",
+    "thread_id_from_context",
+]
+
+
+@warn_beta
+def configure_livekit(
+    *,
+    audio_path_provider: Optional[Callable[[], Optional[Path]]] = None,
+    api_key: Optional[str] = None,
+    project: Optional[str] = None,
+    endpoint: Optional[str] = None,
+    **kwargs: Any,
+) -> Optional[LiveKitLangSmithSpanProcessor]:
+    """Enable LangSmith tracing for a LiveKit Agents worker.
+
+    Builds a ``TracerProvider`` with a :class:`LiveKitLangSmithSpanProcessor`
+    (which rewrites LiveKit's ``lk.*`` spans for LangSmith and exports them to
+    LangSmith's OTLP endpoint) and registers it as both LiveKit's tracer provider
+    and the OTel global. Call before starting the worker.
+
+    To manage your own ``TracerProvider`` instead, skip this function: construct
+    :class:`LiveKitLangSmithSpanProcessor` directly, add it to your provider, and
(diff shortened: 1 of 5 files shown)


python/langsmith/integrations/livekit/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+
+from __future__ import annotations
+
+import logging
+from pathlib import Path
+from typing import Any, Callable, Optional
+
+from langsmith._internal._beta_decorator import warn_beta
+
+from .._voice import set_thread_id, thread_id_from_context
```
jacoblee93: See my comments on Pipecat, we tend to prefer absolute paths

python/langsmith/integrations/livekit/processor.py
```diff
@@ -0,0 +1,704 @@
...
+            self._set_messages(
+                tspan, completion=[{"role": "assistant", "content": str(transcript)}]
+            )
+        self._exclude_from_message_view(tspan)
+
+    def _handle_llm_request(self, tspan: TranslatedSpan) -> None:
+        """``llm_request``: the chat-completion call.
+
+        LiveKit records the request's I/O as span events: one
+        ``gen_ai.{system,user,assistant,tool}.message`` event per chat-context
```
jacoblee93, line 270: So they're misusing the conventions?

python/langsmith/integrations/livekit/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+    Returns:
+        The processor, or ``None`` if LiveKit / OpenTelemetry aren't installed.
+    """
+    try:
+        from opentelemetry import trace as otel_trace
+        from opentelemetry.sdk.trace import TracerProvider
+
+        from livekit.agents import telemetry  # type: ignore[import-not-found]
+    except ImportError as e:
+        logger.warning("Missing dependency for LiveKit tracing: %s", e)
```
jacoblee93, line 69: Should this error include an install command to fix the issue?

jacoblee93 (approved):

python/langsmith/integrations/livekit/processor.py
carolinedivittorio: It's a convention that was an older, experimental convention for a while

python/langsmith/integrations/livekit/__init__.py
carolinedivittorio: I'll update ina  follow up!

python/langsmith/integrations/livekit/__init__.py
carolinedivittorio: yep sorry had not updated here yet!

PR #3116, opened 2026-06-29
fix(py): serialize tuple-keyed trace dictionaries

### LINKED ISSUE #3071: Trace outputs silently dropped when run I/O contains a dict with tuple keys (OPT_NON_STR_KEYS gap)
### Summary

When a traced run's inputs/outputs contain a `dict` with a **tuple** key (or any key type `orjson`'s `OPT_NON_STR_KEYS` doesn't support), LangSmith fails to serialize the run and **silently drops its outputs**, logging:

```
WARNING:langsmith.run_helpers:Unable to process trace outputs: TypeError('keys must be str, int, float, bool or None, not tuple')
```

The traced function returns normally; only the trace data is lost.

### Repro (langsmith 0.9.0, orjson 3.11.9, Python 3.12)

```python

@traceable
def f():
    return {(1, 2): "value"}   # tuple-keyed dict

f()   # logs: Unable to process trace outputs: TypeError('keys must be str...not tuple')
```

Or directly against the serializer:

```python
dumps_json({(1, 2): "x"})   # TypeError: keys must be str... not tuple
```

### Root cause (`langsmith/_internal/_serde.py::dumps_json`)

1. `orjson.dumps(obj, ..., OPT_NON_STR_KEYS)` raises `TypeError: Dict key must a type serializable with OPT_NON_STR_KEYS` — `OPT_NON_STR_KEYS` supports str/int/float/bool/None/datetime/uuid/enum, **not tuples** (same family of limitation as ijl/orjson#156, #604, #414).
2. It falls back to stdlib `json.dumps(obj, default=_serialize_json)`, which also raises `keys must be str...not tuple`.
3. `default=_serialize_json` coerces non-serializable **values** but is never invoked for dict **keys**, so neither path can recover. The `_internal/_orjson.py` json-fallback shim (used when orjson is unavailable) has the same gap (`# for now, don't do anything for this case`).

### Suggested fix

Recursively coerce non-`(str | int | float | bool | None)` dict **keys** to `str` before serialization (so a tuple/other key degrades to a string instead of dropping the whole run's outputs). Happy to send a PR.

### Notes

In our case the offending dict originates in a langchain/langgraph-internal run, not one we pass to `@traceable` directly — so `process_inputs`/`process_outputs` on our own traceables can't reach it; the fix needs to live in the serializer.

jacoblee93:
### Description
Fixes #3071 by keeping the normal orjson path unchanged, then retrying serialization with unsupported dict keys normalized to strings so tuple-keyed run I/O is not dropped. The wrapped fallback also handles tuple-keyed dicts returned from `to_dict`/`model_dump`.

### Test Plan
- [x] Added coverage for top-level, nested, and custom-serializer tuple-keyed dicts
- [x] `make -C /workspace/langsmith-sdk/python format`
- [x] `TEST="tests/unit_tests/test_client.py::test__dumps_json_normalizes_unsupported_dict_keys tests/unit_tests/test_client.py::test__dumps_json" make -C /workspace/langsmith-sdk/python tests`
- [x] `make -C /workspace/langsmith-sdk/python lint` (ruff passed; mypy fails on unrelated Python 3.14 `Executor.map` signature in `langsmith/utils.py`)

Made by [Open SWE](https://openswe.vercel.app/agents/c25a00bb-7509-d577-aed4-c3bca4139d1f)

### DIFF
--- a/python/langsmith/_internal/_serde.py
+++ b/python/langsmith/_internal/_serde.py
@@ -151,7 +151,7 @@ def _normalize_json_keys(obj: Any) -> Any:
             norm_key: Any = (
                 key if isinstance(key, _JSON_KEY_TYPES) else str(_simple_default(key))
             )
-            if norm_key in new and norm_key != key:
+            if norm_key in new:
                 logger.debug(
                     "Dict key collision during JSON key normalization: "
                     f"{key!r} maps to {norm_key!r}, which already exists; "
--- a/python/langsmith/_internal/_serde.py
+++ b/python/langsmith/_internal/_serde.py
@@ -153,9 +153,8 @@ def _normalize_json_keys(obj: Any) -> Any:
             )
             if norm_key in new:
                 logger.debug(
-                    "Dict key collision during JSON key normalization: "
-                    f"{key!r} maps to {norm_key!r}, which already exists; "
-                    f"the previous value will be overwritten."
+                    "Dict key collision during JSON key normalization; "
+                    "an existing value will be overwritten."
                 )
             new[norm_key] = _normalize_json_keys(value)
         return new
--- a/python/langsmith/_internal/_serde.py
+++ b/python/langsmith/_internal/_serde.py
@@ -124,13 +124,8 @@ def _serialize_json(obj: Any) -> Any:
         return str(obj)
 
 
-def _normalize_json_keys(obj: Any) -> tuple[Any, bool]:
+def _normalize_json_keys(obj: Any) -> Any:
     """Recursively stringify dict keys that orjson will reject.
-
-    Returns the normalized object and whether any key was actually coerced
-    (the flag lets callers such as ``dumps_json`` skip a guaranteed-to-fail
-    orjson retry when the original failure was unrelated to keys, e.g. lone
-    UTF-16 surrogates in a value).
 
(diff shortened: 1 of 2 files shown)


jacoblee93:
Want to TAL as well @tnorlund?

tnorlund:
Looks really good. The silent-drop (TypeError, "Unable to process trace outputs" warning, outputs dropped) has been fixed and I confirmed. I like the tests too.

One edge case though. When a dict has _both_ a non-str key and a literal string key that collides with its stringified form (e.g. `{(1,2): 'a', '(1, 2)': 'b'}`), normalization silently merges them. The later one wins. Probably fine as a documented limitation, but flagging in case you'd want to note it. 

LGTM

jacoblee93:
Thanks for taking a look @tnorlund!

> One edge case though. When a dict has both a non-str key and a literal string key that collides with its stringified form (e.g. {(1,2): 'a', '(1, 2)': 'b'}), normalization silently merges them. The later one wins. Probably fine as a documented limitation, but flagging in case you'd want to note it.

Yeah figured this would be pretty rare but maybe better to log a warning if this happens

langchain-infra (approved):

PR #3142, opened 2026-07-01
feat(integrations): helpers to build LangSmith run URLs from OTel spans [LSDK-273]

harisaiharish:
## Description
Add public helpers in `langsmith.integrations.otel` to map a native OpenTelemetry span to its LangSmith run URL without polling or waiting for async ingest. `langsmith_run_id_from_otel_span_id` deterministically derives the run UUID from a span ID (int or bytes) fully offline, and `get_langsmith_run_url_for_span` builds the full run URL for a given project, mirroring `Client.get_run_url`.

## Self-Hosted Release Note
none

## Test Plan
- [x] Added unit tests for int→UUID and bytes→UUID mapping against known vectors, offline operation without opentelemetry, and URL formatting with a mocked client.


python/langsmith/integrations/otel/__init__.py
```diff
@@ -1,15 +1,139 @@
...
+    if span_id is None:
+        raise TypeError(
+            "Expected an OTel span, span context, or object with a span_id; "
+            f"got {type(span)}"
+        )
+    return span_id
+
+
+def get_langsmith_run_url_for_span(
+    span: Any,
```
QuentinBrosse: Should we support only `Span` and `SpanContext` instead of `Any`? Same for `_extract_span_id`

python/langsmith/integrations/otel/__init__.py
```diff
@@ -1,15 +1,139 @@
...
+        span: An OTel span, span context, or any object exposing `span_id`
+            (directly or via `get_span_context()`).
+        project_id: The LangSmith project (session) ID. Offline path.
+        project_name: The project name, resolved to an ID via the network.
+        client: Optional `Client` used to resolve tenant/host/project.
+
+    Returns:
+        The full LangSmith run URL.
+    """
+    from langsmith.client import Client
```
QuentinBrosse: Can we move to top?

python/langsmith/integrations/otel/__init__.py
```diff
@@ -1,15 +1,139 @@
...
+        )
+    return span_id
+
+
+def get_langsmith_run_url_for_span(
+    span: Any,
+    *,
+    project_id: Optional[Union[uuid.UUID, str]] = None,
+    project_name: Optional[str] = None,
+    client: Optional["Client"] = None,
```
QuentinBrosse: Should we require only a `project_id: UUID` and let the caller do the rest?

python/langsmith/integrations/otel/__init__.py
harisaiharish: yeah good call

python/langsmith/integrations/otel/__init__.py
harisaiharish: fixed

python/langsmith/integrations/otel/__init__.py
harisaiharish: good idea

python/langsmith/integrations/otel/__init__.py
```diff
@@ -70,24 +71,23 @@ def langsmith_run_id_from_otel_span_id(span_id: Union[int, bytes]) -> uuid.UUID:
...
-            "Expected an OTel span, span context, or object with a span_id; "
-            f"got {type(span)}"
+            f"Expected an OTel span or span context with a span_id; got {type(span)}"
         )
     return span_id
 
 
 def get_langsmith_run_url_for_span(
-    span: Any,
+    span: "Union[Span, SpanContext]",
```
QuentinBrosse: Why quotes?

python/langsmith/integrations/otel/__init__.py
```diff
@@ -70,24 +71,23 @@ def langsmith_run_id_from_otel_span_id(span_id: Union[int, bytes]) -> uuid.UUID:
     return uuid.UUID(bytes=raw)
 
 
-def _extract_span_id(span: Any) -> int:
+def _extract_span_id(span: "Union[Span, SpanContext]") -> int:
     """Extract the integer span ID from an OTel span or span context."""
-    ctx = span.get_span_context() if hasattr(span, "get_span_context") else span
+    get_ctx = getattr(span, "get_span_context", None)
```
QuentinBrosse: Why not `span.get_span_context()`, now that it's typed?

python/langsmith/integrations/otel/__init__.py
QuentinBrosse: Why quotes?

QuentinBrosse (approved):

python/langsmith/integrations/otel/__init__.py
harisaiharish: dropped them

python/langsmith/integrations/otel/__init__.py
harisaiharish: yep used isinstance and then get_span_context

PR #3144, opened 2026-07-01
feat(integrations): attribute voice traces via ls_integration metadata

carolinedivittorio:
#### PR Dependency Tree

* **PR #3144** 👈
  * **PR #3145**
    * **PR #3146**
    * **PR #3153**

### DIFF
--- a/python/langsmith/_internal/voice/helpers.py
+++ b/python/langsmith/_internal/voice/helpers.py
@@ -1,4 +1,4 @@
-"""Shared helpers for the Track-B ``EventSession`` adapters.
+"""Shared helpers for the voice integration adapters.
 
 * ``dump_event`` — best-effort conversion of an event object to a plain dict
   (Pydantic ``model_dump`` → ``dict`` → ``repr`` fallback).
@@ -15,6 +15,8 @@ import logging
 from typing import Any, Callable
 
 logger = logging.getLogger(__name__)
+
+__all__ = ["dump_event", "scrub", "observe_safely"]
 
 # Longest string kept on a span before truncating. Transcripts are short; this
 # only ever trims an unexpectedly large blob.
--- a/python/langsmith/_internal/voice/session.py
+++ b/python/langsmith/_internal/voice/session.py
@@ -430,6 +430,8 @@ def start_session(
     max_audio_seconds: Optional[float] = None,
     client: Optional[Client] = None,
     replicas: Optional[Sequence[WriteReplica]] = None,
+    integration: str,
+    integration_version: Optional[str] = None,
 ) -> EventSession:
     """Create and post the conversation root span, returning an ``EventSession``.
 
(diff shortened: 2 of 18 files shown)


python/langsmith/_internal/voice/session.py
```diff
@@ -447,9 +449,19 @@ def start_session(
...
+    use) so LangSmith can attribute the trace to a specific integration and the
+    framework version in use. Set authoritatively — a caller cannot shadow them
+    via ``metadata``. ``integration_version`` may be ``None`` when the framework
+    version can't be resolved.
     """
     # Mark the root as an audio-modality run (these are voice conversations).
     md = {"ls_modality": "audio", "thread_id": thread_id, **(metadata or {})}
+    # Integration attribution wins over any caller-supplied metadata.
+    md["ls_integration"] = integration
+    md["ls_integration_version"] = integration_version
```
carolinedivittorio:
```suggestion
md = {

"ls_modality": "audio", "
thread_id": thread_id, 
ls_integration: integration
ls_integration_version: integration_version, 


**(metadata or {})}
```

carolinedivittorio: Inlined into the dict literal as suggested — with one tweak: I kept `ls_integration` / `ls_integration_version` *after* `**(metadata or {})` rather than before it. Placed before the spread, a caller passing `metadata={"ls_integration": ...}` would override the real integration id; after the spread they stay authoritative (there's a `test_caller_metadata_cannot_shadow_integration` covering this). Same clean single-literal shape either way.

python/langsmith/_internal/voice/helpers.py
```diff
@@ -21,6 +23,22 @@
 MAX_STR = 2000
 
 
+def get_package_version(package_name: str) -> Optional[str]:
```
jacoblee93: This must already exist?

jacoblee93: If it doesn't exist this might be slow - should we add a functools cache?

python/langsmith/_internal/voice/helpers.py
```diff
@@ -7,15 +7,22 @@
...
 from __future__ import annotations
 
 import logging
 from typing import Any, Callable
 
+from langsmith._internal._package_version import get_package_version
+
 logger = logging.getLogger(__name__)
 
+__all__ = ["dump_event", "scrub", "observe_safely", "get_package_version"]
```
jacoblee93: Hmm do we need to re-export? Would prefer not since this isn't end-user facing

python/langsmith/integrations/google_adk_live/__init__.py
jacoblee93, line 1: Hmm what happened here? Bad merge?

jacoblee93 (approved):

PR #3145, opened 2026-07-01
feat(integrations): set ls_provider/ls_model_name on voice llm runs for cost

carolinedivittorio:
#### PR Dependency Tree

* **PR #3145** 👈
  * **PR #3146**
  * **PR #3153**

### DIFF
--- a/python/langsmith/integrations/google_adk_live/_plugin.py
+++ b/python/langsmith/integrations/google_adk_live/_plugin.py
@@ -32,7 +32,7 @@ Trace shape — one conversation = one trace::
     realtime_session                       (root; transcript + stereo WAV)
     ├── input_transcription                (curated user_message — user speech)
     ├── get_weather                        (tool — args in / response out; real latency)
-    ├── output_transcription               (agent speech + token usage)
+    ├── output_transcription               (agent speech — llm run: usage + cost)
     ├── turn_complete                      (marker)
     └── interrupted                        (marker — on barge-in)
 """
@@ -142,6 +142,19 @@ class _LiveEventView:
         return self._transcript("output_transcription", final_only=True)
 
 
+def _resolve_model(invocation_context: Any) -> str | None:
+    """Best-effort model name for the run's agent (for ``ls_model_name``).
+
+    ADK's ``LlmAgent.model`` is either a model-name string or a ``BaseLlm``
+    instance (whose ``.model`` holds the name). Returns ``None`` when neither is
+    available so cost simply isn't priced rather than mis-attributed.
+    """
+    model = getattr(getattr(invocation_context, "agent", None), "model", None)
+    if isinstance(model, str):
+        return model or None
+    return getattr(model, "model", None) or None
+
+
 def _usage_metadata(event: Any) -> dict[str, int] | None:
     """Map an ADK event's ``usage_metadata`` to LangSmith token usage, if any.
 
@@ -171,8 +184,10 @@ class _AdkLiveTracer:
     tracer per concurrent conversation — the plugin keys them by ADK session id.
     """
 
-    def __init__(self, session: EventSession) -> None:
+    def __init__(self, session: EventSession, *, model: str | None = None) -> None:
         self._trace = session
+        # Model name for the agent-response ``llm`` span's ls_model_name.
+        self._model = model
         # Open tool spans awaiting their function_response, keyed by
         # FunctionCall.id (falling back to the tool name when ADK omits an id).
         # A FIFO list per key disambiguates repeated/parallel identical calls.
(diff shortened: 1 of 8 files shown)


python/langsmith/integrations/google_adk_live/_plugin.py
```diff
@@ -136,6 +136,19 @@ def final_agent_transcript(self) -> str | None:
         return self._transcript("output_transcription", final_only=True)
 
 
+def _resolve_model(invocation_context: Any) -> str | None:
```
carolinedivittorio, line 145: why is this Any?

python/langsmith/integrations/openai_realtime/_connection.py
```diff
@@ -215,6 +219,13 @@ def observe(self, event: Any) -> None:
         if etype is None:
             return  # not a recognizable Realtime event; nothing to span
 
+        # Capture the session's model (for the llm span's ls_model_name); the
+        # response.done payload has no model field, but the session events do.
+        if etype in ("session.created", "session.updated"):
+            model = getattr(getattr(event, "session", None), "model", None)
+            if model:
```
carolinedivittorio, line 224: do we need this guard? why?

python/langsmith/integrations/openai_realtime/_connection.py
```diff
@@ -303,11 +314,19 @@ def _record_first_audio(self, received_at: float) -> None:
             self._await_audio_since = None
 
     def _record_response_llm(self, run: RunTree, response: Any) -> None:
+        # ls_provider + ls_model_name let LangSmith match a pricing entry; the
+        # usage_metadata already carries audio/cache token detail for correct
+        # (audio-rate) cost.
```
carolinedivittorio: remove

python/langsmith/integrations/openai_realtime/_connection.py
```diff
@@ -201,6 +201,10 @@ def __init__(
         # The currently open event span, kept active until the next event:
         # (event_span_cm, tracing_context_cm). None = nothing open.
         self._open: tuple[Any, Any] | None = None
+        # Model name captured from session.created / session.updated (the
+        # response.done payload doesn't carry it), used to tag the ``llm`` span
+        # so LangSmith can price it. None until a session event arrives.
```
carolinedivittorio: remove

python/langsmith/integrations/openai_realtime/_session.py
```diff
@@ -245,9 +262,13 @@ def __init__(
         session: EventSession,
         *,
         on_message: Optional[Callable[[str, str], None]] = None,
+        model: Optional[str] = None,
     ) -> None:
         self._trace = session
         self._on_message = on_message
+        # Model name for the assistant ``llm`` span's ls_model_name (the Agents
+        # SDK emits no per-response usage, so this is attribution, not cost).
```
carolinedivittorio: remove

python/langsmith/integrations/google_adk_live/_plugin.py
carolinedivittorio: The whole ADK plugin duck-types ADK objects as `Any` — `before_run_callback` / `on_event_callback` take `invocation_context: Any` / `event: Any` and read everything via `getattr`, so we never import ADK's heavy/optional `InvocationContext` type. `_resolve_model` follows that same convention (it just does `getattr(getattr(ctx, "agent", None), "model", None)`). Happy to add a `TYPE_CHECKING` Protocol if you'd prefer stronger typing, but this would be the only spot doing so.

python/langsmith/integrations/openai_realtime/_connection.py
carolinedivittorio: Yes — it guards two cases: (1) a session event whose `session.model` is missing/`None`, where without the guard we'd store `str(None)` → the literal string `"None"` as the model name (breaks price matching); and (2) a later `session.updated` that omits `model` clobbering the value captured at `session.created`. The guard keeps the first non-empty model.

python/langsmith/integrations/openai_realtime/_connection.py
carolinedivittorio: Done — removed.

python/langsmith/integrations/openai_realtime/_connection.py
carolinedivittorio: Done — removed.

python/langsmith/integrations/openai_realtime/_session.py
carolinedivittorio: Done — removed.

python/langsmith/integrations/google_adk_live/_plugin.py
```diff
@@ -171,8 +184,10 @@ class _AdkLiveTracer:
     tracer per concurrent conversation — the plugin keys them by ADK session id.
     """
 
-    def __init__(self, session: EventSession) -> None:
+    def __init__(self, session: EventSession, *, model: str | None = None) -> None:
         self._trace = session
+        # Model name for the agent-response ``llm`` span's ls_model_name.
```
carolinedivittorio, line 189: remove

jacoblee93 (approved):

PR #3146, opened 2026-07-01
feat(integrations): standardize voice usage (audio detail, stt/tts duration & characters)

carolinedivittorio:
#### PR Dependency Tree

* **PR #3145**
  * **PR #3146** 👈

### DIFF
--- a/python/langsmith/integrations/google_adk_live/_plugin.py
+++ b/python/langsmith/integrations/google_adk_live/_plugin.py
@@ -155,24 +155,69 @@ def _resolve_model(invocation_context: Any) -> str | None:
     return getattr(model, "model", None) or None
 
 
-def _usage_metadata(event: Any) -> dict[str, int] | None:
+def _audio_tokens(details: Any) -> int:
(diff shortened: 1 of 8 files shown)


python/langsmith/integrations/google_adk_live/_plugin.py
```diff
@@ -149,24 +149,72 @@ def _resolve_model(invocation_context: Any) -> str | None:
...
+    for key, attr in (
+        ("input_tokens", "prompt_token_count"),
+        ("output_tokens", "candidates_token_count"),
+        ("total_tokens", "total_token_count"),
+    ):
+        value = getattr(um, attr, None)
+        if isinstance(value, int):
+            usage[key] = value
+    if not usage:
+        return None
```
carolinedivittorio, line 200: why did you change this? it was easier to read before

carolinedivittorio: Reverted — the totals are back to the original `mapping = {...}` dict + comprehension. The only additions now are the audio/cache/reasoning detail blocks appended after it (needed so Gemini Live bills audio tokens at audio rates rather than text rates).

PR #3154, opened 2026-07-02
fix: default to 'latest' for prompt identifier with trailing colon

Osamaali313:
### Problem

`parse_prompt_identifier` (`python/langsmith/utils.py`, also reached via `parse_hub_identifier`) resolves the commit part of a prompt identifier:

```python
parts = identifier.split(":", 1)
owner_name = parts[0]
commit = parts[1] if len(parts) > 1 else "latest"
```

For an identifier with a **trailing colon and no commit** — e.g. `"name:"` or `"owner/name:"` — `split(":", 1)` yields `["name", ""]`, so `len(parts) > 1` is `True` and `commit` becomes the **empty string** instead of defaulting to `"latest"`.

That empty commit flows into prompt URL/path building: `Client._get_prompt_url` produces a URL ending in a bare colon (`.../hub/-/name:`), and `pull_prompt` only treats `"latest"`/`None` as "fetch latest", so an empty string is used as a literal commit ref.

```
parse_prompt_identifier("name")        -> ("-", "name", "latest")   ✓
parse_prompt_identifier("name:")       -> ("-", "name", "")         ✗  (should be "latest")
parse_prompt_identifier("owner/name:") -> ("owner", "name", "")     ✗
```

The JS SDK (`js/src/utils/prompts.ts`) with the same docstring already handles this correctly via `commit || "latest"` — this is a Python/JS parity divergence.

### Fix

```python
commit = parts[1] if len(parts) > 1 and parts[1] else "latest"
```

### Tests

Extended `test_parse_prompt_identifier` with the trailing-colon cases (`"name:"`, `"owner/name:"` → `"latest"`). Fails before the fix, passes after; existing cases unchanged.

### DIFF
--- a/python/langsmith/utils.py
+++ b/python/langsmith/utils.py
@@ -728,7 +728,9 @@ def parse_prompt_identifier(identifier: str) -> tuple[str, str, str]:
 
     parts = identifier.split(":", 1)
     owner_name = parts[0]
-    commit = parts[1] if len(parts) > 1 else "latest"
+    # A trailing colon with no commit (e.g. "name:") splits to ["name", ""];
+    # treat the empty commit as "latest" (matches the JS SDK's `commit || "latest"`).
+    commit = parts[1] if len(parts) > 1 and parts[1] else "latest"
 
     if "/" in owner_name:
         owner, name = owner_name.split("/", 1)
--- a/python/tests/unit_tests/test_utils.py
+++ b/python/tests/unit_tests/test_utils.py
@@ -309,6 +309,13 @@ def test_parse_prompt_identifier():
         "commit",
     )
     assert ls_utils.parse_prompt_identifier("name:commit") == ("-", "name", "commit")
+    # A trailing colon with no commit defaults to "latest" (not "").
+    assert ls_utils.parse_prompt_identifier("name:") == ("-", "name", "latest")
+    assert ls_utils.parse_prompt_identifier("owner/name:") == (
+        "owner",
+        "name",
+        "latest",
+    )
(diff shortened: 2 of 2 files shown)


jacoblee93:
Thanks for the PR!

Not sure we want to encourage this at all - just an oversight that JS parsing logic defaults to `latest`

What's the use-case? Or is this purely for parity?

Osamaali313:
Thanks for taking a look! Honestly it's mostly parity — but there's a small concrete wrinkle behind it.

Today the two "no specific commit" spellings diverge:

- `"name"`  → `parse_prompt_identifier` returns commit `"latest"` → `GET /commits/{owner}/{name}/latest` ✅
- `"name:"` → returns commit `""` → `GET /commits/{owner}/{name}/` (empty hash) — which isn't "latest" and isn't a clean error either; it's just a malformed request.

So the current Python behavior for a trailing colon is the awkward middle: it neither resolves to latest (like JS) nor rejects the input the way other malformed identifiers do (`/name`, `name/`, `a/b/c` all raise `ValueError`).

I went with the JS-parity route, but I don't have a strong real use-case beyond "someone builds the identifier via string concatenation and ends up with a stray trailing colon." If you'd rather not encourage `name:` at all, I'm happy to flip this the other way and have `parse_prompt_identifier` **raise the standard `ValueError`** for a trailing colon instead of defaulting to latest — that still removes the silent empty-hash request, just via a clear error rather than parity. And of course happy to close if you'd prefer to leave it as-is. Let me know which you'd like.

PR #3155, opened 2026-07-02
feat(js): Extends wrapAnthropic JS wrapper to support Claude Managed Agents methods

jacoblee93:


js/src/wrappers/utils/anthropic_managed.ts
jacoblee93, line 1: nit: move to `wrappers/utils/anthropic` since it's not an entrypoint

dqbd (approved):

PR #3164, opened 2026-07-07
feat(client): expose threads and traces resource accessors on Python and JS clients

QuentinBrosse:
## Description

Adds `threads` and `traces` properties to the Python `Client`, Python `AsyncClient`, and JS `Client` that return the generated OpenAPI v2 resources exposed in #3162 — consistent with the existing `runs`, `sandboxes`, and `datasets` accessors (following the pattern established in #3109).

- **Python `Client`** (`client.py`): adds `threads -> AsyncThreadsResource` and `traces -> AsyncTracesResource`, both with the standard backend version check
- **Python `AsyncClient`** (`async_client.py`): adds the same two accessors + type imports
- **JS `Client`** (`client.ts`): adds `get threads(): Threads` and `get traces(): Traces` + imports
- **Python integration tests**: `test_v2_threads.py` (query, list_traces, stats) and `test_v2_traces.py` (query, list_runs)
- **JS integration tests**: `threads.int.test.ts` (query, listTraces, stats) and `traces.int.test.ts` (query, listRuns)

Notes from testing against the live backend:
- `threads.query` requires an explicit `min_start_time`/`max_start_time` window (no default), unlike `traces.query` which defaults to the last 24h.
- `threads.stats` aggregates are computed by an eventually-consistent background job, so that test only checks the call succeeds rather than asserting `turns` is populated.

### DIFF
--- a/python/tests/integration_tests/test_v2_traces.py
+++ b/python/tests/integration_tests/test_v2_traces.py
@@ -1,0 +1,139 @@
+"""Smoke integration tests for the v2 traces resource.
+
+Covers `Client.traces` (query, list_runs), exposed in #3162.
+"""
+
+import datetime
+import logging
+import time
+from datetime import timedelta
+from typing import Any, Callable
+
+import pytest
+
+from langsmith import uuid7
+from langsmith.client import Client
+from langsmith.utils import get_env_var
+
+logger = logging.getLogger(__name__)
+
+
+def _wait_for_sync(
+    condition: Callable[[], bool],
+    max_sleep_time: int = 60,
+    sleep_time: int = 3,
+) -> None:
+    start = time.time()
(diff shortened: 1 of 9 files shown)


python/tests/integration_tests/test_v2_traces.py
```diff
@@ -0,0 +1,139 @@
...
+import pytest
+
+from langsmith import uuid7
+from langsmith.client import Client
+from langsmith.utils import get_env_var
+
+logger = logging.getLogger(__name__)
+
+
+def _wait_for_sync(
```
KiewanVillatel, line 21: nit: we might want to extract those in a helper.

KiewanVillatel (approved):

PR #3183, opened 2026-07-09
fix(py,js)!: make trace sampling deterministic

emil-lc:
## Description
Use stable run-ID hashing for tracing sampling in both Python and JS SDKs, and apply the same decision when creating run feedback. This prevents feedback for sampled-out runs from being ingested and retried.

## Release Note
Tracing sampling is now deterministic by run ID and also filters feedback for sampled-out runs.

Sampling now keys on the trace, not the individual run. With sampling enabled, direct `create_run`/`update_run` (`createRun`/`updateRun`) calls **on child runs** must pass a `trace_id` (or `dotted_order`) to stay consistent with their trace; without one the child's own ID is hashed and can land on the opposite side of the sample. Runs traced via `@traceable`/`RunTree` are unaffected.

## Test Plan
- [x] Python focused sampling tests: `TEST='tests/unit_tests/test_client.py -k sampling' make -C /workspace/langsmith-sdk/python tests`
- [x] JS client/batch sampling tests: `NODE_OPTIONS=--experimental-vm-modules pnpm --dir /workspace/langsmith-sdk/js exec jest src/tests/client.test.ts src/tests/batch_client.test.ts --runInBand --no-colors`
- [x] Python lint: `make -C /workspace/langsmith-sdk/python lint`
- [x] JS lint/typecheck: `pnpm --dir /workspace/langsmith-sdk/js lint && pnpm --dir /workspace/langsmith-sdk/js exec tsc --noEmit`

Linear: LIN-369

Made with [Open SWE](https://github.com/langchain-ai/open-swe)

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -666,6 +666,33 @@ const getTracingSamplingRate = (configRate?: number) => {
   return samplingRate;
 };
 
+const SAMPLING_HASH_MODULUS = 1_000_000n;
+const FNV_64_OFFSET_BASIS = 14_695_981_039_346_656_037n;
+const FNV_64_PRIME = 1_099_511_628_211n;
+const FNV_64_MASK = (1n << 64n) - 1n;
+
+const isSampledById = (identifier?: string | null, samplingRate?: number) => {
+  if (samplingRate === undefined || samplingRate >= 1) {
+    return true;
+  }
+  if (samplingRate <= 0) {
+    return false;
+  }
+  if (identifier === undefined || identifier === null) {
+    return true;
+  }
+  let hashValue = FNV_64_OFFSET_BASIS;
+  const value = identifier.toLowerCase();
+  for (let i = 0; i < value.length; i += 1) {
+    hashValue ^= BigInt(value.charCodeAt(i));
+    hashValue = (hashValue * FNV_64_PRIME) & FNV_64_MASK;
+  }
+  const threshold = BigInt(
+    Math.floor(samplingRate * Number(SAMPLING_HASH_MODULUS)),
+  );
+  return hashValue % SAMPLING_HASH_MODULUS < threshold;
+};
+
 // utility functions
 const isLocalhost = (url: string): boolean => {
   const strippedUrl = url.replace("http://", "").replace("https://", "");
@@ -897,8 +924,6 @@ export class Client implements LangSmithTracingClientInterface {
   private omitTracedRuntimeInfo?: boolean;
 
   private tracingSampleRate?: number;
-
-  private filteredPostUuids = new Set();
 
   private autoBatchTracing = true;
 
@@ -1717,56 +1742,20 @@ export class Client implements LangSmithTracingClientInterface {
   }
 
   // Allows mocking for tests
-  private _shouldSample(): boolean {
-    if (this.tracingSampleRate === undefined) {
-      return true;
-    }
-    return Math.random() < this.tracingSampleRate;
+  private _shouldSample(identifier?: string | null): boolean {
+    return isSampledById(identifier, this.tracingSampleRate);
   }
 
   private _filterForSampling(
     runs: CreateRunParams[] | UpdateRunParams[],
     patch = false,
   ) {
+    void patch;
     if (this.tracingSampleRate === undefined) {
       return runs;
     }
 
-    if (patch) {
-      const sampled = [];
-      for (const run of runs) {
-        if (!this.filteredPostUuids.has(run.trace_id)) {
-          sampled.push(run);
-        } else if (run.id === run.trace_id) {
-          this.filteredPostUuids.delete(run.trace_id);
-        }
-      }
-      return sampled;
-    } else {
-      // For new runs, sample at trace level to maintain consistency
-      const sampled = [];
-      for (const run of runs) {
-        const traceId = run.trace_id ?? run.id;
-
-        // If we've already made a decision about this trace, follow it
-        if (this.filteredPostUuids.has(traceId)) {
-          continue;
-        }
-
-        // For new traces, apply sampling
-        if (run.id === traceId) {
-          if (this._shouldSample()) {
-            sampled.push(run);
-          } else {
-            this.filteredPostUuids.add(traceId);
-          }
-        } else {
-          // Child runs follow their trace's sampling decision
-          sampled.push(run);
-        }
-      }
-      return sampled;
-    }
+    return runs.filter((run) => this._shouldSample(run.id));
   }
 
   private async _getBatchSizeLimitBytes(): Promise<number> {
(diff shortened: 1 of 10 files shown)


js/src/client.ts
```diff
@@ -666,6 +666,33 @@ const getTracingSamplingRate = (configRate?: number) => {
   return samplingRate;
 };
 
+const SAMPLING_HASH_MODULUS = 1_000_000n;
```
jacoblee93: Can we move all this into a util? Don't want these constants getting lost and this file is already pretty long

js/src/client.ts
```diff
@@ -666,6 +666,33 @@ const getTracingSamplingRate = (configRate?: number) => {
...
+    return false;
+  }
+  if (identifier === undefined || identifier === null) {
+    return true;
+  }
+  let hashValue = FNV_64_OFFSET_BASIS;
+  const value = identifier.toLowerCase();
+  for (let i = 0; i < value.length; i += 1) {
+    hashValue ^= BigInt(value.charCodeAt(i));
+    hashValue = (hashValue * FNV_64_PRIME) & FNV_64_MASK;
```
jacoblee93: I don't quite understand what this is doing and why do this vs. something like `xxhash` - can you explain in a comment?

js/src/client.ts
```diff
@@ -1717,56 +1742,20 @@ export class Client implements LangSmithTracingClientInterface {
...
-    return Math.random() < this.tracingSampleRate;
+  private _shouldSample(identifier?: string | null): boolean {
+    return isSampledById(identifier, this.tracingSampleRate);
   }
 
   private _filterForSampling(
     runs: CreateRunParams[] | UpdateRunParams[],
     patch = false,
   ) {
+    void patch;
```
jacoblee93: What's this

js/src/client.ts
emil-lc: This way, the hashing implementation is consistent between Python and JS - so if, e.g., a Python service is ingesting the trace, and a JS service is ingesting the feedback pointing to the same trace, both will be evaluated to exactly the same sampling cutoff. Same if we add Ho/Java/etc later on

emil-lc: It's a tradeoff, native hashing is easier and cleaner but all other things being equal, my personal preference would be to keep the cross-language hashing consistency.

js/src/client.ts
emil-lc: Slop from a previous rework that got past me 🙄 Removed.

js/src/client.ts
emil-lc: Yes - good idea. Restructured.

js/src/client.ts
jacoblee93: Can we not just use `xxhash`? I think both languages have an implementation?

emil-lc: Will do - thanks for pointing out that it's a dependency already

emil-lc: Switched to xxhash

js/src/client.ts
```diff
@@ -3124,7 +3079,7 @@ export class Client implements LangSmithTracingClientInterface {
     }
     // TODO: Untangle types
     const data: UpdateRunParams = { ...run, id: runId };
-    if (!this._filterForSampling([data], true).length) {
+    if (!this._filterForSampling([data]).length) {
       return;
```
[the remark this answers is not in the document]
emil-lc: That's a trade-off for the bare form that does not include a trace ID - then the Run's own ID is used for a sampling decision, as a fallback.

Regular use —including the calls we make internally from the SDK — is covered. External calls to create/updateRun in a sampling environment on a child run would be affected.

js/src/utils/sampling.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+  if (samplingRate <= 0) {
+    return false;
+  }
+  if (identifier === undefined || identifier === null) {
+    return true;
+  }
+  // XXH3-128 over the UTF-8 bytes, matching the Python SDK's `xxhash.xxh3_128`.
+  // The identifier is sampled in when its fraction of the modulus falls below
+  // the rate, which is also expressed in [0, 1).
```
jacoblee93, line 34: I vaguely recall there being some endian difference or something like that in Python vs. JS implementations here, would try to make sure an id sampled in one env is also sampled in the other

js/src/utils/sampling.ts
```diff
@@ -0,0 +1,38 @@
...
+  identifier: string | null | undefined,
+  samplingRate: number | undefined,
+): boolean => {
+  if (samplingRate === undefined || samplingRate >= 1) {
+    return true;
+  }
+  if (samplingRate <= 0) {
+    return false;
+  }
+  if (identifier === undefined || identifier === null) {
```
jacoblee93: Just use `identifer == null`

js/src/client.ts
```diff
@@ -5567,6 +5534,10 @@ export class Client implements LangSmithTracingClientInterface {
       start_time: startTime,
       extend_trace_retention: extendTraceRetention,
     };
+    const samplingId = traceId ?? runId;
+    if (samplingId !== null && !this._shouldSample(samplingId)) {
```
jacoblee93: Usually nicer to just do `!= null` and conflate `undefined` vs `null` as one vs treating them separately

js/src/client.ts
```diff
@@ -5871,6 +5842,8 @@ export class Client implements LangSmithTracingClientInterface {
           feedbackSourceType: "model",
           sessionId: run?.session_id ?? sessionId,
           startTime: run?.start_time,
+          // A targetRunId may name a run in another trace.
+          traceId: runId_ === run?.id ? run?.trace_id : undefined,
```
jacoblee93, line 5847: Hmm this looks unrelated?

python/langsmith/client.py
```diff
@@ -8078,7 +8057,12 @@ def _submit_feedback(**kwargs):
                 feedback_source_type=ls_schemas.FeedbackSourceType.MODEL,
                 project_id=project_id if run is None else None,
                 extra=res.extra,
-                trace_id=getattr(run, "trace_id", None) if run else None,
+                # A target_run_id may name a run in another trace.
+                trace_id=(
```
jacoblee93, line 8062: See above

jacoblee93 (approved):

js/src/utils/sampling.ts
emil-lc: Yep, this is tested explicitly with "golden" UUIDs that return the same result in JS and Python.

python/langsmith/client.py
emil-lc: Replied above.

js/src/client.ts
emil-lc: It is relevant, but I compressed it to a single line, beyond recognition. Updated with a more readable comment now.

js/src/client.ts
emil-lc: fixed

js/src/utils/sampling.ts
emil-lc: fixed

PR #3190, opened 2026-07-10
feat(voice): capture STT/TTS/realtime model costs across voice integrations

carolinedivittorio:
## Summary

Makes model cost / token usage flow correctly through all five voice-agent framework integrations. LangSmith prices runs server-side from `usage_metadata` (token counts + `audio` / `cache_read` / `reasoning` detail buckets) × a per-`(provider, model)` rate. Each integration was dropping cost signal at a different point; this routes what each framework actually emits into `gen_ai.usage.*` / `usage_metadata`.

Per-modality **audio** token detail is the important one — for realtime/live models audio tokens dominate a voice turn and are priced very differently from text, so capturing only aggregate counts (the prior behavior in several paths) mispriced them as text.

## Changes by integration

| Integration | Before | After |
|---|---|---|
| **LiveKit** | LLM cost only (framework sets `gen_ai.usage.*`); TTS/realtime-detail dropped | TTS `characters_count` → tokens; realtime **audio + cached** token detail lifted; LLM `prompt_cached_tokens` → `cache_read` |
| **Pipecat** | LLM cost only; realtime S2S spans fell through untranslated | TTS `metrics.character_count` → tokens; cache/reasoning LLM detail mapped; **realtime S2S** spans (`llm_response`/`llm_setup`/`llm_request`) handled |
| **OpenAI Realtime (raw WS)** | Already correct | Unchanged; mapper refactored, existing audio+cache lock test retained |
| **OpenAI Agents realtime** | No cost (SDK emits no semantic usage event) | Usage recovered from the raw `response.done` (forwarded as `raw_server_event`) and attached to the assistant span |
| **Google ADK Live (Gemini)** | Aggregate token counts only → audio priced as text | Captures the `AUDIO` modality split from `prompt_tokens_details` / `candidates_tokens_details`, plus `cache_read` and `reasoning` |

Shared: added `TranslatedSpan.set_usage(...)` which writes `gen_ai.usage.*` with detail blocks as `str(dict)` — the exact form the OTLP ingester round-trips.

## Known limitation (follow-up)

Cascade **STT** cost is not tracked: LiveKit exposes `STTMetrics.audio_duration` only via a separate `metrics_collected` event (never on the `user_turn` span), and Pipecat emits no STT usage metric at all. Both STT handlers carry a `TODO(voice-cost)` describing the metrics-event hook needed for LiveKit. Realtime/live models are unaffected — their STT is folded into the model's audio tokens.

## Test plan

- [x] `feat`: added 40 unit tests across the five integrations (TTS char lift, LLM cache/reasoning detail, LiveKit realtime audio/cache detail, Pipecat realtime `llm_response`, Agents `response.done` usage capture + single-consume, Gemini audio-modality split incl. plain-string modality)
- [x] Full wrappers suite passes: `pytest tests/unit_tests/wrappers/` → 287 passed
- [x] `ruff format` + `ruff check` clean on changed files
- [x] `mypy` clean on changed modules

### DIFF
--- a/python/langsmith/_internal/voice/translated_span.py
+++ b/python/langsmith/_internal/voice/translated_span.py
@@ -18,6 +18,13 @@ from opentelemetry.sdk.trace import Event, ReadableSpan
 from langsmith._internal.otel._span_utils import rebuild_readable_span
 
 
+def _clean_token_details(details: Optional[dict[str, Any]]) -> dict[str, int]:
+    """Keep only the int-valued detail keys, dropping ``None`` / non-numeric."""
+    if not details:
+        return {}
+    return {k: int(v) for k, v in details.items() if isinstance(v, (int, float))}
+
+
 @dataclass
 class TranslatedSpan:
     """A span being translated into LangSmith's namespaces before export.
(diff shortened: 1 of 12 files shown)


carolinedivittorio:
Superseded by the per-integration stack: #3191 (LiveKit), #3192 (Pipecat), #3193 (OpenAI Realtime), #3194 (Google ADK Live).

PR #3192, opened 2026-07-10
feat(voice): capture Pipecat TTS/realtime model costs

carolinedivittorio:
## Summary

Captures usage data for Pipecat, including pipecat with the realtime model.

## Test plan
- [x] `TestPipecatUsageCapture`: LLM usage + cache/reasoning detail; realtime `llm_response` priced + transcribed; realtime user+agent rollup; wrapper spans → chain
- [x] `pytest tests/unit_tests/wrappers/` → 280 passed
- [x] `ruff format` + `ruff check` clean; `mypy` clean

#### PR Dependency Tree

* **PR #3192** 👈
  * **PR #3193**
    * **PR #3194**

### DIFF
--- a/python/langsmith/integrations/pipecat/_helpers.py
+++ b/python/langsmith/integrations/pipecat/_helpers.py
@@ -22,6 +22,34 @@ def parse_llm_messages(input_data: Any) -> list[dict[str, Any]]:
     return [m for m in raw if isinstance(m, dict)]
 
 
+def _as_int(value: Any) -> Any:
+    """Coerce a numeric span-attribute value to ``int``, or ``None`` otherwise."""
+    return int(value) if isinstance(value, (int, float)) else None
+
+
+def extract_llm_usage(attributes: dict[str, Any]) -> dict[str, Any]:
+    """Read Pipecat's ``gen_ai.usage.*`` span keys into ``set_usage`` kwargs."""
+    usage: dict[str, Any] = {}
+    if (v := _as_int(attributes.get("gen_ai.usage.input_tokens"))) is not None:
+        usage["input_tokens"] = v
+    if (v := _as_int(attributes.get("gen_ai.usage.output_tokens"))) is not None:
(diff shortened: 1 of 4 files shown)


python/langsmith/integrations/pipecat/_helpers.py
```diff
@@ -22,6 +22,34 @@ def parse_llm_messages(input_data: Any) -> list[dict[str, Any]]:
     return [m for m in raw if isinstance(m, dict)]
 
 
+def _as_int(value: Any) -> Any:
+    """Coerce a numeric span-attribute value to ``int``, or ``None`` otherwise."""
+    return int(value) if isinstance(value, (int, float)) else None
+
+
+def extract_llm_usage(attributes: dict[str, Any]) -> dict[str, Any]:
```
jacoblee93, line 30: Can we refactor this into a shared handler with other utilities or is this too bespoke? Maybe we can just pass it a map of span mappings?

```py
def extract_llm_usage(attributes: dict[str, Any], mapping: dict[str, Any]) -> dict[str, Any]:
  
```

Where `mapping` is e.g.:

```py
{
  "gen_ai.usage.input_tokens": "input_tokens",
  ...
}
```

python/langsmith/integrations/pipecat/processor.py
```diff
@@ -46,7 +49,11 @@
     TranslatedSpan,
 )
 
-from ._helpers import build_completion_message, parse_llm_messages
+from ._helpers import (
```
jacoblee93: nit: Missed this on initial review but use absolute paths

jacoblee93 (approved):

python/langsmith/integrations/pipecat/_helpers.py
carolinedivittorio: we can, but I don't think we get a ton out of it. It's only in two places, and here maps pipecat to langsmith.

PR #3193, opened 2026-07-10
feat(voice): capture OpenAI Agents realtime model costs

carolinedivittorio:
## Summary

Part of a 4-PR stack on voice model-cost capture. Stacked on #3192 (Pipecat).

This PR covers **`integrations/openai_realtime`** (both the raw-WebSocket and Agents SDK realtime flavors).

The raw-WS integration already captures audio + cached token detail (unchanged, still covered by its existing test). The gap was the **Agents SDK realtime backend**, which emits no *semantic* usage event:

- Recover token usage from the raw `response.done` the SDK still forwards as a `raw_model_event` (`raw_server_event`), and attach it to that turn's assistant `model` span — so the conversation is priced (audio tokens included) where before it had none.
- Refactor the shared mapper into `usage_metadata_from_usage` so both realtime flavors map through the same canonical shape.

## Test plan

- [x] Added to `test_openai_realtime_agents.py`: `raw_response_usage` helper (extraction + None cases), usage attached to the assistant span with audio/cache detail, and usage consumed once (not inherited by a later turn)
- [x] Existing raw-WS audio/cache detail test retained
- [x] `pytest tests/unit_tests/wrappers/` → 287 passed
- [x] `ruff format` + `ruff check` clean; `mypy` clean on changed modules

#### PR Dependency Tree

* **PR #3193** 👈
  * **PR #3194**

### DIFF
--- a/python/langsmith/integrations/openai_realtime/_session.py
+++ b/python/langsmith/integrations/openai_realtime/_session.py
@@ -15,8 +15,11 @@ per id wins → partials collapse); each finalized message emits one span (a
 curated ``user_message`` or an assistant ``model`` ``llm`` span), grouped into
 turns. Each SDK-run tool becomes a single ``tool`` span spanning its
 ``tool_start``→``tool_end`` pair (so the span duration is the real tool latency);
-``audio_interrupted`` flags the turn. (Token usage isn't available — the SDK has
-no per-response usage events yet.)
+``audio_interrupted`` flags the turn. Token usage — which the SDK exposes no
+*semantic* event for — is recovered from the raw ``response.done`` the SDK still
+forwards as a ``raw_model_event`` (``raw_server_event``), then attached to that
+turn's assistant ``model`` span so the conversation is priced (audio tokens
+included).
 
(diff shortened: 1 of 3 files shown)


python/langsmith/integrations/openai_realtime/_session.py
```diff
@@ -175,6 +181,27 @@ def raw_input_transcript(event: Any) -> tuple[str | None, str] | None:
...
+    ``response.done`` (as a ``raw_server_event``), which carries ``response.usage``
+    — the only place per-turn token counts appear on this backend. Maps it via the
+    shared Realtime usage mapper (so audio/cached detail is kept). Returns ``None``
+    for any other event or a payload without usage.
+    """
+    data = getattr(event, "data", None)
+    if getattr(data, "type", None) != "raw_server_event":
+        return None
+    raw = getattr(data, "data", None)
+    if not isinstance(raw, dict) or raw.get("type") != "response.done":
```
jacoblee93, line 195: Hmm why not just simplify and just make it this check?

jacoblee93 (approved):

PR #3194, opened 2026-07-10
feat(voice): capture Google ADK Live (Gemini) model costs

carolinedivittorio:
## Summary

Final PR of a 4-PR stack on voice model-cost capture. Stacked on #3193 (OpenAI Realtime).

This PR covers **`integrations/google_adk_live`** (Gemini Live).

Gemini Live's usage was captured as aggregate token counts only, so audio tokens were priced as text — wrong, since audio dominates a voice turn. Now capture:

- the **`AUDIO` modality split** from `prompt_tokens_details` / `candidates_tokens_details` (lists of `ModalityTokenCount`) → `input`/`output` `audio` detail,
- `cached_content_token_count` → `cache_read`,
- `thoughts_token_count` → `reasoning`,

so each is priced at its own rate.

## Test plan

- [x] Extended `TestUsageMetadata` in `test_google_adk_live.py`: audio + cache_read + reasoning detail captured, and audio modality accepted as either an enum or a plain string
- [x] `pytest tests/unit_tests/wrappers/` → 287 passed
- [x] `ruff format` + `ruff check` clean; `mypy` clean on changed modules

#### PR Dependency Tree

* **PR #3193**
  * **PR #3194** 👈

### DIFF
--- a/python/langsmith/integrations/google_adk_live/_plugin.py
+++ b/python/langsmith/integrations/google_adk_live/_plugin.py
@@ -159,23 +159,70 @@ def _resolve_model(invocation_context: Any) -> str | None:
     return getattr(model, "model", None) or None
 
 
-def _usage_metadata(event: Any) -> dict[str, int] | None:
+def _audio_tokens(details: Any) -> int | None:
+    """Sum the ``AUDIO`` modality token count from a ``ModalityTokenCount`` list.
+
+    Gemini reports per-modality breakdowns as a list of ``{modality, token_count}``
+    entries; the audio bucket is what the cost engine prices at the (much higher)
+    audio rate. ``MediaModality`` has a single audio member (``AUDIO``), so an
+    exact match is enough. Returns ``None`` when no audio modality is present.
+    """
+    if not isinstance(details, (list, tuple)):
+        return None
+    total = 0
+    found = False
+    for entry in details:
+        modality = getattr(entry, "modality", None)
+        modality = getattr(modality, "value", modality)
+        count = getattr(entry, "token_count", None)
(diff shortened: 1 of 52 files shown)


python/langsmith/integrations/google_adk_live/_plugin.py
```diff
@@ -159,23 +159,69 @@ def _resolve_model(invocation_context: Any) -> str | None:
...
+    """
+    if not isinstance(details, (list, tuple)):
+        return None
+    total = 0
+    found = False
+    for entry in details:
+        modality = getattr(entry, "modality", None)
+        modality = getattr(modality, "value", modality)
+        count = getattr(entry, "token_count", None)
+        if str(modality).upper().endswith("AUDIO") and isinstance(count, int):
```
jacoblee93: How many different audio modalities are there? Is this overly defensive?

jacoblee93 (approved):

python/langsmith/integrations/google_adk_live/_plugin.py
carolinedivittorio: Good call — there's only one audio member in google.genai's `MediaModality` (`MODALITY_UNSPECIFIED, TEXT, IMAGE, VIDEO, AUDIO, DOCUMENT`), and the line above already extracts `.value` so it's a bare `"AUDIO"`. Tightened to an exact match; the suffix check was guarding against nothing real (and would've silently swept in a hypothetical future `*_AUDIO` rather than surfacing it). Done in d12eae67.

carolinedivittorio: oops sorry claude responded

PR #3206, opened 2026-07-14
chore: deprecate run_ids in AsyncClient.add_runs_to_annotation_queue [LSDK-291]

ayoung19:
> [!WARNING]
> Do not merge — gated on public deprecation comms ([LSDK-291](https://linear.app/langchain/issue/LSDK-291)).

## Description

Emits a `DeprecationWarning` when the legacy `run_ids=` form is passed to `AsyncClient.add_runs_to_annotation_queue`, pointing callers to `runs=` (the SmithDB `/runs/by-key` path).

#3241 landed this deprecation for the sync `Client` (`client.py`) and the JS `Client` (`client.ts`), but did not cover `async_client.py` — so the async client was silently missing the warning while its docstring still said the path "will be deprecated in a future release". This fills that gap with wording **identical** to `client.py`: same message, same `#annotation-queues-add-runs` anchor, same Jan 31 2027 removal date.

Non-breaking — no signatures change.

## Notes for reviewers

This PR was originally the Python + JS deprecation. Since #3241 merged into the base branch and covers `client.py` / `client.ts` with better text than this branch had (anchored link + removal date), the base branch's versions of those two files are taken as-is and this branch's edits to them are dropped. That is what the review comments asked for — align with #3241, and add the anchor link, both of which #3241 already does.

Net effect: the diff is now a single file, `python/langsmith/async_client.py`.

Linear: [LSDK-291](https://linear.app/langchain/issue/LSDK-291) · Go/Java analog: langchain-ai/langchainplus#30308


python/langsmith/async_client.py
```diff
@@ -1458,6 +1459,14 @@ async def add_runs_to_annotation_queue(
                 ls_client._serialize_run_key(run, i) for i, run in enumerate(runs)
             ]
         elif run_ids is not None:
+            warnings.warn(
+                "The `run_ids` parameter of `add_runs_to_annotation_queue` is "
```
QuentinBrosse: Can you mention https://docs.langchain.com/langsmith/smithdb-sdk-migration for more info?

python/langsmith/client.py
```diff
@@ -8841,6 +8842,14 @@ def add_runs_to_annotation_queue(
             path = f"/annotation-queues/{_as_uuid(queue_id, 'queue_id')}/runs/by-key"
             json_body = [_serialize_run_key(run, i) for i, run in enumerate(runs)]
         elif run_ids is not None:
+            warnings.warn(
+                "The `run_ids` parameter of `add_runs_to_annotation_queue` is "
```
QuentinBrosse: Can you link https://docs.langchain.com/langsmith/smithdb-sdk-migration for more info?

python/langsmith/async_client.py
ayoung19: fixed

python/langsmith/client.py
ayoung19: fixed

js/src/client.ts
```diff
@@ -5882,6 +5803,12 @@ export class Client implements LangSmithTracingClientInterface {
         }),
       );
     } else {
+      warnOnce(
```
QuentinBrosse, line 5926: Let's align with whatever we decided here? https://github.com/langchain-ai/langsmith-sdk/pull/3241

js/src/client.ts
```diff
@@ -5882,6 +5803,12 @@ export class Client implements LangSmithTracingClientInterface {
...
+        "Passing an array of run-ID strings to addRunsToAnnotationQueue is " +
+          "deprecated and will be removed in a future version. Pass an array of " +
+          "RunKey objects instead, each with runId, sessionId, and startTime. " +
+          "See https://docs.langchain.com/langsmith/smithdb-sdk-migration for details.",
```
QuentinBrosse: Let's add the anchor link?

KiewanVillatel (approved):

PR #3209, opened 2026-07-15
feat: support stable tracing project IDs [closes LSDK-319]

yumuzi234:
## Description
Adds `LANGSMITH_PROJECT_ID` as a validated, stable alternative to `LANGSMITH_PROJECT` in the Python and JS tracing SDKs. If both presented, send warning and use ID, if both undefined, follows existing behavior.

## Release Note
Python and JS SDKs can route traces by stable project ID with `LANGSMITH_PROJECT_ID`. If both name and ID are presented, send warning and use ID, if both undefined, follows the previous behavior.

## Test Plan
- [x] UT


js/src/utils/project.ts
```diff
@@ -10,3 +11,8 @@ export const getDefaultProjectName = () => {
     "default"
   );
 };
+
+export const getDefaultProjectId = () => {
+  const projectId = getEnvironmentVariable("LANGSMITH_PROJECT_ID");
+  return projectId ? assertUuid(projectId, "LANGSMITH_PROJECT_ID") : undefined;
```
dqbd, line 17: _question_: do we need to check for UUIDv7?

yumuzi234: chatted offline
1. I think check better in BE/API
2. We want to support old project id (uuid 4) if they are create earlier, if they want to use that.

PR #3232, opened 2026-07-20
feat(client): shared-run accessor +  runs.share and public.runs itg tests

emil-lc:
## Description

Exposes the generated OpenAPI v2 `public` resource on the langsmith clients and
adds live-backend integration tests for the v2 `runs.share` and `public.runs`
resources.

The `public.runs` (retrieve/query) and `runs.share` (create/delete) resources
already existed in the vendored `_openapi_client`; `runs.share` was already
reachable via `client.runs.share`, but `public` had no top-level accessor. This
PR adds only the thin `public` wrapper accessor and the tests - no generated
`_openapi_client` code is modified.

### Accessor

- **Python `Client`** (`client.py`): adds `public -> AsyncPublicResource`, with
  the standard `_check_backend_version` check, consistent with the existing
  `runs`/`threads`/`traces` accessors.
- **Python `AsyncClient`** (`async_client.py`): adds `public -> AsyncPublicResource`
  (no per-call version check, matching the other async accessors + the imports).
- **JS `Client`** (`client.ts`): adds `get public(): Public`.

### Integration tests

- **Python**: `test_v2_runs_share.py` (`runs.share.create`, `runs.share.delete`)
  and `test_v2_public_runs.py` (`public.runs.query`, `public.runs.retrieve`).
- **JS**: `runs_share.int.test.ts` and `public_runs.int.test.ts` covering the
  same four methods.

Each test creates a throwaway `__test_v2_*` project + root run, mints a share
token via `runs.share.create`, reads it back through the share-token-scoped
`public.runs` endpoints, and deletes the project in cleanup — mirroring the
existing `test_v2_*` conventions (self-contained `langchain_client` fixture,
`_wait_for_*` / `waitUntil` retry wrappers for eventual consistency.

**Note**
Public sharing must be enabled for the org for the tests to pass

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -82,6 +82,7 @@ import { Sandboxes } from "./_openapi_client/resources/sandboxes/sandboxes.js";
 import { Datasets } from "./_openapi_client/resources/datasets/datasets.js";
 import { Threads } from "./_openapi_client/resources/threads.js";
 import { Traces } from "./_openapi_client/resources/traces.js";
+import { Public } from "./_openapi_client/resources/public/public.js";
 import { assertUuid } from "./utils/_uuid.js";
 import { warnOnce } from "./utils/warn.js";
 import { _MIN_BACKEND_VERSION } from "./utils/constants.js";
@@ -1501,6 +1502,11 @@ export class Client implements LangSmithTracingClientInterface {
   /** Access the traces resource (query, listRuns). */
   public get traces(): Traces {
     return this.openAPIClient.traces;
+  }
+
+  /** Access the public shared-run resource (runs.retrieve, runs.query). */
+  public get public(): Public {
+    return this.openAPIClient.public;
   }
 
   private async processInputs(inputs: KVMap): Promise<KVMap> {
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -1504,7 +1504,7 @@ export class Client implements LangSmithTracingClientInterface {
     return this.openAPIClient.traces;
   }
 
-  /** Access the public shared-run resource (runs.retrieve, runs.query). */
+  /** Access the public shared-run resource. */
   public get public(): Public {
     return this.openAPIClient.public;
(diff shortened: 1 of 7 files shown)


python/langsmith/client.py
```diff
@@ -1524,6 +1527,12 @@ def traces(self) -> AsyncTracesResource:
         _check_backend_version(self.info.version)
         return self._get_langsmith_api().traces
 
+    @property
+    def public(self) -> AsyncPublicResource:
```
QuentinBrosse, line 1531: Did we think about the `client.runs.public` vs. `client.public.runs`? I guess the current implementation is the natural 1:1 of the endpoint schema. I was wondering which one is the best for discoverability.

emil-lc: Those are very related concepts indeed, but the `client.public` vs `client.runs` have very different population targets - one is for "anonymous" clients to read, and the other is for authenticated "admins" to write. A user of one would not call the other, and vice versa.

We also have datasets as a shareable resource, and right now at least, it lives under `public.datasets`, even though rendering for Py + Ts is skipped.

QuentinBrosse: Discussed internally.

js/src/client.ts
```diff
@@ -1503,6 +1504,11 @@ export class Client implements LangSmithTracingClientInterface {
     return this.openAPIClient.traces;
   }
 
+  /** Access the public shared-run resource (runs.retrieve, runs.query). */
```
QuentinBrosse: Should we remove that, as the list might be longer? Also, we don't want to maintain this comment when we add new generated methods.
```suggestion
  /** Access the public shared-run resource. */
```

Same comment for the 2 python clients.

QuentinBrosse (approved):

js/src/client.ts
emil-lc: Yeah, makes sense

PR #3233, opened 2026-07-20
feat(py,js): Expose method for calculating replica ids and add concept of a primary replica

jacoblee93:
Allows users to predictably calculate what a run id be remapped to when sent to different replicas

Useful for e.g. leaving feedback on all replica runs going to different destinations

### DIFF
--- a/python/langsmith/run_trees.py
+++ b/python/langsmith/run_trees.py
@@ -77,7 +77,9 @@ class WriteReplica(TypedDict, total=False):
     """
 
 
-_HEADER_SAFE_REPLICA_FIELDS: frozenset[str] = frozenset({"project_name", "updates"})
+_HEADER_SAFE_REPLICA_FIELDS: frozenset[str] = frozenset(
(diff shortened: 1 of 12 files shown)


python/langsmith/run_trees.py
```diff
@@ -702,18 +707,29 @@ def _slice_parent_id(self, parent_id: str, run_dict: dict) -> None:
...
+        if primary is None and project_name == self.session_name:
             return run_dict
 
         if updates and updates.get("reroot", False):
             distributed_parent_id = _DISTRIBUTED_PARENT_ID.get()
             if distributed_parent_id:
                 self._slice_parent_id(distributed_parent_id, run_dict)
 
+        if primary:
+            dup = utils.deepish_copy(run_dict)
```
dqbd, line 731: _question_: why we need to copy here?

jacoblee93: Because the primary replica can have `updates`

dqbd (approved):

PR #3240, opened 2026-07-21
fix(python): promote usage_metadata from run.extra to OTEL span attributes

AXEG0:
## Problem

The `claude_agent_sdk` integration (`langsmith/integrations/claude_agent_sdk/_client.py::TurnLifecycle._set_usage_from_message` and `_transcripts.py::_patch_usage_on_llm_runs`) captures real per-turn token usage from the Claude Agent SDK and stores it in `run.extra["metadata"]["usage_metadata"]`.

The OTEL exporter (`langsmith/_internal/otel/_otel_exporter.py`) only promotes token usage into the `gen_ai.usage.input_tokens` / `gen_ai.usage.output_tokens` / `gen_ai.usage.total_tokens` span attributes from two places:
- the top-level run token fields (`prompt_tokens` / `completion_tokens` / `total_tokens`), or
- `run.outputs[...]["usage_metadata"]`, via `get_unified_run_tokens`.

It never looks at `run.extra["metadata"]["usage_metadata"]`. As a result, any run whose usage was only captured there -- as is the case for every LLM turn produced by the `claude_agent_sdk` integration, and for any other integration that follows the same pattern -- is exported to OTEL with **no usage attributes at all**, so any OTEL backend consuming these spans sees `usage=0` even though LangSmith itself has the correct token counts.

## Repro

1. Trace a Claude Agent SDK session with LangSmith's `claude_agent_sdk` integration and `LANGSMITH_OTEL_ENABLED=true` (or otherwise wire an OTEL exporter/backend).
2. Inspect the exported spans for the LLM turns: `gen_ai.usage.input_tokens` / `gen_ai.usage.output_tokens` / `gen_ai.usage.total_tokens` are absent, even though the corresponding LangSmith run has `extra["metadata"]["usage_metadata"]` populated with real counts.

## Fix

In `_otel_exporter.py`, `_set_io_attributes` now falls back to `run_info["extra"]["metadata"]["usage_metadata"]` when neither the run's top-level token fields nor `run.outputs` already supplied usage. The fallback:
- reuses the existing `_extract_unified_run_tokens` key-extraction helper, so it accepts the same `input_tokens`/`output_tokens` shape as every other source;
- is strictly lowest precedence -- it never overrides usage already found via `run.outputs` or the run token fields;
- does not depend on `run.outputs` being present at all, since usage can arrive on an operation (e.g. a patch) that carries no outputs.

## Tests

Added to `tests/unit_tests/test_otel_exporter.py`, driving `OTELExporter.export_batch` end-to-end against a real `opentelemetry.sdk.trace.TracerProvider` + `InMemorySpanExporter` (no mocking of the exporter logic under test):
- `test_export_batch_promotes_usage_from_extra_metadata` -- usage present only in `extra.metadata.usage_metadata` is promoted to the span.
- `test_export_batch_promotes_usage_from_extra_metadata_without_outputs` -- same, when the operation has no outputs at all.
- `test_export_batch_outputs_usage_takes_precedence_over_extra_metadata` -- usage in `run.outputs` still wins when both sources are present.
- `test_export_batch_no_usage_anywhere_sets_no_usage_attributes` -- no usage source means no usage attributes are set (no false positives).

`make format`, `make lint` (ruff + mypy), and `make tests` (full `tests/unit_tests`, **2134 passed / 6 skipped / 1 xfailed**) are all green at exact head `67f53026592f84252b9375e10f72e7da629aec37`, based on current `main` at `4ddfa249823d102183054c750fc4ba4a9a356899`.

### DIFF
--- a/python/langsmith/_internal/otel/_otel_exporter.py
+++ b/python/langsmith/_internal/otel/_otel_exporter.py
@@ -612,7 +612,7 @@ class OTELExporter:
                 span.set_attribute(GEN_AI_SERIALIZED_DOC, serialized["doc"])
 
         # Set inputs/outputs if available
-        self._set_io_attributes(span, op)
+        self._set_io_attributes(span, op, run_info)
 
     def _set_gen_ai_system(self, span: Span, run_info: dict) -> None:
         """Set the gen_ai.system attribute on the span based on the model provider.
@@ -702,12 +702,18 @@ class OTELExporter:
                 GEN_AI_REQUEST_PRESENCE_PENALTY, invocation_params["presence_penalty"]
             )
 
-    def _set_io_attributes(self, span: Span, op: SerializedRunOperation) -> None:
+    def _set_io_attributes(
+        self, span: Span, op: SerializedRunOperation, run_info: dict
+    ) -> None:
         """Set input/output attributes on the span.
 
         Args:
             span: The span to set attributes on.
             op: The serialized run operation.
+            run_info: The deserialized run info. Used to fall back to token
+                usage recorded under ``extra.metadata.usage_metadata`` (e.g.
+                by the claude_agent_sdk integration) when neither the
+                top-level run token fields nor the outputs carry usage.
         """
         if op.inputs:
             try:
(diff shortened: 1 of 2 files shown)


AXEG0:
Rebuilt on current `main`: exact head `67f53026592f84252b9375e10f72e7da629aec37`, base `4ddfa249823d102183054c750fc4ba4a9a356899`.

Full current-main verification is green: `make format`, Ruff, mypy, and `make tests` completed with **2,134 passed, 6 skipped, 1 xfailed**. GitHub reports the PR mergeable. The `run.extra.metadata.usage_metadata` fallback remains lowest precedence and independent of `outputs`. Please review the current head.

AXEG0:
Bumping this for review. GitHub currently reports the PR as mergeable. The change is intentionally narrow: `extra.metadata.usage_metadata` is used only as a last-resort OTEL token-usage source, preserving existing top-level/output precedence, with end-to-end exporter tests covering fallback, no-output, precedence, and no-usage cases. Maintainer edits are enabled.

PR #3244, opened 2026-07-21
fix(py): merge sandbox request headers case-insensitively

ramon-langchain:
Merge sandbox request headers case-insensitively so a per-request override replaces a base header that differs only in casing, instead of emitting both.

## Problem

`merge_headers` merges per-request overrides onto the client’s default headers with a plain `dict.update`. The client stores its defaults as `httpx.Headers`, which normalizes names to **lowercase**, so the base dict has e.g. `x-service-key`. When a caller passes a per-request override using a different casing (e.g. Title-Case `X-Service-Key`), `dict.update` treats it as a distinct key and **both** headers are sent:

```
x-service-key: <client default>     # sent first
X-Service-Key: <override>           # sent second
```

HTTP header names are case-insensitive, so this is ambiguous. A server that reads the first value (Go’s `http.Header.Get`, for one) sees the **client default**, not the override — so a caller that overrides an auth header per-request silently gets the wrong one, e.g. an auth rejection.

## Fix

`merge_headers` now drops any base header whose name matches an override name case-insensitively before applying the override, so exactly one header (the override’s) is emitted.

## Test Plan

- [x] `uv run pytest tests/unit_tests/sandbox/test_helpers.py` — new `merge_headers` tests, including a regression that a Title-Case override replaces an `httpx.Headers` (lowercased) base and only one header reaches the wire
- [x] `uv run pytest tests/unit_tests/sandbox/` — full sandbox unit suite (499 passed)

### DIFF
--- a/python/langsmith/sandbox/_helpers.py
+++ b/python/langsmith/sandbox/_helpers.py
@@ -34,9 +34,21 @@ def merge_headers(
     base_headers: Optional[Mapping[str, str]] = None,
     override_headers: Optional[Mapping[str, str]] = None,
 ) -> dict[str, str]:
-    """Merge request headers, giving precedence to overrides."""
+    """Merge request headers, giving precedence to overrides.
+
+    Matching is case-insensitive: an override replaces a base header that
+    differs only in casing, rather than both being sent. HTTP treats header
+    names case-insensitively, so emitting both is ambiguous — a server that
+    reads the first value would see the base header instead of the override.
+    """
     merged: dict[str, str] = dict(base_headers or {})
     if override_headers:
+        overridden = {name.lower() for name in override_headers}
+        merged = {
+            name: value
+            for name, value in merged.items()
+            if name.lower() not in overridden
+        }
         merged.update(override_headers)
     return merged
 
--- a/python/langsmith/sandbox/_helpers.py
+++ b/python/langsmith/sandbox/_helpers.py
@@ -36,20 +36,15 @@ def merge_headers(
 ) -> dict[str, str]:
     """Merge request headers, giving precedence to overrides.
 
-    Matching is case-insensitive: an override replaces a base header that
-    differs only in casing, rather than both being sent. HTTP treats header
-    names case-insensitively, so emitting both is ambiguous — a server that
-    reads the first value would see the base header instead of the override.
+    Names are normalized to lowercase so an override replaces a base header that
+    differs only in casing. HTTP header names are case-insensitive, so keeping
+    both would be ambiguous — a server reading the first value would see the
+    base header instead of the override.
     """
-    merged: dict[str, str] = dict(base_headers or {})
-    if override_headers:
-        overridden = {name.lower() for name in override_headers}
-        merged = {
-            name: value
-            for name, value in merged.items()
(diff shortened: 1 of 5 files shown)


python/tests/unit_tests/sandbox/test_helpers.py
```diff
@@ -0,0 +1,36 @@
...
+    # still replace the base rather than produce a second, duplicate header.
+    base = httpx.Headers({"X-Service-Key": "base", "X-Api-Key": ""})
+    merged = merge_headers(base, {"X-Service-Key": "override"})
+
+    assert [v for k, v in merged.items() if k.lower() == "x-service-key"] == [
+        "override"
+    ]
+    # And only one X-Service-Key actually goes on the wire.
+    request = httpx.Request("GET", "https://example.com", headers=merged)
+    on_wire = [v.decode() for k, v in request.headers.raw if k.lower() == b"x-service-key"]
```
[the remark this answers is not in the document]
ramon-langchain: Fixed — wrapped the comprehension so the line is under 88 chars; `ruff check` + `ruff format --check` now pass on both files locally.

— Claude

wbbradley (approved):

python/langsmith/sandbox/_helpers.py
```diff
@@ -34,9 +34,21 @@ def merge_headers(
...
+    """Merge request headers, giving precedence to overrides.
+
+    Matching is case-insensitive: an override replaces a base header that
+    differs only in casing, rather than both being sent. HTTP treats header
+    names case-insensitively, so emitting both is ambiguous — a server that
+    reads the first value would see the base header instead of the override.
+    """
     merged: dict[str, str] = dict(base_headers or {})
     if override_headers:
+        overridden = {name.lower() for name in override_headers}
```
wbbradley: nit: This is a bit roundabout with the if filter. It could just lowercase everything and overwrite in a loop over `merged` then `override_headers`? So, the end result is always lowercased and there is no if statement or set?

ramon-langchain: Good instinct, but I kept the filter deliberately. `merge_headers` also feeds `buildAuthHeaders` on the WS-exec path (`_ws_execute.py:60`), whose tests assert header names are **preserved** as-is (e.g. `X-Test-Header`, `X-Api-Key`). Lowercasing every key normalizes those and breaks `test_ws_execute.py::TestBuildAuthHeaders` (I tried it — 4 failures). The set-based filter resolves only the casing *collision* (drop a base header that an override replaces under a different casing) while leaving all other names untouched, so it fixes the bug without changing the WS surface. Happy to switch to full lowercasing + update those WS tests if you'd prefer that as the contract.

— Claude

ramon-langchain: Done — switched to lowercasing every name (loop over base then override, no set/if), so the result carries no case-sensitive assumption and the override always wins. Updated the tests that asserted preserved casing on merge outputs (`TestBuildAuthHeaders`, `_ws_default_headers`) to the normalized lowercase contract. Full sandbox unit suite green (499 passed). Pushed in 8c53b329.

— Claude

PR #3249, opened 2026-07-22
fix: add session_id and start_time to feedback generated from evaluat…

agola11:
…ions

Surface area:
* `evaluate` / `aevaluate` (python)
* `evaluate` (JS)
* Pytest / Jest

### DIFF
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -7624,7 +7624,7 @@ class Client:
                     Optional[ls_schemas.FeedbackConfig], res.feedback_config
                 ),
                 feedback_source_type=ls_schemas.FeedbackSourceType.MODEL,
-                project_id=run.session_id if run and run.session_id else project_id,
+                project_id=project_id if run is None else None,
                 extra=res.extra,
                 trace_id=run.trace_id if run else None,
                 session_id=run.session_id if run and run.session_id else project_id,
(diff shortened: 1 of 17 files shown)


python/langsmith/client.py
```diff
@@ -7570,6 +7570,7 @@ def _log_evaluation_feedback(
         source_info: Optional[dict[str, Any]] = None,
         project_id: Optional[ID_TYPE] = None,
         *,
+        session_id: Optional[ID_TYPE] = None,
```
KiewanVillatel: We already have a `project_id` parameter, isn't it the same?

python/langsmith/client.py
```diff
@@ -7607,7 +7608,7 @@ def _submit_feedback(**kwargs):
                 project_id=project_id,
                 extra=res.extra,
                 trace_id=run.trace_id if run else None,
-                session_id=run.session_id if run else None,
+                session_id=run.session_id if run and run.session_id else session_id,
```
KiewanVillatel: Should we raise a warning in case both are None? The API allows it, not sure if that would / could happen in practive.

KiewanVillatel (approved):

KiewanVillatel (commented):
Two small comments / questions

python/langsmith/client.py
agola11: @jacoblee93 would defer to you on this. Don't have much context here

jacoblee93: Yes let's just reuse this. Will update.

python/langsmith/client.py
jacoblee93: Let's do in a followup along with more general migration hints

PR #3260, opened 2026-07-23
fix(js): Ensure dataset version for an experiment is after all examples

jacoblee93:
On the first execution of a test case, `/v2/datasets/{dataset_id}/experiment-runs` could return no results even though `/v2/traces/query` returned the corresponding root run.

The SmithDB experiment-runs endpoint reconstructs dataset examples as of the experiment’s `dataset_version`. When that metadata is absent, it falls back to the experiment start time. Because Jest/Vitest creates new examples after starting the experiment, those examples fall outside that snapshot, causing their runs to be omitted.

Subsequent executions generally work because the deterministic example already exists before the new experiment starts.

## Fix

Include the final example modification time as `dataset_version` when updating Jest/Vitest experiment metadata:

```ts
dataset_version: finalModifiedAt
```

This ensures the experiment-runs endpoint uses a dataset snapshot that includes examples created during the test run.

### DIFF
--- a/js/src/utils/jestlike/index.ts
+++ b/js/src/utils/jestlike/index.ts
@@ -434,6 +434,7 @@ export function generateWrapperFromJestlikeMethods(
             await client.updateProject(datasetInfo.project.id, {
               metadata: {
                 ...suiteMetadata,
+                dataset_version: finalModifiedAt,
                 commit,
                 branch,
                 dirty,
--- a/js/src/utils/jestlike/index.ts
+++ b/js/src/utils/jestlike/index.ts
@@ -23,7 +23,6 @@ import {
   TestWrapperAsyncLocalStorageData,
   testWrapperAsyncLocalStorageInstance,
   _logTestFeedback,
-  syncExamplePromises,
   trackingEnabled,
   DEFAULT_TEST_CLIENT,
 } from "./globals.js";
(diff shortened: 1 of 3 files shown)


js/src/utils/jestlike/index.ts
```diff
@@ -672,21 +674,25 @@ export function generateWrapperFromJestlikeMethods(
...
-                      datasetId: dataset.id,
-                      inputs,
-                      outputs: referenceOutputs ?? {},
-                      metadata,
-                      split,
-                      createdAt,
-                    }),
+                if (context.syncExamplePromises === undefined) {
+                  throw new Error(
+                    "Could not identify suite example state. Please contact us for help.",
```
KiewanVillatel: Could we rephrase a bit the error message?

> Could not identify suite example state.

Not sure this part would be understandable for a user.

> Please contact us for help.

Who is us? :-) 

Should we be more specific and ask the user to create a ticket?
```suggestion
                    "Internal error: syncExamplePromises not initialized. This is a bug in the LangSmith SDK — please open an issue at https://github.com/langchain-ai/langsmith-sdk/issues.",
```

QuentinBrosse (approved):

KiewanVillatel (approved):

js/src/utils/jestlike/index.ts
jacoblee93: Nice yea

PR #3262, opened 2026-07-23
feat: migrate internal list_runs/read_run usages to SmithDB v2 API

KiewanVillatel:
## Summary

Migrates the Python SDK's single-run read path to the SmithDB v2 API for backends where ClickHouse query support is disabled, while keeping the legacy v1 path as the default everywhere else. JS is untouched in this PR (see below).

## Key changes

**`python/langsmith/_internal/_v2_migration_utils.py`**
- New `QueryBackend` enum (`CLICKHOUSE_ONLY` / `SMITHDB_ONLY` / `DUAL`) and `get_query_backend(instance_flags)` — the single place that interprets the `ch_query_enabled` / `sdb_query_enabled` instance flags from `/info`. All call sites below use this instead of reading the flags inline.
- New `_read_run_v2()` helper: fetches a single run via `runs.retrieve` (v2), converting the response to `schemas.Run` via the existing `_v2_run_to_schema`.

**`python/langsmith/client.py`**
- `read_run(run_id, load_child_runs=False, *, project_id=None, start_time=None)`: v1 `GET /runs/{id}` is always tried first/by default. Only falls back to the v2 `runs.retrieve` path when the backend is SmithDB-only (`ch_query_enabled=False`, `sdb_query_enabled=True`), in which case `project_id` and `start_time` are both required (raises `LangSmithError` if missing).
- `load_child_runs` is deprecated: passing it (`read_run`, `evaluate_run`, `aevaluate_run`) emits a `DeprecationWarning`, and raises on SmithDB-only backends since child-run loading isn't supported there.
- `_load_child_runs` no longer attempts a v2 fallback — it raises `LangSmithError` outright on SmithDB-only backends.
- `evaluate_run` / `aevaluate_run` gain optional `project_id` and `start_time` params (forwarded to `read_run` when resolving a bare run ID), and now accept the v2 `Run` datamodel directly in addition to `schemas.Run` / `RunBase` / `str` / `UUID`. Resolving a run from an ID without `project_id` warns (deprecated) on normal backends and raises on SmithDB-only ones.

**`python/langsmith/async_client.py`**
- `read_run` gets the same v1-priority / SmithDB-only v2 fallback via `get_query_backend`, requiring `project_id` on the v2 path (no `start_time`/`load_child_runs` support here, matching its existing narrower scope).

**`python/langsmith/evaluation/evaluator.py`, `string_evaluator.py`**
- Widened `RunEvaluator.evaluate_run` / `aevaluate_run` (and `DynamicRunEvaluator`'s overrides, `StringEvaluator.evaluate_run`) to accept `Union[schemas.Run, RunBase, v2 Run]`, since `Client.evaluate_run` can now hand evaluators a v2 `Run` object directly. This is a real contract widening, not a cast — callers passing a v2 `Run` are responsible for using an evaluator that doesn't depend on legacy-only fields (e.g. `.latency`, `.input_tokens`, lowercase `run_type`).

**`python/langsmith/beta/_evals.py`, `python/langsmith/evaluation/_runner.py`**
- Nested-trace loaders (`_load_nested_traces` / `_load_traces_v2`) keep their original, looser routing — use the v2 API whenever SmithDB is available at all (`get_query_backend() != CLICKHOUSE_ONLY`), not just when it's the only option — unchanged in behavior, just re-expressed via `get_query_backend`.


js/src/client.ts
```diff
@@ -2924,13 +2925,31 @@ export class Client implements LangSmithTracingClientInterface {
...
+    if (await this._supportsSDBQuery()) {
+      const fetched: Run[] = [];
+      const pager = this.runs.queryV2({
+        project_ids: [run.session_id!],
+        is_root: false,
+        trace_id: run.trace_id ?? undefined,
+        min_start_time: run.start_time
+          ? new Date(run.start_time).toISOString()
+          : undefined,
+        selects: V2_RUN_SELECTS,
```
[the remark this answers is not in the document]
KiewanVillatel: Addressed in 8e4a3f85 — expanded V2_RUN_SELECTS and _V2_RUN_SELECTS to include TAGS, EXTRA, EVENTS, FEEDBACK_STATS, FIRST_TOKEN_TIME, APP_PATH, and token/cost fields; updated v2RunToRun and _v2_run_to_schema to map them.

python/langsmith/_internal/_v2_migration_utils.py
```diff
@@ -82,3 +84,82 @@ def _v2_run_to_schema(run: Any) -> schemas.Run:
...
+    ``schemas.Run`` objects and tree-building with ``parent_run_id`` works.
+    """
+    pager = client._get_langsmith_api_sync().runs.query_v2(
+        project_ids=[str(run.session_id)],
+        is_root=False,
+        trace_id=str(run.trace_id),
+        min_start_time=run.start_time,
+        selects=_V2_RUN_SELECTS,
+    )
+    child_runs = [_v2_run_to_schema(r) for r in pager]
```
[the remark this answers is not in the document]
KiewanVillatel: Addressed in 8e4a3f85 — _v2_run_to_schema now calls run.run_type.lower(), matching what JS v2RunToRun already did.

python/tests/unit_tests/test_v2_migration_utils.py
```diff
@@ -0,0 +1,213 @@
...
+# ---------------------------------------------------------------------------
+# parent_run_id
+# ---------------------------------------------------------------------------
+
+
+def test_parent_run_id_set_to_last_ancestor():
+    grandparent = str(uuid.uuid4())
+    parent = str(uuid.uuid4())
+    result = _v2_run_to_schema(_make_v2_run(parent_run_ids=[grandparent, parent]))
+    assert result.parent_run_id == parent
```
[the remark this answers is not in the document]
KiewanVillatel: Addressed in 6cfe6fdd — changed assertion to `str(result.parent_run_id) == parent` since `schemas.Run` coerces the field to `uuid.UUID` on construction.

python/tests/integration_tests/test_async_client.py
```diff
@@ -36,26 +36,29 @@ async def async_client():
...
-            run = await async_client.read_run(run_id)
+            run = await async_client.runs.retrieve(run_id, project_id=project_id, selects=["NAME"], start_time=start_time)
             return run.name == "test_run"
         except ls_utils.LangSmithError:
             return False
 
     await wait_for(check_run)
-    run = await async_client.read_run(run_id)
+
+    run = await async_client.runs.retrieve(run_id, project_id=project_id, selects=["NAME","INPUTS"], start_time=start_time)
```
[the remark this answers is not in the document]
KiewanVillatel: This is enforced automatically by the linter (auto-fixed in 9ddd8bce). Resolving.

python/langsmith/client.py
```diff
@@ -7567,20 +7678,27 @@ def evaluate_run(
...
+        run_ = self._resolve_run_id(
+            run,
+            load_child_runs=load_child_runs,
+            project_id=project_id,
+            start_time=start_time,
+        )
         reference_example_ = self._resolve_example_id(reference_example, run_)
         evaluator_response = evaluator.evaluate_run(
-            run_,
+            cast(ls_schemas.Run, run_),
```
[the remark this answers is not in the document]
KiewanVillatel: Addressed in bb673844 — widened RunEvaluator.evaluate_run/aevaluate_run (base interface), DynamicRunEvaluator's overrides, and StringEvaluator.evaluate_run to accept Union[Run, RunBase, V2Run] instead of just Run, and removed the cast(ls_schemas.Run, run_) calls in client.py's evaluate_run/aevaluate_run that previously lied about the runtime type. This is now a real, typed contract widening rather than a cast — an evaluator that accesses legacy-only fields (.latency, .input_tokens, lowercase run_type) will still raise if handed a v2 Run, but that's now an explicit part of the contract callers opt into by passing a v2 Run, not a hidden runtime mismatch.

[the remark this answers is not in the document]
KiewanVillatel: Reverted in ea2edfd4 — after further review we decided not to widen the RunEvaluator contract (this comment's follow-up about DynamicRunEvaluator's stored callable type not being widened either was a fair catch, and fixing that fully would have meant threading the v2 Run type through the whole evaluator stack). evaluate_run/aevaluate_run are back to cast(ls_schemas.Run, run_) before invoking the evaluator — the original runtime-compatibility gap you flagged is real and still present: an evaluator that touches legacy-only fields (.latency, .input_tokens, lowercase run_type) will still break if handed a v2 Run. Passing a v2 Run into evaluate_run/aevaluate_run remains something callers need to do carefully, with an evaluator known to be compatible with it.

python/langsmith/beta/_evals.py
```diff
@@ -175,6 +173,13 @@ def convert_runs_to_test(
 
 
 def _load_nested_traces(project_name: str, client: Client) -> list[ls_schemas.Run]:
+    from langsmith._internal import _v2_migration_utils
```
QuentinBrosse: nit: move to the top of the file?

python/langsmith/client.py
```diff
@@ -4043,18 +4070,50 @@ def read_run(
...
         if load_child_runs:
-            run = self._load_child_runs(run)
+            warnings.warn(
+                "load_child_runs is deprecated and will be removed in a"
+                " future release.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+        run_id_ = _as_uuid(run_id, "run_id")
+        from langsmith._internal import _v2_migration_utils
```
QuentinBrosse: Move to the top of the file?

python/langsmith/client.py
```diff
@@ -4043,18 +4070,50 @@ def read_run(
...
-            response.json(), attachments_key="s3_urls", api_url=self.api_url
-        )
-        run = ls_schemas.Run(
-            attachments=attachments, **response.json(), _host_url=self._host_url
-        )
-
         if load_child_runs:
-            run = self._load_child_runs(run)
+            warnings.warn(
+                "load_child_runs is deprecated and will be removed in a"
```
QuentinBrosse: Should we add an example in https://docs.langchain.com/langsmith/smithdb-sdk-migration#examples-2 to explain what the user should doinstead?

python/langsmith/client.py
```diff
@@ -4019,15 +4031,30 @@ def _load_child_runs(self, run: ls_schemas.Run) -> ls_schemas.Run:
         return run
 
     def read_run(
-        self, run_id: ID_TYPE, load_child_runs: bool = False
+        self,
```
QuentinBrosse, line 4034: I don't understand why we fix this method. I thought the plan was to entirely deprecate it (later this week) and remove it in v1. This method will basically never be compatible with Smith DB-only.

Should we fix the caller instead?

KiewanVillatel: As discussed, I removed the deprecated warning on `load_child_runs` (will be covered by the global deprecation on the whole method)

python/langsmith/client.py
KiewanVillatel: Yes will do in a separate doc PR.

python/langsmith/async_client.py
```diff
@@ -580,13 +580,48 @@ async def update_run(
...
+
+        Args:
+            run_id: The ID of the run to read.
+            project_id: The ID of the project (session) that owns the run.
+                Required on SmithDB-only backends (no ClickHouse query
+                support), where it's used to look up the run via the v2 API.
+        """
+        run_id_ = ls_client._as_uuid(run_id)
+        info = await self.info()
+        from langsmith._internal._v2_migration_utils import (
```
QuentinBrosse: move import to the top?

python/langsmith/client.py
```diff
@@ -4019,15 +4029,30 @@ def _load_child_runs(self, run: ls_schemas.Run) -> ls_schemas.Run:
...
+        run_id: ID_TYPE,
+        load_child_runs: bool = False,
+        *,
+        project_id: Optional[ID_TYPE] = None,
+        start_time: Optional[datetime.datetime] = None,
```
QuentinBrosse, line 4039: Do we also need the `start_time` in the `AsyncClient.read_run`?

python/langsmith/client.py
```diff
@@ -4043,18 +4068,41 @@ def read_run(
...
+            response = self.request_with_retries("GET", f"/runs/{run_id_}")
+            attachments = _convert_stored_attachments_to_attachments_dict(
+                response.json(), attachments_key="s3_urls", api_url=self.api_url
+            )
+            run = ls_schemas.Run(
+                attachments=attachments, **response.json(), _host_url=self._host_url
+            )
+            if load_child_runs:
+                run = self._load_child_runs(run)
+        else:
```
QuentinBrosse: very nit: we could return the run here and remove the else for readability

python/langsmith/client.py
KiewanVillatel: For async client and JS it's not necessary (not used by other internal methods)

python/langsmith/async_client.py
KiewanVillatel: Done in 2ceb0175 — moved to the module-level import block (no circular-import risk: `_v2_migration_utils` only imports `langsmith.schemas` and the generated client at module scope).

python/langsmith/client.py
KiewanVillatel: Done in 2ceb0175 — the legacy path now returns early and the SmithDB-only branch is dedented, so the `else` is gone.

python/langsmith/client.py
```diff
@@ -4019,15 +4029,30 @@ def _load_child_runs(self, run: ls_schemas.Run) -> ls_schemas.Run:
...
             load_child_runs (bool, default=False):
-                Whether to load nested child runs.
+                Whether to load nested child runs. **Deprecated**: this will
+                be removed in a future release, and raises on SmithDB-only
+                backends (no ClickHouse query support), where child runs are
+                never loaded.
+            project_id (Optional[Union[UUID, str]]):
+                The ID of the project (session) that owns the run. Required on
+                SmithDB-only backends (no ClickHouse query support), where it's
+                used to look up the run via the v2 API.
```
QuentinBrosse:
```suggestion
                SmithDB-only backends (no ClickHouse query support).
```

python/langsmith/client.py
```diff
@@ -4019,15 +4029,30 @@ def _load_child_runs(self, run: ls_schemas.Run) -> ls_schemas.Run:
...
+            start_time (Optional[datetime]):
+                The run's start time. Required on SmithDB-only backends (no
+                ClickHouse query support), where it speeds up the v2 lookup.
```
QuentinBrosse: It's not clear whether it's required in SmithDB or optional just to speed things up? I think it's required right?

python/langsmith/client.py
```diff
@@ -5127,20 +5174,24 @@ def get_test_results(
 
         import pandas as pd  # type: ignore
 
-        runs = self.list_runs(
-            project_id=project_id,
-            project_name=project_name,
+        if project_id is None and project_name is not None:
+            project_id = self.read_project(project_name=project_name).id
+        if project_id is None:
+            raise ValueError("Either project_id or project_name must be provided.")
```
QuentinBrosse: Will I also receive this message if I send a wrong `project_name`?

python/langsmith/client.py
```diff
@@ -5127,20 +5174,24 @@ def get_test_results(
...
+        runs = self._get_langsmith_api_sync().runs.query(
```
QuentinBrosse: Should we use the old endpoint on ClickHouse-only envs?

python/langsmith/client.py
```diff
@@ -7604,6 +7708,13 @@ def _submit_feedback(**kwargs):
             else:
                 self.create_feedback(**kwargs)
 
+        # `session_id` on the legacy `schemas.Run`; `project_id` on the v2 `Run`.
+        run_session_id = (
+            (getattr(run, "session_id", None) or getattr(run, "project_id", None))
```
QuentinBrosse, line 7744: Ah yeah, I see `RunBase` doesn't even have the `session_id` field :(

python/langsmith/_internal/_v2_migration_utils.py
```diff
@@ -2,17 +2,49 @@
...
 
+class QueryBackend(enum.Enum):
+    """Which backend(s) a LangSmith instance can serve run/trace queries from."""
+
+    CLICKHOUSE_ONLY = "clickhouse_only"
+    SMITHDB_ONLY = "smithdb_only"
+    DUAL = "dual"
+
+
+def get_query_backend(
```
QuentinBrosse, line 28: Should we add a small unit test now that we extensively use this function?

python/langsmith/client.py
KiewanVillatel: Indeed it's not clear, let me rephrase. Yes it's required.

python/langsmith/client.py
KiewanVillatel: No, an exception will be raised.

python/langsmith/client.py
KiewanVillatel: Yes, good point.

python/langsmith/_internal/_v2_migration_utils.py
KiewanVillatel: Added in 32bc24f7 — `test_get_query_backend` covers the full `ch_query_enabled` / `sdb_query_enabled` matrix, including the absent-flag defaults (`None`/`{}` → CLICKHOUSE_ONLY, since `ch_query_enabled` defaults to enabled for older backends and `sdb_query_enabled` to disabled), the both-disabled fallback, and a second case asserting the `bool()` coercion of non-bool flag values.

python/langsmith/client.py
KiewanVillatel: Rephrased in 32bc24f7 — dropped the "where it speeds up the v2 lookup" wording; it now just says required on SmithDB-only backends, which matches the `LangSmithError` the code raises when it is missing.

python/langsmith/client.py
KiewanVillatel: Done in 32bc24f7 — `get_test_results` now branches on `get_query_backend`: ClickHouse-only *and* dual keep the original `self.list_runs(...)` call with the lowercase `select` fields untouched, and only SmithDB-only goes through v2 `runs.query`. The `project_name` → `project_id` resolution via `read_project` moved inside the v2 branch, so the legacy path still passes `project_name` straight through to `list_runs` as before.

One wrinkle worth flagging: the two backends return different `feedback_stats` shapes — legacy `schemas.Run` gives a dict-of-dicts (`v["avg"]`) while the v2 model gives objects (`v.avg`). Rather than duplicate the row-building loop I added a small `stat_field(stat, field)` helper that reads either form. Verified both branches produce identical dataframes (same `feedback.*` columns, `notes`, and `execution_time`).

python/langsmith/async_client.py
```diff
@@ -580,13 +586,41 @@ async def update_run(
...
-    async def read_run(self, run_id: ls_client.ID_TYPE) -> ls_schemas.Run:
-        """Read a run."""
-        response = await self._arequest_with_retries(
-            "GET",
-            f"/runs/{ls_client._as_uuid(run_id)}",
+    async def read_run(
+        self,
+        run_id: ls_client.ID_TYPE,
+        *,
+        project_id: Optional[ls_client.ID_TYPE] = None,
```
QuentinBrosse, line 593: Do we need a start_time here?

python/langsmith/client.py
```diff
@@ -3986,8 +3987,17 @@ def _load_child_runs(self, run: ls_schemas.Run) -> ls_schemas.Run:
...
-            LangSmithError: If a child run has no parent.
+            LangSmithError: If a child run has no parent, or on SmithDB-only
+                backends (no ClickHouse query support), where loading child
+                runs isn't supported.
         """
+        backend = _v2_migration_utils.get_query_backend(self.info.instance_flags)
+        if backend == _v2_migration_utils.QueryBackend.SMITHDB_ONLY:
+            raise ls_utils.LangSmithError(
+                "Loading child runs is not supported on SmithDB-only"
+                " backends (no ClickHouse query support)."
```
QuentinBrosse: I think we should add an alternative in the migration guide and link it here.

python/langsmith/client.py
```diff
@@ -4043,19 +4067,41 @@ def read_run(
...
+        if start_time is None:
+            raise ls_utils.LangSmithError(
+                "read_run requires start_time on SmithDB-only backends"
+                " (no ClickHouse query support)."
+            )
         if load_child_runs:
-            run = self._load_child_runs(run)
-        return run
+            raise ls_utils.LangSmithError(
+                "load_child_runs is not supported on SmithDB-only"
```
QuentinBrosse, line 4098: same https://github.com/langchain-ai/langsmith-sdk/pull/3262/changes#r3672673779

python/langsmith/async_client.py
KiewanVillatel: No in the async client it's not needed, it's not used internally. So to use SmithDB users would need to switch to the new runs.retrieve method as per the migration guide.

python/langsmith/client.py
QuentinBrosse: You can use this migration guide path: `langsmith/smithdb-sdk-migration#load-a-run’s-child-runs`

PR that added the example: https://github.com/langchain-ai/docs/pull/5176

QuentinBrosse (approved):

PR #3274, opened 2026-07-27
feat: require sessionID for feedback when ClickHouse unavailable

emil-lc:
## Summary

On SmithDB-only deployments (ClickHouse queries disabled), the server cannot associate
run-level feedback with its run unless the client supplies `session_id` — `POST /feedback`
returns **HTTP 501** pointing at the migration guide. Both SDKs let `session_id` default to
`None`, so callers only discover this at runtime, on the deployment that lacks ClickHouse.

This adds a client-side check to `create_feedback` / `createFeedback`: a hard error when the
backend advertises SmithDB-only, and a deprecation warning everywhere else, so the requirement
surfaces before the request is sent and ahead of the future removal.

## Changes

- **Capability-gated check.** When feedback has a `run_id` but no `session_id`, the SDKs read
  the `/info` flag `instance_flags["ch_query_enabled"]`. When explicitly `false`, they raise
  (`ValueError` / `Error`) without sending the request; otherwise they emit a one-time
  deprecation warning (`LangSmithWarning` / `warnOnce`) linking to
  [the migration guide](https://docs.langchain.com/langsmith/smithdb-sdk-migration#feedback-create).
  `/info` is only consulted when `session_id` is actually missing, so the common path adds no
  request. `/info` failures degrade to the warning, never the error.
- **Session-level feedback is exempt.** Feedback created with `project_id` and no run has no run
  to locate, and is unaffected.
- **JS params-object overload.** `createFeedback(params)` now accepts a single object in which
  `sessionId` is **required** by the type, so the constraint is enforced at compile time rather
  than only at runtime. The positional `createFeedback(runId, key, options)` signature is kept
  and marked `@deprecated`, following the existing `logEvaluationFeedback` convention. The params
  type is a discriminated union — `{ runId, sessionId, … }` or `{ projectId, … }` — so
  session-level feedback is still expressible.
- **Docs.** `session_id` / `sessionId` is documented as required for run-level feedback with the
  migration link; `start_time` is described as a lookup-skipping optimization.

## Compatibility

- No behavior change on ClickHouse-enabled backends beyond a one-time deprecation warning.
- Matches the server's own gate exactly: only `session_id` is required. `start_time` stays
  optional.
- The deprecated JS positional signature still compiles unchanged, including the no-op `eager`
  option, which is retained on the legacy options type for source compatibility.
- Internal SDK callers (per-run, comparative, and summary evaluators; the pytest integration)
  already pass a session and are unaffected.

## Testing

- Python unit tests, sync and async: raises without `session_id` when `ch_query_enabled: false`
  and sends nothing; accepts `session_id` without `start_time`; warns on other backends;
  session-level feedback exempt.
- JS unit tests for the same matrix, plus a `@ts-expect-error` assertion that a run-level
  params-object call without `sessionId` fails to compile (`tsconfig` covers `src/tests`, so CI's
  `tsc` enforces it).

### DIFF
--- a/js/src/client.ts
+++ b/js/src/client.ts
@@ -549,6 +549,56 @@ type RecordStringAny = Record<string, any>;
 
 export type FeedbackSourceType = "model" | "api" | "app";
 
+export type CreateFeedbackOptions = {
+  /** The metric name, tag, or aspect to provide feedback on. */
+  key: string;
+  score?: ScoreType;
+  value?: ValueType;
+  correction?: object;
+  comment?: string;
+  sourceInfo?: object;
+  feedbackSourceType?: FeedbackSourceType;
+  feedbackConfig?: FeedbackConfig;
+  sourceRunId?: string;
+  feedbackId?: string;
+  comparativeExperimentId?: string;
+  /**
+   * The run's start time, ISO string or epoch ms. Better performance if provided.
+   */
+  startTime?: number | string;
+  /** If false, create feedback without extending the trace's retention tier. */
+  extendTraceRetention?: boolean;
+};
+
+/** @deprecated Pass all params within an object and populate sessionId. */
+export type CreateFeedbackLegacyOptions = Omit<CreateFeedbackOptions, "key"> & {
+  /**
+   * The session (project) ID of the run. Required for run-level feedback;
+   * omitting it is deprecated. See
+   * https://docs.langchain.com/langsmith/smithdb-sdk-migration#feedback-create
+   */
+  sessionId?: string;
+  /** The project (or experiment) to provide feedback on, for session-level feedback. */
+  projectId?: string;
+};
+
+export type CreateFeedbackParams = CreateFeedbackOptions &
+  (
+    | {
+        /** The run to provide feedback on. */
+        runId: string;
+        /** The session (project) ID of the run. */
+        sessionId: string;
+        projectId?: never;
+      }
+    | {
+        runId?: null;
+        /** The project (or experiment) to provide feedback on. */
+        projectId: string;
+        sessionId?: never;
+      }
+  );
+
 export type CreateExampleOptions = {
   /** The ID of the dataset to create the example in. */
   datasetId?: string;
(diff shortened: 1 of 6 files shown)


js/src/client.ts
jacoblee93, line 5140: Let's add an overload here to further assist people:

```ts
public async createFeedback(
   params: { runId: string | null, sessionId: string, ...others }
);
/** @deprecated Pass all parameters in an object */
public async createFeedback(
    runId: string | null,
    key: string,
    {
      score,
      value,
      correction,
      comment,
      sourceInfo,
...
    }
);
public async createFeedback(
   runIdOrParams: { runId: string | null, sessionId: string } string | null
   ...
);
```

python/langsmith/client.py
```diff
@@ -7773,11 +7784,11 @@ def create_feedback(
             extra (Optional[Dict]):
                 Metadata for the feedback.
             session_id (Optional[Union[UUID, str]]):
```
jacoblee93, line 7786: Dunno if we can do something similar to the above

js/src/client.ts
emil-lc: Updated with an overload. Please take another look.

jacoblee93: I noticed in v2 runs we have it as `project_id` - are we standardizing around that?

I guess we can make the change in v2 just would be good to have that as a firm action item

jacoblee93 (approved):

python/langsmith/client.py
emil-lc: Not worth it. We'd need to repeat every param of the ~30-line func signature for the overloaded copies. Using overloads is not much of a win in Py-land, esp. in this case.

js/src/client.ts
emil-lc: `project_id` has different semantics - it is the parameter used when one sends session-level feedback, and in _that case_ RunID must not be set. It's confusing, and yeah, we'd need to clean this up with a V2 API eventually.

python/langsmith/client.py
```diff
@@ -686,6 +678,25 @@ def _format_feedback_score(score: Union[float, int, bool, None]):
     return score
 
 
+def _check_feedback_session_id(info: ls_schemas.LangSmithInfo) -> None:
+    """Raise on SmithDB-only deployments, warn elsewhere.
+
+    Call only when run-level feedback has no ``session_id``.
+    """
+    docs = "https://docs.langchain.com/langsmith/smithdb-sdk-migration#feedback-create"
+    if (info.instance_flags or {}).get("ch_query_enabled") is False:
```
QuentinBrosse, line 687: We have this util now if you want to reuse: https://github.com/langchain-ai/langsmith-sdk/blob/c443de705564da573405b22062cd3dce5f345486/python/langsmith/_internal/_v2_migration_utils.py#L28 (can be done in a fast-follow)

Same in TypeScript.

emil-lc: Thanks - it's a nice internal optimization, but I'll leave this to another PR

PR #3275, opened 2026-07-28
ci: test Python integration and evals on self-hosted environments

KiewanVillatel:
## Description

The Python and JS integration and eval suites only ever ran against beta, so nothing verified that the SmithDB-backed SDK methods work on a self-hosted deployment. This adds environment selection to CI and fixes the tests that couldn't pass anywhere but beta.

Closes LIN-470

**Environment selection**

- New `integration-environment` choice input on the CI `workflow_dispatch`: `beta` (default), `v15`, `v16-dual`, `v16-smithdb`, or `all`. It drives both the Python and the JS suites.
- An `integration-matrix` job builds two matrices — `python_matrix` (6 integration shards + evals per environment, doctests once on beta) and `js_matrix` (one integration job and one eval job per environment) — expanding the test types across the selected environments. Unrecognized input falls back to `beta`, and the dispatch input is passed to the matrix script via an env var rather than interpolated into the shell.
- Pushes and pull requests behave exactly as before — beta only, and the beta job names are byte-identical so required status checks keep matching.
- All three composite actions (Python integration, JS integration, JS vitest eval) take `langsmith-endpoint` / `langsmith-api-key` instead of hardcoding beta, and fail fast with a clear error when no key is configured for the target environment.
- The JS suite mirrors the pytest markers through `src/tests/utils/markers.ts`: CI passes the markers a deployment can't serve in `LANGSMITH_TEST_EXCLUDE_MARKERS` and `requires(...)` swaps `test`/`describe` for their `.skip` variants (jest has no `describe.skipIf`). `require_v2` skips 1 test on `v15`, `require_clickhouse` skips 19 on `v16-smithdb`, and `require_beta_dataset` skips 2 tests pinned to a beta-only share token on every self-hosted environment.

**Full failure visibility**

- The `-x` in the Makefile recipes meant each CI shard aborted at its first failure, so a red environment only reported one broken test. New `PYTEST_EXIT_FIRST` variable defaults to `-x`, and a `python-integration-exit-first` dispatch input can clear it so a shard reports every failure. Push and PR runs keep `-x`.
- Symmetric knob on the JS side: a `js-integration-exit-first` dispatch input adds `--bail` to the jest integration run and `--bail=1` to the vitest eval run. It defaults to off, so push and PR runs keep reporting every failure exactly as before.

**Per-environment test selection**

- `require_clickhouse` marks the 35 tests that call a legacy pre-SmithDB method (`list_runs`, `read_run`, `list_threads`, `read_thread`, `get_experiment_results`) directly or through a file-local helper. Deselected on `v16-smithdb`.
- `require_v2` marks the 13 tests that drive the autogenerated OpenAPI client (`client.runs/traces/threads/public/datasets/evaluators`). Deselected on `v15`, which doesn't serve those endpoints.
- The two marker sets are disjoint, so neither environment loses coverage the other keeps. Exclusions are delivered through `PYTEST_ADDOPTS` from a per-environment map in the matrix builder.

**JS test harness fixes**

- `getAssumedTreeFromCalls` compared request pathnames to the literal `/runs` and `/api/v1/runs`, so every deployment whose API base carries an `/api` prefix produced `/api/runs`, matched nothing and returned an empty tree. That accounted for 26 of the 36 non-beta failures across the telemetry, claude-agent-sdk and wrapped-openai suites; it now matches the endpoint suffix.
- `pollRunsUntilCount` waited for ingestion through `listRuns`, which SmithDB-only backends don't serve, so the v2 suites (threads, traces, public runs, run sharing) timed out in setup before reaching their assertions. It now counts through `client.runs.query` and falls back to `listRuns` for deployments without the v2 runs endpoints — the same migration the Python v2 fixtures got.
- Declares the `anthropic-api-key` input the JS action referenced but never declared, which GitHub warned about on every run.

**Test fixes**

- The v2 fixtures polled for indexing with `list_runs` and read the trace id with `read_run`, so the v2 suite needed ClickHouse just to set itself up. Migrated to `runs.query` / `runs.retrieve` per the [SmithDB migration guide](https://docs.langchain.com/langsmith/smithdb-sdk-migration); three fixtures became async generators.
- Three eval tests cloned a public dataset from the prod app and then read `"Evaluate Examples"` at `as_of="test_version"` — a tag that only exists on the prod copy, so the name resolved only because that dataset already existed on the beta tenant. They now seed their own dataset in a module-scoped fixture and delete it afterwards. Version tagging is best-effort with a 30s retry, since some deployments 404 on `/datasets/{id}/version` for a freshly seeded dataset.
- `create_feedback` in `test_list_feedback` now passes `session_id`, which the SmithDB-backed endpoint requires.

**Out of scope of this PR**
- Migrating ingestion integration tests to new methods


.github/workflows/ci.yml
```diff
@@ -514,34 +528,71 @@ jobs:
...
+              # Doctests don't exercise the backend, so run them once against beta.
+              if environment == "beta":
+                  include.append({**base, "test-type": "doctest", "shard_total": "", "shard_index": ""})
+
+          print("matrix=" + json.dumps({"include": include}))
+          EOF
+
+  python-integration:
+    name: "Python ${{ matrix.test-type }} Test ${{ matrix.shard_index }}${{ matrix.name_suffix }}"
+    needs: [integration-changes, python-integration-matrix]
```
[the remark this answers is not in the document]
KiewanVillatel: Good catch — fixed in 06c84112 by adding `python-integration-matrix` to `ci-success`'s `needs`, so a matrix-build failure now propagates into `needs.*.result` instead of being invisible behind skipped shards.

.github/workflows/ci.yml
```diff
@@ -503,9 +530,103 @@ jobs:
...
+          SELECTED_ENVIRONMENT: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.integration-environment || 'beta' }}
+        run: |
+          python3 <<'EOF' >> "$GITHUB_OUTPUT"
+          import json
+          import os
+
+          # Allowlist: anything unrecognized falls back to beta.
+          ENDPOINTS = {
+              "beta": "https://beta.api.smith.langchain.com",
+              "v15": "https://aks.smith.langchain.dev/api",
```
QuentinBrosse: It's now a `v16-ch-only` env.
I can ask if we can have a v15 env, but I think we'll need to run v15 tests manually for now.

QuentinBrosse (approved):

.github/workflows/ci.yml
KiewanVillatel: Renamed

PR #3287, opened 2026-07-29
feat(js,python): check min backend version per resource

QuentinBrosse:
## What

The global `_MIN_BACKEND_VERSION` was auto-bumped to the latest backend release, so it warned self-hosted users whose backend was fine for the methods they call. Each resource accessor now declares its own `min_version` (`0.16.0` today), so only users touching a v2 resource can see the warning.

Supersedes #3276, which drops the check entirely.

## Python

- `_check_backend_version(backend_version, *, min_version)` — no constant, returns early when `/info` reported no version.
- The 7 `Client` properties pass `min_version="0.16.0"`.
- `AsyncClient` gets the same per-resource check. Its properties are sync while `info()` is async, so `_schedule_backend_version_check` runs it as a background task, once per version, mirroring the JS client. The unconditional check in `__aenter__` — which warned on every context-manager entry — and the eager `/info` fetch it needed are gone.

## JS

- Deleted `js/src/utils/constants.ts`; `minVersion` is now required.
- `_stainlessVersionChecked` boolean → `Set<string>`, so each distinct min version warns once.
- The `public` getter now checks too; it previously did not, unlike Python.
- Empty-version guard moved into `_checkBackendVersion` to match Python.

Both warnings link to https://docs.langchain.com/langsmith/smithdb-sdk-migration, which documents the `>=0.16` self-hosted requirement.

## Tests

Same comparator matrix in Python and JS against `0.16.0` (older minor, pre-release below min, equal, newer patch, pre-release above min, newer minor, empty, unparseable), plus a wiring test per client proving its accessors call the check — nothing covered that before. Verified by mutation: deleting a call site, or lowering its declared min to `0.15.0`, fails all three suites.

## Follow-ups

- `langchainplus/.github/workflows/bump_sdk_min_backend_version.yaml` must be deleted — it seds the constant that no longer exists.
- August: `min/max_self_hosted_version` annotations in the OpenAPI spec, CI-enforced, generated into methods by `stlc`, replacing these manual per-resource checks.


js/src/client.ts
```diff
@@ -1534,41 +1488,42 @@ export class Client implements LangSmithTracingClientInterface {
   }
 
   public get evaluators(): Evaluators {
-    this._checkStainlessVersion();
+    this._checkStainlessVersion("0.16.0");
```
KiewanVillatel, line 1545: Maybe double check this one with @sineha-mani, I don't remember if those endpoints were new or just old endpoints that we exposed with auto-generated oapi client. In case of doubt we can keep 0.16.0 to be on the safe side.

KiewanVillatel (approved):

PR #3241, opened 2026-07-21
chore: deprecate all legacy SDK methods from SmithDB migration guide

KiewanVillatel:
## Summary

Marks all legacy SDK methods from the [SmithDB SDK migration guide](https://docs.langchain.com/langsmith/smithdb-sdk-migration) as deprecated. Every deprecation includes a \`warnings.warn(DeprecationWarning)\` at runtime, a \`.. deprecated::\` RST docstring directive (Python), or a \`@deprecated\` JSDoc tag (TypeScript). Each message links to the specific section in the guide and mentions removal after Jan 31, 2027.

**Python \`Client\` and \`AsyncClient\` deprecated methods:**
- \`read_run()\` → \`client.runs.retrieve()\` ([#runs-retrieve](https://docs.langchain.com/langsmith/smithdb-sdk-migration#runs-retrieve))
- \`list_runs()\` → \`client.runs.query()\` ([#runs-query](https://docs.langchain.com/langsmith/smithdb-sdk-migration#runs-query))
- \`read_thread()\` → \`client.threads.list_traces()\` ([#threads-list-traces](https://docs.langchain.com/langsmith/smithdb-sdk-migration#threads-list-traces))
- \`list_threads()\` → \`client.threads.query()\` ([#threads-query](https://docs.langchain.com/langsmith/smithdb-sdk-migration#threads-query))
- \`get_run_stats()\` → \`client.threads.stats()\` ([#threads-stats](https://docs.langchain.com/langsmith/smithdb-sdk-migration#threads-stats))
- \`get_experiment_results()\` → \`client.datasets.experiment_runs.query()\` ([#dataset-experiment-runs-query](https://docs.langchain.com/langsmith/smithdb-sdk-migration#dataset-experiment-runs-query))
- \`add_runs_to_annotation_queue(run_ids=...)\` → \`add_runs_to_annotation_queue(runs=...)\` ([#annotation-queues-add-runs](https://docs.langchain.com/langsmith/smithdb-sdk-migration#annotation-queues-add-runs))

**Python legacy exception classes in \`langsmith.utils\`** ([#exceptions](https://docs.langchain.com/langsmith/smithdb-sdk-migration#exceptions)):
- All 10 \`LangSmith*\` exception classes deprecated with section-anchored URL

**TypeScript \`Client\` deprecated methods:**
- \`readRun()\`, \`listRuns()\`, \`readThread()\`, \`listThreads()\`, \`getRunStats()\`
[...]

PR #3299, opened 2026-07-30
chore: deprecate legacy SmithDB-migration SDK methods

KiewanVillatel:
## Summary

Rolls up the three stacked PRs that landed on `kiewan/deprecate-legacy-methods` into a single merge to `main`. Every deprecation is annotation + warning only — no signature changes, no behaviour change beyond the warning.

**1. Deprecate all legacy SDK methods from the [SmithDB SDK migration guide](https://docs.langchain.com/langsmith/smithdb-sdk-migration)** (#3241)

Python `Client` / `AsyncClient` — `@_deprecated` decorator plus a `.. deprecated::` RST directive on each docstring:

| Deprecated | Replacement |
| --- | --- |
| `read_run()` | `client.runs.retrieve()` |
| `list_runs()` | `client.runs.query()` |
| `read_thread()` | `client.threads.list_traces()` |
| `list_threads()` | `client.threads.query()` |
| `get_experiment_results()` | `client.datasets.experiment_runs.query()` |
| `add_runs_to_annotation_queue(run_ids=...)` | `add_runs_to_annotation_queue(runs=...)` |

TypeScript `Client` — `@deprecated` JSDoc plus a `warnOnce()` at call time on `readRun()`, `listRuns()`, `readThread()`, `listThreads()`,  and the `string[]` path of `addRunsToAnnotationQueue()`.

Also deprecates the 10 legacy `LangSmith*` exception classes in `langsmith.utils` in favour of their modern equivalents (`LangSmithError` → `LangsmithError`, `LangSmithAPIError` → `InternalServerError`, `LangSmithRequestTimeout` → `APITimeoutError`, …).

Every message links to the specific guide section and states the Jan 31, 2027 removal date.

This PR originally also warned from `create_feedback` / `createFeedback` when `run_id` was given without `session_id`. **That part is dropped** — `main` has since landed `_check_feedback_session_id` / `_checkFeedbackSessionId`, which raises on SmithDB-only deployments (`ch_query_enabled: false`) and warns elsewhere. That supersedes the unconditional warning, so main's version and its tests are taken as-is here.

**2. Deprecate `run_ids=` in `AsyncClient.add_runs_to_annotation_queue`** (#3206, [LSDK-291](https://linear.app/langchain/issue/LSDK-291))

#3241 covered the sync `Client` and the JS `Client` but not `async_client.py`, so the async client was silently missing the warning while its docstring still said the path "will be deprecated in a future release". Fills the gap with wording identical to `client.py` — same message, same `#annotation-queues-add-runs` anchor, same removal date. Go/Java analog: langchain-ai/langchainplus#30308.

**3. Deprecate `evaluate_run` and `aevaluate_run`** (#3296)

**No replacement** — callers should run the evaluator themselves and log the result with `client.create_feedback()`. Python only; there is no `evaluateRun` on the JS client. Note that `langsmith.beta._evals.compute_test_metrics` calls `client.evaluate_run` internally and will therefore surface the warning to its callers, and `python/README.md` still shows `client.evaluate_run(run, evaluator)` in its evaluation example — left for a follow-up since there is no replacement snippet to swap in.


python/langsmith/client.py
```diff
@@ -7541,6 +7586,11 @@ def _is_eval_results(results: Any) -> TypeGuard[ls_evaluator.EvaluationResults]:
             )
         return results_
 
+    @_deprecated(
+        "evaluate_run() is deprecated and will be removed after Jan 31, 2027. "
+        "There is no replacement: run the evaluator yourself and log the result "
+        "with client.create_feedback()."
+    )
     def evaluate_run(
```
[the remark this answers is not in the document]
KiewanVillatel: Fixed. get_test_results and getRunUrl will only warn in clickhouse only environments. Warning is valid, the methods would stop working if they don't switch to smith db.

python/langsmith/run_trees.py
```diff
@@ -919,7 +919,19 @@ def wait(self) -> None:
...
+        session_id = (
+            self.session_id
+            or self.client.read_project(project_name=self.session_name).id
+        )
+        response = self.client._get_langsmith_api_sync().runs.get_url(
+            str(self.id),
+            project_id=str(session_id),
+            trace_id=str(self.trace_id),
+            start_time=self.start_time.isoformat(),
+        )
```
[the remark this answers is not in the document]
KiewanVillatel: Fixed

python/langsmith/run_trees.py
```diff
@@ -919,7 +920,24 @@ def wait(self) -> None:
...
```
QuentinBrosse: This turns `get_url()` from a local string build into a network call on every invocation (plus `client.info`, and `read_project` when `session_id` is unset). Before, `_construct_run_url` only needed the tenant id, which is cached, so it was effectively free after the first call.

People call `get_url()` from logging and callbacks, so that's a request per log line — and it can now raise on a transient blip where it previously couldn't fail at all. Could we cache the result on the RunTree, and/or fall back to `_construct_run_url` if the v2 call errors? You already have that fallback right above for ClickHouse-only.

python/langsmith/run_trees.py
```diff
@@ -919,7 +920,24 @@ def wait(self) -> None:
...
         """Return the URL of the run."""
-        return self.client.get_run_url(run=self)
+        backend = _v2_migration_utils.get_query_backend(self.client.info.instance_flags)
+        if backend == _v2_migration_utils.QueryBackend.CLICKHOUSE_ONLY:
+            # The v2 endpoint doesn't exist on ClickHouse-only backends, which may
+            # predate it; build the URL locally there instead.
+            return self.client._construct_run_url(run=self)
+        session_id = (
+            self.session_id
+            or self.client.read_project(project_name=self.session_name).id
```
QuentinBrosse: `_construct_run_url` ends with an `else: raise ValueError` for the case where neither `session_id` nor `session_name` is set. Here that falls through to `read_project(project_name=None)`, so instead of the clear error you get whatever that surfaces. Worth keeping the guard.

QuentinBrosse:
### Missing coverage: the `Share and read public runs` section

The [migration guide](https://docs.langchain.com/langsmith/smithdb-sdk-migration#share-and-read-public-runs) has a section for the run-sharing methods, but none of them got touched in this stack. Unlike every other section, there's no decorator, no `warnOnce()`, no docstring admonition and no mention of the replacement — so a user reading the guide is told to migrate off `share_run()`, while a user reading the SDK gets zero signal that anything is changing (no runtime warning, nothing in the reference docs, no IDE strikethrough).

| Method | Replacement |
| --- | --- |
| `Client.share_run`, `AsyncClient.share_run`, `Client.shareRun` | `runs.share.create()` |
| `Client.unshare_run`, `Client.unshareRun` | `runs.share.delete()` |
| `Client.read_run_shared_link`, `AsyncClient.read_run_shared_link`, `Client.readRunSharedLink` | `runs.retrieve(selects=["SHARE_URL"])` |
| `Client.read_shared_run` | `public.runs.retrieve()` |
| `Client.list_shared_runs`, `Client.listSharedRuns` | `public.runs.query()` |

Happy for this to be a follow-up rather than growing this stack — just want it tracked somewhere, since right now it's the one guide section with no code-side counterpart.

One gotcha worth flagging up front: `run_is_shared()` calls `read_run_shared_link()` internally in both `client.py` and `async_client.py`, so it needs the `_suppress_deprecation_warning()` wrap from #3308 or it reintroduces exactly the nested-warning bug that PR fixes.

<details>
<summary>Prompt for Claude Code, if useful</summary>

```
Deprecate the run-sharing methods listed in the "Share and read public runs" section of
https://docs.langchain.com/langsmith/smithdb-sdk-migration, following the exact pattern
already used in this repo for read_run/list_runs/read_thread/list_threads.

Methods and replacements:
  python/langsmith/client.py
    share_run             -> client.runs.share.create()
    unshare_run           -> client.runs.share.delete()
    read_run_shared_link  -> client.runs.retrieve(selects=["SHARE_URL"])
    read_shared_run       -> client.public.runs.retrieve()
    list_shared_runs      -> client.public.runs.query()
  python/langsmith/async_client.py
    share_run             -> client.runs.share.create()
    read_run_shared_link  -> client.runs.retrieve(selects=["SHARE_URL"])
  js/src/client.ts
    shareRun              -> client.runs.share.create()
    unshareRun            -> client.runs.share.delete()
    readRunSharedLink     -> client.runs.retrieve({ selects: ["SHARE_URL"] })
    listSharedRuns        -> client.public.runs.query()

For each Python method, apply all three parts, copying the wording style of the
existing deprecations in client.py:
  1. @_deprecated(...) from langsmith._internal._beta_decorator, message reading
     "<name>() is deprecated and will be removed after Jan 31, 2027. Use <replacement>
     instead. See https://docs.langchain.com/langsmith/smithdb-sdk-migration
     #share-and-read-public-runs for the migration guide."
  2. a ".. admonition:: Deprecated" block as the first thing in the docstring, before
     Args, naming the replacement with :meth:, the guide link and the removal date.
  3. keep the signature and behaviour identical — annotation and warning only.

For each TypeScript method: a one-line `@deprecated` JSDoc tag plus a warnOnce() call at
the top of the body, using the { type: "DeprecationWarning", code: "LANGSMITH_DEPRECATED_<NAME>" }
option shape already used by readRun/listRuns.

Important: run_is_shared() in client.py and async_client.py calls read_run_shared_link()
internally. Wrap those calls in _suppress_deprecation_warning() (from
langsmith._internal._beta_decorator) so a supported method doesn't warn about a
deprecated one the caller never invoked. Check for any other internal callers of these
methods and do the same. Do not wrap blocks that execute caller-supplied code.

Add unit tests mirroring the existing ones: each deprecated method warns exactly once
with the right message, and run_is_shared() warns for itself but never for
read_run_shared_link(). Stub only the HTTP layer so the real decorated method runs —
verify each new test actually fails when the deprecation or the suppression is removed,
so it isn't vacuous.

Then run: cd python && uv run --frozen pytest tests/unit_tests -q
and: cd js && pnpm test:single src/tests/client.test.ts
```

</details>

python/langsmith/run_trees.py
KiewanVillatel: Good catch — done in 2b6800b5. `get_url()` now resolves once and memoizes on the RunTree (`_cached_url`), and the whole remote path (`client.info`, `read_project`, `runs.get_url`) is wrapped so any failure — or a null `url` in the response — logs at debug and falls back to `_construct_run_url`, same as the ClickHouse-only branch. So it can no longer raise where the local string build never could, and it is at most one request per run.

python/langsmith/run_trees.py
KiewanVillatel: Slight correction: `_construct_run_url` does not raise there — its final `else` falls back to `ls_utils.get_tracer_project()` and reads that project. (`read_project(project_name=None)` does raise a clear `ValueError("Must provide project_name or project_id")`, so it was not silent, just a different error than before.)

Either way the asymmetry was real, so the v2 path now resolves the project the same way: `self.session_name or utils.get_tracer_project()`, with a test covering it. And since the fallback from the other comment catches everything, a run with no project resolvable at all ends up exactly where it did before this PR.

QuentinBrosse (approved):

KiewanVillatel:
Tracked and done as a follow-up rather than growing this stack: #3312, stacked on #3308.

Covers all five Python methods and all four JS ones from your table, with the `_suppress_deprecation_warning()` wrap on the `run_is_shared()` → `read_run_shared_link()` call in both `client.py` and `async_client.py` (that was the only internal caller of any of them in either language). Tests verified non-vacuous by stripping the decorators / `warnOnce()` calls / suppressions.

PR #3305, opened 2026-07-31
fix(js): apply caller-supplied headers to the v2 endpoints

KiewanVillatel:
Follow-up to #3304, which fixed the equivalent problem on the Python side. The JS client has no TLS/proxy gap — both halves speak `fetch`, and `fetch`/`fetchOptions` are already forwarded to the generated client — but its **headers** are forwarded inconsistently.

## Problem

There are two ways to supply headers, and each one reaches only one half of the client:

| Channel | Handwritten paths | Generated OpenAPI client (v2 resources) |
|---|---|---|
| `config.headers` | applied via `_mergedHeaders` | **dropped** — `defaultHeaders` was only set in the no-auth case |
| `config.fetchOptions.headers` | applied, but **replaces the whole header set including `x-api-key`** | **dropped** — destructured out |

So `new Client({ headers: { "X-Corp-Trace": "..." } })` silently did nothing for `client.runs`, `client.evaluators`, `client.sandboxes`, `client.datasets`, `client.threads`, `client.traces` and friends. And `fetchOptions.headers` was an auth footgun on the other half, because the ~40 handwritten call sites spread `...this.fetchOptions` *after* `headers: this._mergedHeaders`.

## ⚠️ Behavior change for `fetchOptions.headers` users

**`fetchOptions.headers` is no longer a whole-set replacement.** It used to be spread after `headers: this._mergedHeaders`, so it replaced every header on the request; it is now merged *underneath* the headers the SDK requires. Anyone relying on the old behavior to override a header the SDK sets will silently lose that override:

- `x-api-key` / `x-tenant-id` / `Authorization` — intentional, that was the auth footgun.
- **`Content-Type`** — worked before and no longer does, because the call sites apply it after `_mergedHeaders` (e.g. `headers: { ...this._mergedHeaders, "Content-Type": "application/json" }`). This is a plausible legitimate use, so it's worth calling out in the release notes.

Headers the SDK does *not* set are unaffected and now reach both halves.

## Approach

Both channels are merged once, in `_callerHeaders`, and applied *below* the headers the SDK requires:

- **`fetchOptions.headers` is split out at construction.** Storing the rest as `Omit<RequestInit, "headers">` means no call site can clobber the required headers by spreading `fetchOptions` late — including call sites added later. The alternative was reordering ~40 spread sites, which any new site could silently regress.
- **The merged result is forwarded as `defaultHeaders`, minus the headers the SDK actually sets.** This filtering is load-bearing: the generated client's `buildHeaders` order is `[base, authHeaders, defaultHeaders, bodyHeaders, options.headers]` and later entries win, so an unfiltered forward would let a caller-supplied header overwrite the auth headers.
- **Only what the SDK *actually* supplies is reserved.** Passing an explicit `Authorization` or `x-api-key` with no configured credential is a supported way to authenticate (`hasExplicitAuthHeader` exists for it), so those survive when the SDK has nothing to replace them with. A caller-supplied `x-api-key` is additionally handed to the generated client as its `apiKey` option rather than as a header, because its `validateHeaders` only accepts the header when the client's own `apiKey` is set.
- **The generated client is built on first use and rebuilt whenever its derived auth/headers change.** It captures `defaultHeaders` once, while the handwritten paths recompute `_mergedHeaders` per request — so a snapshot can go stale in two ways: `get headers` returns `_customHeaders` by reference (`client.headers["X-Foo"] = "1"` never fires the setter), and `_sdkControlledHeaders` is time-dependent (for a profile holding only a refresh token, `currentAuthHeaderFromProfile` returns undefined, so `authorization` isn't reserved and a caller's header would stay in the snapshot for good, making `hasExplicitAuthHeader` skip profile auth on every v2 request). Rebuilding on change closes both.
- **Both channels are normalized and validated.** Names/values round-trip through `Headers`, which rejects a name containing a newline, and duplicate spellings fold case-insensitively with the last value winning — within a single channel as well as across them. `{"X-Foo": "a", "x-foo": "b"}` used to survive as both keys and be joined by `Headers` into `"a, b"` instead of overriding. `_customHeaders` is stored as the caller gave it (the public getter hands it back), so folding happens when headers are built; malformed input still fails eagerly in the constructor and the setter.

## Testing

16 tests in `js/src/tests/client_headers.test.ts` — a new file rather than an addition to `client.test.ts`, see below. They cover: both channels reaching the generated client end-to-end (asserted on the outgoing request headers via `fetchImplementation`), neither channel being able to override `x-api-key` on either half, required headers surviving `fetchOptions.headers` on the handwritten path, a caller credential surviving when the SDK has none, headers mutated through the getter reaching the v2 resources, a refreshed profile header taking over from a caller's `Authorization`, intra- and cross-channel duplicate spellings resolving to the last value, and malformed names being rejected through all three entry points.

`pnpm run check:types`, `pnpm run lint` (0 errors) and `pnpm run format` are clean.

### Note on the test file location

These tests initially lived in `client.test.ts` and turned `openai_agents_sdk.test.ts` red in CI (19 failures) while passing locally. Bisected by pushing the source change with the tests removed — CI went fully green, so `client.ts` was never implicated.

That suite depends on no run-tree context being active when a test starts, and the `AsyncLocalStorage` instance lives on `globalThis`, so it is shared across test files within a jest worker. Lengthening `client.test.ts` changes how jest distributes files across workers, which changes what runs before that suite. Keeping these tests in their own file avoids perturbing it.

The underlying fragility is still there: `onTraceEnd` restores context with `enterRunTreeContext(previousStore)` where `previousStore` may be `undefined`. Left alone here as out of scope — worth its own fix, since any change that reshuffles jest's file distribution can set this off.


QuentinBrosse:
Two things I'd fix before this lands, both small.

**Empty-string caller `x-api-key` breaks every v2 call.** In `_openAPIAuth`, `apiKey = headers[callerApiKeyName] ?? undefined` (L1642) accepts `""`, which then skips the `X-API-Key: null` escape hatch just below it. In the generated client `apiKeyAuth` sends the empty header anyway (`this.apiKey == null` is false) and `validateHeaders` throws — falsy `apiKey`, nothing in `nulls`, no `tenantID`:

```
new Client({ headers: { "x-api-key": "" } })   // no other credential configured

  v2:          throws "Could not resolve authentication method..."
  handwritten: sends x-api-key: "" → 401
```

On main this input worked, since `config.headers` was dropped and `defaultHeaders` was `{"X-API-Key": null}`. So it's a regression, and it's the same two-halves-disagree split the PR is closing everywhere else. `hasValue()`, like the constructor does at L1378, covers it.

**`stops forwarding a caller Authorization once profile auth supplies one` fails with `LANGSMITH_API_KEY` in the environment.**

```
$ LANGSMITH_API_KEY=... jest src/tests/client_headers.test.ts
  ✕ stops forwarding a caller Authorization once profile auth supplies one
      Expected: "Bearer caller-token"   Received: null
  15 passed, 1 failed
```

`apiKey: undefined` doesn't mean "no key" — `config.apiKey ?? defaultConfig.apiKey` picks up the env one, `profileAuth` gets nulled at L1379 so the stub is never consulted, and `applyCurrentAuthHeaders` deletes the caller's `Authorization`. The third `describe` block already has the right `beforeEach` (env scrub + empty `LANGSMITH_CONFIG_FILE`); hoisting it so all three share it should do it. Green in CI only because CI's env is clean, which is the same thing you were fighting when you bisected earlier.

Rest of it looks good to me — the `Omit<RequestInit, "headers">` trick and the rebuild-on-signature-change getter are the right shapes, and the `Content-Type` callout in the description is a good catch.

QuentinBrosse (approved):

PR #3317, opened 2026-07-31
fix(js,py): stamp ls_agent_type on wrap_openai LLM runs

ybathula707:
## What

`wrap_openai` never emitted `ls_agent_type` on any traced LLM run. In OpenAI Responses stateful mode, each traced run's input is just the incremental context, so the BE messages/v2 layer hashes every run into its own "unique" conversation, including single-turn LLM runs. Main-conversation selection then picks a "winner" by row count, so turns in the losing conversations get dropped.

`ls_agent_type` tagging fixes this — anchors runs to the root conversation via the `ls_agent_type == "root"` short-circuit in the association layer, and tags `middleware` / `subagent` so the parser can filter or group non-main content.

## How

Default-tag resolution + shared constants live in a small SDK-agnostic util so future maintainability is easier. Less places to update if this field starts to accept more params. 

- `python/langsmith/_internal/_ls_agent_type.py` 
- `js/src/utils/ls_agent_type.ts` 

`wrap_openai` (Py + JS) stamps at the wrapper-config layer so user-supplied values are preserved.

**Tagging precedence per run:**

1. User-supplied `ls_agent_type` (any known value: `root` / `middleware` / `subagent` / `compaction`) → always preserved. Propagates down via traceable's `outer_metadata` mechanism.
2. Default `root` stamped **only on the trace-root run** when the user didn't supply. Nested runs don't restamp — propagation carries the trace root's tag down.
3. `ls_agent_type: None` (Py) / `null` (JS) → explicit opt-out. Passes through as a null value that actively overrides any propagated parent tag via traceable's `metadata_.update(wrapper_metadata)`. Backend treats null as no-tag (`ls_agent_type == "root"` check fails → content-hash grouping fallback, same as an absent key).

Fixes [LSO-3680](https://linear.app/langchain/issue/LSO-3680). Related [LSO-3088](https://linear.app/langchain/issue/LSO-3088).

## Test Plan

- [x] Python util: `python/tests/unit_tests/test_ls_agent_type.py` — resolver + constants. 5/5 pass.
- [x] Python wrap_openai: `python/tests/unit_tests/wrappers/test_openai_processing.py` — wrapper-config layer, per-call user tag, null opt-out (top-level + nested), and E2E propagation through `_context._METADATA`. 22/22 pass.
- [x] JS util: `js/src/tests/ls_agent_type.test.ts` — resolver + spread helper + constants. 7/7 pass.
- [ ] Post-merge: `POST /v1/trajectory` against reference prod thread `7ef2a18d-779f-432b-b030-088f141fd7a4` returns full 3-turn conversation.


jacoblee93 (approved):
Looks fine but let's do for JS too

js/src/wrappers/openai.ts
```diff
@@ -380,6 +403,7 @@ const getChatModelInvocationParamsFn = (
         ...prepopulatedInvocationParams,
         ...ls_invocation_params,
       },
+      ...(lsAgentType !== undefined ? { ls_agent_type: lsAgentType } : {}),
```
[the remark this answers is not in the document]
ybathula707: Yeah traced this — traceable doesn't expose a "delete propagated key" primitive without touching `run_helpers.py`, so went with null-value-as-opt-out instead of strict absence. Null passes through and overrides propagation via `metadata_.update()`, which is functionally the same for the backend (`ls_agent_type == "root"` fails on null too, same content-hash fallback). Added `test_nested_none_opt_out_overrides_propagated_tag` for the middleware-parent case you flagged. Only visible diff is `{"ls_agent_type": null}` vs `{}` — reopen if that matters for consumers we haven't audited.

js/src/wrappers/openai.ts
```diff
@@ -7,6 +7,25 @@ import {
...
 } from "../traceable.js";
 import { InvocationParamsSchema, KVMap } from "../schemas.js";
+import { getCurrentRunTree } from "../singletons/traceable.js";
+
+export const _resolveLsAgentType = (
+  requestMetadata: Record<string, unknown>,
+): string | undefined => {
+  const supplied = requestMetadata.ls_agent_type;
+  if (
+    supplied === "root" ||
```
jacoblee93: What if we add more in the future?

Would also generally prefer this to be a util since we'll be reusing it across wrappers

python/langsmith/wrappers/_openai.py
```diff
@@ -154,6 +154,30 @@ def _infer_invocation_params(
...
 
 
+def _resolve_default_ls_agent_type() -> Optional[str]:
+    """Default ls_agent_type: root at the trace root, None when nested."""
+    if run_helpers.get_current_run_tree() is None:
+        return "root"
+    return None
+
+
+def _traceable_kwargs_with_ls_agent_type(textra: dict) -> dict:
```
jacoblee93, line 158: See above - given that this will be used everywhere, let's just make it a shared util

python/langsmith/wrappers/_openai.py
```diff
@@ -154,6 +154,30 @@ def _infer_invocation_params(
...
+    """Return textra with ls_agent_type defaulted at the wrapper-config layer.
+
+    A user-supplied value (including None) is preserved unchanged; None flows
+    through as an explicit opt-out that overrides any propagated parent tag via
+    traceable's metadata_.update(wrapper_metadata) step.
+    """
+    result = dict(textra)
+    metadata = dict(result.get("metadata") or {})
+    if "ls_agent_type" not in metadata:
```
jacoblee93: nit: invert and just early return?

python/langsmith/wrappers/_openai.py
ybathula707: that makes a lot of sense. Let me extract  into a utils.

js/src/wrappers/openai.ts
ybathula707: extracted.

python/langsmith/wrappers/_openai.py
ybathula707: done.

jacoblee93 (approved):

PR #3318, opened 2026-07-31
fix(js,py): preserve user-supplied ls_agent_type in openai-agents-sdk integration

ybathula707:
## What
OpenAI Agents SDK integration had two bugs preventing users from controlling ls_agent_type:

1. `on_trace_start` never read `trace.metadata` — so `RunConfig.trace_metadata` (Py) / `RunConfig.traceMetadata` (JS) was silently dropped.
2. The processor unconditionally stamped `ls_agent_type: "root"` on trace-start and `"subagent"` on the subagent-detection path — so even wrapper-level metadata (`OpenAIAgentsTracingProcessor(metadata=...)`) got overwritten.

## How

Both hook sites in both languages now :
1. Read `trace.metadata`
2. Delegate default-tag resolution to the shared  ls_agent_type util (Py) /(JS) util where key-presence check preserves whatever the user set, including `None`/`null` opt-out, and stamping root if it's a root run that's not been tagged.
3. Use `NON_ROOT_LS_AGENT_TYPES` from the util at the subagent-detection site 

In Py, updated `on_trace_start` and the subagent stamp in `on_span_start` (`_openai_agents.py`).
In JS, updated `onTraceStart` and the subagent stamp in `onSpanStart` (`openai_agents.ts`).

Fixes [LSO-3608](https://linear.app/langchain/issue/LSO-3608).

## Test Plan

- [x] Python unit tests: `python/tests/unit_tests/wrappers/test_openai_agents_processor.py` — trace-start + subagent cases 15/15 pass.
- [x] Python util tests: `python/tests/unit_tests/test_ls_agent_type.py`  7/7 pass.
- [x] JS unit tests: `js/src/tests/openai_agents_sdk.test.ts` — parallels the Python precedence coverage 39/39 pass.
- [x] JS util tests: `js/src/tests/ls_agent_type.test.ts`. 10/10 pass.
- [ ] Post-merge: fixture regen on langchainplus — 6

### DIFF
--- a/js/src/wrappers/openai_agents.ts
+++ b/js/src/wrappers/openai_agents.ts
@@ -12,6 +12,7 @@ import {
   AsyncLocalStorageProviderSingleton,
   getCurrentRunTree,
 } from "../singletons/traceable.js";
+import { defaultLsAgentTypeMetadata } from "../utils/ls_agent_type.js";
 import type { ContextPlaceholder } from "../singletons/types.js";
 import type {
   AgentSpanData,
@@ -589,10 +590,10 @@ export class OpenAIAgentsTracingProcessor implements TracingProcessor {
       ...((trace.metadata as Record<string, unknown> | undefined) ?? {}),
       ls_integration: "openai-agents-sdk",
     };
-    // Default root only at true trace root; nested traces inherit via create_child.
-    if (!("ls_agent_type" in mergedMetadata) && currentRunTree === undefined) {
-      mergedMetadata.ls_agent_type = "root";
-    }
+    Object.assign(
+      mergedMetadata,
+      defaultLsAgentTypeMetadata(mergedMetadata, currentRunTree),
+    );
     const runExtra: Record<string, unknown> = { metadata: mergedMetadata };
 
     const traceDict = (trace.toJSON() as Record<string, unknown>) ?? {};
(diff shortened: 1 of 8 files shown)


js/src/wrappers/openai_agents.ts
```diff
@@ -583,14 +587,18 @@ export class OpenAIAgentsTracingProcessor implements TracingProcessor {
...
+    // Merge three sources: processor defaults, trace.metadata (user's per-run
+    // ls_agent_type), and force-set ls_integration.
+    const mergedMetadata: Record<string, unknown> = {
+      ...this._metadata,
+      ...((trace.metadata as Record<string, unknown> | undefined) ?? {}),
+      ls_integration: "openai-agents-sdk",
     };
+    Object.assign(
+      mergedMetadata,
+      defaultLsAgentTypeMetadata(mergedMetadata, currentRunTree),
```
jacoblee93, line 599: Why have this as a separate step instead of spreading? What makes this special?

jacoblee93 (commented):
Just a small q

jacoblee93 (approved):

js/src/wrappers/openai_agents.ts
ybathula707: @jacoblee93 
this is cuz `defaultLsAgentTypeMetadata` takes `mergedMetadata` as an argument (to check if user already set ls_agent_type), before assigning it back to the metadata. So mergedMetadata has to be built first.

Idt you can spread it inline in the same block (line 594) because mergedMetadata is still being created at that time

ybathula707: keeping but noticing that object.assign is not typically preferred vs using 'let'  since mutations can be dangerous.

PR #3328, opened 2026-08-03
fix!: fail-closed when per-function anonymization callables fail

emil-lc:
## Problem

`process_inputs` / `process_outputs` (Python `@traceable`) and `processInputs` / `processOutputs` (JS `traceable()`) strip secrets before a run is traced. When one of them raised, the SDK logged the error and uploaded the run with the **raw, unredacted** payload.

The client-level anonymizer (`anonymizer` / `hide_inputs` / `hide_outputs`) already fails closed. This brings the per-function hooks in line.

## Change

When a redaction callable raises or rejects, its payload is replaced with a marker and the run is traced as normal:

```json
{"ls_error": "Processing failed; inputs dropped (process_inputs: KeyError)"}
{"ls_error": "Processing failed; outputs dropped (processOutputs)"}
```

All messages share the `Processing failed; ` prefix, built from one constant, so they are greppable as a set.

Python names the exception **type**, never its message: a redactor's error often echoes the payload it was reading (`KeyError: 'super-secret-value'`; pydantic quotes `input_value` verbatim), so the full text is logged locally and never uploaded. The name is rejected unless it is a plain identifier, since `__name__` is assignable. JS omits the type entirely — reading `e.constructor` can run user code (a throwing getter or a Proxy `get` trap) from inside the `catch`, which would escape and skip the traced function.

The raw value never reaches `create_run` / `update_run`. The wrapped function still runs and returns normally. The run is ended, not marked errored — an error status is for the traced function failing, not for tracing.

`RunTree` is untouched.

Also fixed: in Python, a throwing `process_outputs` used to skip the patch and leave the run un-ended. The redactor call moved out of the `try` around `_container_end`.

## ⚠️ Breaking change

Reading these runs back now gives the marker instead of the values. Evaluators will fail rather than score unredacted data. The SDK passes `process_inputs` internally in the eval path (`evaluation/_runner.py`, `evaluation/evaluator.py`, `evaluation/_arunner.py`), so a bug there degrades a whole experiment, not just a hand-written redactor.

This is intended. A redactor that throws has redacted nothing, and uploading its input is the leak. A `WARNING` naming the run is logged.

**Escape hatch:** set `LANGSMITH_ALLOW_UNPROCESSED_PAYLOADS=true` (or `LANGCHAIN_`) to restore the old fail-open behavior if this blocks you. It uploads data no redactor processed, so it is off by default and logs a warning naming itself whenever it takes effect.

## Tests

Python: two for the marker, two for the escape hatch. JS `traceable`: three for the marker (sync throw, async rejection, outputs) plus two for the escape hatch. Vercel telemetry (vitest): three for the marker plus one for the escape hatch. The two pre-existing `"… throwing error does not affect invocation"` tests were flipped — they had encoded the fail-open leak.

The Python and `traceable` tests raise an error that embeds the secret, then assert it reaches no payload, that the marker is present, and that the wrapped function still returns normally. The telemetry tests assert the marker on the run each hook governs.

### DIFF
--- a/js/src/run_trees.ts
+++ b/js/src/run_trees.ts
@@ -311,6 +311,13 @@ export class RunTree implements BaseRun {
   distributedParentId?: string;
 
   /**
+   * When true, postRun()/patchRun() are no-ops. Set when a `processInputs`
+   * redactor fails, so the run is dropped rather than uploaded unredacted
+   * (mirrors the client-level anonymizer).
+   */
+  dropped = false;
+
+  /**
    * @interface
    */
   private _serialized_start_time: string | undefined;
@@ -807,6 +814,8 @@ export class RunTree implements BaseRun {
     if (this._awaitInputsOnPost) {
       this.inputs = await (this.inputs as Promise<KVMap>);
     }
+    // After resolving inputs, so an async redactor that rejected is honored.
+    if (this.dropped) return;
     try {
       const runtimeEnv = getRuntimeEnvironment();
       if (this.replicas && this.replicas.length > 0) {
@@ -874,6 +883,7 @@ export class RunTree implements BaseRun {
    * to `false`.
    */
   async patchRun(options?: { excludeInputs?: boolean }): Promise<void> {
+    if (this.dropped) return;
     const excludeInputs = options?.excludeInputs ?? getExcludeInputsOnPatch();
     if (this.replicas && this.replicas.length > 0) {
       for (const {
--- a/js/src/run_trees.ts
+++ b/js/src/run_trees.ts
@@ -311,13 +311,6 @@ export class RunTree implements BaseRun {
   distributedParentId?: string;
 
   /**
-   * When true, postRun()/patchRun() are no-ops. Set when a `processInputs`
-   * redactor fails, so the run is dropped rather than uploaded unredacted
-   * (mirrors the client-level anonymizer).
-   */
-  dropped = false;
-
-  /**
    * @interface
    */
   private _serialized_start_time: string | undefined;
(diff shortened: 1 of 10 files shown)


js/src/run_trees.ts
```diff
@@ -305,6 +305,13 @@ export class RunTree implements BaseRun {
 
   distributedParentId?: string;
 
+  /**
+   * When true, postRun()/patchRun() are no-ops. Set when a `processInputs`
+   * redactor fails, so the run is dropped rather than uploaded unredacted
```
jacoblee93: Why don't we just clear inputs/outputs/metadata instead with a message like `Processing/inputs/outputs/metadata failed, dropping inputs/outputs/metadata`? Seems clearer to the user what's happening and we don't have to add anything to the run tree

jacoblee93 (commented):
I would rather make this an opt-in setting on specific sensitive `processInputs` and/or do this in a minor bump

js/src/run_trees.ts
emil-lc: Now we're replacing with static constants. I agree this is a Minor bump. Are you OK with leaving it at that and not doing yet another obscure feature flag?

python/langsmith/run_helpers.py
```diff
@@ -58,6 +58,33 @@
...
+    """Build the marker that replaces a payload whose redactor raised.
+
+    Carries the exception type, never its message: an exception routinely echoes
+    the payload it was reading (``KeyError: 'secret'``), so the full text is
+    logged locally and never uploaded. ``__name__`` is assignable, hence the
+    identifier check.
+    """
+    kind = type(e).__name__
+    kind = kind[:64] if kind.isidentifier() else "unknown"
+    return f"{_PROCESSING_FAILED_PREFIX}{payload} dropped (process_{payload}: {kind})"
```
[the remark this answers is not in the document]
emil-lc: That would be an extremely rare and unusual occurrence, if theoretically possible. Adressing it is not worth the extra complexity/loss of visibility.

js/src/run_trees.ts
emil-lc: Implemented the flag after all - just in case someone needs a quick escape hatch after an upgrade introduces unexpected filtering - `LANGSMITH_ALLOW_UNPROCESSED_PAYLOADS`

QuentinBrosse (approved):

PR #3345, opened 2026-08-06
Fix create_child() ignoring tracing_context(replicas=...) when a parent is set

### LINKED ISSUE #3344: tracing_context(replicas=...) is silently ignored for any run created via RunTree.create_child() (i.e. anything with a parent)
`tracing_context(replicas=[...])` is documented as an ambient override for where new runs get written, alongside `parent`, `project_name`, `tags`, etc. It works correctly for a **root** run (no parent in context), but is silently ignored for **any run created via `RunTree.create_child()`** — which is every run inside a distributed-tracing scenario (`tracing_context(parent=<headers>)`), since the placeholder parent reconstructed by `RunTree.from_headers()` always has empty `replicas` unless the caller explicitly forwarded a `langsmith-replicas` baggage field.

**Root cause:** `RunTree.create_child()` (`python/langsmith/run_trees.py`) passes `replicas=self.replicas` explicitly when constructing the child `RunTree`. Since this is never `None`, it bypasses the `if values.get("replicas") is None: values["replicas"] = _REPLICAS.get()` fallback in `RunTree`'s model validator — the only place the ambient `_REPLICAS` context var is ever actually consulted for a non-root run.

```python
# run_trees.py, RunTree.create_child()
run = RunTree(
    ...
    replicas=self.replicas,   # always set (even if []), so the contextvar fallback never triggers
    ...
)
```

## Repro

```python
import asyncio
from langsmith.run_helpers import tracing_context, get_current_run_tree

@traceable(run_type="chain", name="caller_root", project_name="project-a")
async def caller():
    headers = get_current_run_tree().to_headers()
    headers.pop("baggage", None)  # simulate a distributed caller sending only the trace id
    await callee(headers)

@traceable(run_type="chain", name="callee_root")
async def callee(headers):
    with tracing_context(
        parent=headers,
        replicas=[{"project_name": "project-b", "primary": True}],
    ):
        await child()

@traceable(run_type="llm", name="child")
async def child():
    return "ok"

asyncio.run(caller())
```

**Expected:** `callee_root` and `child` are written to `project-b`, per the `replicas` override.

**Actual:** `replicas` is silently ignored — both runs land in the default project (`LANGSMITH_PROJECT` / `session_name`) instead. No error, no warning — makes this very hard to notice in production.

## Environment
- `langsmith` == 0.10.15
- Python 3.14

## Suggested fix

```python
replicas=self.replicas or _REPLICAS.get(),
```
in `create_child()`, matching the fallback precedence already used for root-level runs. I vone-line change resolves the issue (both replicas fire correctly afterward).

moh3nhadavi:
Fixes #3344

## Description

`RunTree.create_child()` passed `replicas=self.replicas` unconditionally when constructing the child. Since this is never `None` (an empty list, not `None`, when the parent has no replicas of its own), it bypassed the `if values.get("replicas") is None: values["replicas"] = _REPLICAS.get()` fallback in `RunTree`'s validator — the only place the ambient `_REPLICAS` contextvar is consulted for a non-root run. So `tracing_context(replicas=...)` was silently ignored for any run with a parent, including every run inside a distributed-tracing scenario (`tracing_context(parent=<headers>)`), since the placeholder parent from `RunTree.from_headers()` starts with an empty replicas list unless the caller forwarded one via baggage.

## Fix

```python
replicas=self.replicas or _REPLICAS.get(),
```

in `create_child()`, matching the fallback already used for root-level runs.

## Testing

Added `test_create_child_honors_ambient_replicas_when_parent_has_none` to `TestRunTreeReplicas`. Fails on `main` before this change, passes after; full `test_replica_endpoints.py` suite still passes.

### DIFF
--- a/python/langsmith/run_trees.py
+++ b/python/langsmith/run_trees.py
@@ -652,7 +652,7 @@ class RunTree(ls_schemas.RunBase):
             extra=child_extra,
             parent_run=self,
             project_name=self.session_name,
-            replicas=self.replicas,
+            replicas=self.replicas or _REPLICAS.get(),
             ls_client=self.ls_client,
             tags=tags,
             attachments=attachments or {},  # type: ignore
--- a/python/tests/unit_tests/test_replica_endpoints.py
+++ b/python/tests/unit_tests/test_replica_endpoints.py
@@ -10,6 +10,7 @@ import pytest
 
 from langsmith import Client, compute_run_id_for_secondary_replica
 from langsmith import utils as ls_utils
+from langsmith.run_helpers import tracing_context
 from langsmith.run_trees import (
     ApiKeyAuth,
     AuthHeaders,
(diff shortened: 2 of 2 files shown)


emil-lc:
Closed in favor of https://github.com/langchain-ai/langsmith-sdk/pull/3346

PR #3359, opened 2026-08-10
refactor(py): drop redundant set handling in _serialize_json

emil-lc:
Stacked on #3358. LSDK-415 cleanup item **S2** (last in the series).

`_serialize_json` special-cased `(set, tuple)` and returned `list(obj)`, which duplicates `_simple_default`'s existing `set/frozenset/deque -> list` branch. Narrowed the check to `tuple` so:

- NamedTuples still convert via `_asdict()` and plain tuples via `list()` (unchanged);
- bare sets fall through to `_simple_default`, producing the identical `list` output with a single owner for that behavior.

Verified equivalent: `_serialize_json({1,2,3})` and `_simple_default({1,2,3})` both yield `[1,2,3]`. (The full `(set, tuple)` branch could **not** be removed — `_simple_default((1,2))` returns the string `"(1, 2)"`, so `tuple` must stay.)

- Tiny trade-off: a set now walks the (cold) method-loop `hasattr` checks before reaching `_simple_default`. Negligible; sets are rare in payloads.
- Added `test__dumps_json_serializes_set_as_array`.

### DIFF
--- a/python/langsmith/_internal/_serde.py
+++ b/python/langsmith/_internal/_serde.py
@@ -104,11 +104,13 @@ _serialization_methods: list[tuple[str, dict[str, Any]]] = [
 #               rust/crates/langsmith-pyo3/src/serialization/mod.rs
 def _serialize_json(obj: Any) -> Any:
     try:
-        if isinstance(obj, (set, tuple)):
+        if isinstance(obj, tuple):
             if hasattr(obj, "_asdict") and callable(obj._asdict):
                 # NamedTuple
                 return obj._asdict()
             return list(obj)
+        # NB: bare set/frozenset/deque fall through to _simple_default, which
+        # converts them to lists (same result, one owner for that behavior).
 
         for attr, kwargs in _serialization_methods:
             if (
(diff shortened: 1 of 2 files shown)


emil-lc:
Dropping this one. The Open SWE review is correct: removing `set` from the `(set, tuple)` branch is **not** the pure no-op this PR claimed. A `set` subclass exposing `model_dump`/`dict`/`to_dict` now hits the method-probe loop before `_simple_default`, so e.g. `class S(set): def to_dict(self): ...` serializes via `to_dict` instead of `list(obj)` — a behavior change for a valid input. (Bare `set`/`frozenset`/`deque` are unaffected; verified.)

Blast radius is ~nil in practice, but the whole justification for S2 was "redundant, no behavior change," and that no longer holds. The suggested guard restores exact behavior but *adds* a line to make set's list-conversion owner explicit — the opposite of the "drop redundant code" goal. The original `isinstance(obj, (set, tuple))` was correct and clear, so the right call is to abandon this marginal cleanup rather than relocate logic to defend it.

The rest of the stack (#3354 S3, #3355 S6, #3356 S1, #3357 S4, #3358 S5) stands.

PR #3362, opened 2026-08-10
perf(py): collapse redundant attribute lookups in _serialize_json probe loop

emil-lc:
## What

Micro-cleanup of the `_serialize_json` probe loop (the orjson `default` hook on the tracing serialization path). Per method probe it did **three** attribute accesses — `hasattr(obj, attr)`, `callable(getattr(obj, attr))`, then `getattr(obj, attr)` again — and re-evaluated the loop-invariant `isinstance(obj, type)` on every iteration.

Now:
```python
if isinstance(obj, type):          # invariant — checked once, hoisted out
    return _simple_default(obj)
for attr, kwargs in _serialization_methods:
    method = getattr(obj, attr, None)   # one lookup per probe
    if callable(method):
        ...
```
Also drops a dead `pass`.

## Behavior

Unchanged. Verified `pydantic model → dict`, `namedtuple → field names preserved`, `plain tuple/set → array`, `class object → str`, and `to_dict → dict` all serialize identically to `main`. Added a small regression test locking the hoisted class-object case; existing `dumps_json` tests cover the rest (3 passed).

### DIFF
--- a/python/tests/unit_tests/test_client.py
+++ b/python/tests/unit_tests/test_client.py
@@ -2436,6 +2436,12 @@ def test__dumps_json_does_not_swallow_keyboard_interrupt():
         _dumps_json({"x": _RaisesInterrupt()})
 
 
+def test__dumps_json_type_object_serializes_as_str():
+    # A class object has no useful instance serialization method; it must fall
+    # through to _simple_default -> str rather than be probed like an instance.
+    assert isinstance(_orjson.loads(_dumps_json({"cls": dict}))["cls"], str)
+
+
 @patch("langsmith.client.requests.Session", autospec=True)
 def test_host_url(_: MagicMock) -> None:
     client = Client(api_url="https://api.foobar.com/api", api_key="API_KEY")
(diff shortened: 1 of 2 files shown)


python/tests/unit_tests/test_client.py
```diff
@@ -2420,6 +2420,12 @@ def to_dict(self) -> Dict:
     }
 
 
+def test__dumps_json_type_object_serializes_as_str():
```
KiewanVillatel, line 2439: From Claude:

The new regression test doesn't actually test anything — it passes identically with the
`isinstance(obj, type)` guard deleted. `dict` has no `model_dump` / `dict` / `to_dict`
attribute, so it never enters the probe loop either way; the test only exercises
`_simple_default`'s `str()` fallback, which was never at risk.

To pin the hoisted behavior, the subject needs a callable probe attribute *at the class
level*. A metaclass method works; a plain `@classmethod to_dict` doesn't (the `dict` /
`model_dump` probes short-circuit first, so it returns `str` even unguarded):

```python
def test__dumps_json_type_object_serializes_as_str():
    class Meta(type):
        def to_dict(cls):
            return {"meta": "yes"}

    class WithClassLevelToDict(metaclass=Meta):
        pass

    # Guarded: falls through to _simple_default -> str.
    # Unguarded: probed like an instance, emits {"meta": "yes"}.
    assert isinstance(
        _orjson.loads(_dumps_json({"cls": WithClassLevelToDict}))["cls"], str
    )

KiewanVillatel (approved):

python/tests/unit_tests/test_client.py
emil-lc: Ack. Updated test accordingly.

PR #3367, opened 2026-08-11
perf(py): call pydantic core serializer directly, skip orjson key checks

emil-lc:
## What & why

Two changes to the tracing serialization path. Both keep Pydantic's json-mode output exactly as it is today.

**1. Call the Pydantic core serializer directly.** `model_dump()` is a Python wrapper that processes its keyword arguments and then calls the model's fast, compiled serializer. `_serialize_json` now calls that compiled serializer directly. Our 132msg repro reaches this hook ~2,000 times per traced invoke, so the savings add up.

Only used where it is safe:
- the object is a Pydantic v2 model whose class did not override `model_dump()`. Everything else falls through to the existing `_serialization_methods` list, unchanged.
- the lookup is cached per **class** (not per instance) in a dict capped at 1024 entries.
- if the serializer raises, the class is marked ineligible just once.

**2. Drop `OPT_NON_STR_KEYS` from `dumps_json`'s first attempt.** That option makes orjson check the type of **every** dict key. Non-`str` keys now raise `TypeError`, and the same payload is retried with the full options, so orjson coerces those keys exactly as it did before. Only the keys orjson cannot coerce itself go on to `_normalize_json_keys`. Payloads with non-`str` keys serialize twice; payloads without them skip the check.

## Measured improvement

**What is measured:** `dumps_json()` on the 77 payloads it receives during one traced invoke of a 132-message agent thread — 2,126 `default` hook calls, 2,027 of them Pydantic models. Serialization only: no multipart, compression, HTTP or background threads.

**How:** main's `_serde.py` and this branch's are loaded as two modules in one process and run on the same payload objects, alternating rep by rep.

| metric | main | this PR | delta |
|---|---|---|---|
| CPU per pass | 7.111 ms | **5.745 ms** | **−19.2%** |
| Allocations per pass (memray) | 29,424 | 29,420 | −0.0% |
| Peak allocated per pass (tracemalloc) | 1034.2 KiB | 1030.0 KiB | −0.4% |

Output is unchanged: identical bytes on all 77 payloads, and the same 2,126 hook calls.

Python 3.13.10, pydantic 2.13.4, orjson 3.11.9, langchain-core 1.4.9.

## Tests

- `test__dumps_json_honors_json_only_field_serializer` — a field redacted with `@field_serializer(when_used="json")` stays redacted, including nested behind `Any`, in a dict and in a list. Custom serializers live in the core schema, so calling that serializer directly still applies them.
- `test__dumps_json_honors_model_dump_override` — a class that overrides `model_dump()` keeps control of its serialization.
- `test__dumps_json_coerces_json_native_dict_keys` — `int`/`float`/`bool`/`None`/`date` keys still produce the same strings. They now reach the retry, so the output needed pinning.
- `test__dumps_json_coerces_enum_dict_keys_by_value` — `Enum` keys still serialize as their value. Without the retry they would arrive as `"Color.RED"` instead of `"red"`.

### DIFF
--- a/python/langsmith/_internal/_serde.py
+++ b/python/langsmith/_internal/_serde.py
@@ -263,6 +263,15 @@ def dumps_json(obj: Any) -> bytes:
     except TypeError as e:
         # Usually caused by UTF surrogate characters or non-str dict keys
         logger.debug(f"Orjson serialization failed: {repr(e)}. Falling back to json.")
+        try:
+            # Let orjson coerce non-str keys. Only stringify the ones it can't handle.
+            return _orjson.dumps(
+                obj,
+                default=_serialize_json,
+                option=_ORJSON_OPTIONS,
+            )
+        except TypeError:
+            pass
         normalized_obj = _normalize_json_keys(obj)
         try:
             return _orjson.dumps(
--- a/python/langsmith/_internal/_serde.py
+++ b/python/langsmith/_internal/_serde.py
@@ -12,6 +12,8 @@ import re
 import uuid
 from typing import Any
 
+from pydantic import BaseModel
+
 from langsmith._internal import _orjson
 
 try:
@@ -29,6 +31,9 @@ _ORJSON_OPTIONS = (
     | _orjson.OPT_SERIALIZE_UUID
     | _orjson.OPT_NON_STR_KEYS
 )
+# Turn off OPT_NON_STR_KEYS, trading better speed in the general case where
+# all dict keys are strings for worse speed in the exceptional case
+_ORJSON_OPTIONS_FAST = _ORJSON_OPTIONS & ~_orjson.OPT_NON_STR_KEYS
 _JSON_KEY_TYPES = (str, int, float, bool, type(None))
 # Matches escaped lone UTF-16 surrogates (e.g. b"\\ud800") in ensure_ascii
 # json.dumps output; used to strip them on the stdlib-json fallback path.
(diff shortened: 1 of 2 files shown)


python/langsmith/_internal/_serde.py
```diff
@@ -85,6 +90,49 @@ def _simple_default(obj):
...
 
 
+_MISSING = object()
+# Maps a Class to its Pydantic core serializer, or to None when the fast path
+# below does not apply to it. Bounded rather than weakly keyed.
+_PYDANTIC_SERIALIZER_CACHE_MAX = 1024
+_pydantic_core_serializers: dict[type, Any] = {}
+
+
+def _remember_pydantic_serializer(cls: type, serializer: Any) -> None:
```
QuentinBrosse, line 100: nit: will it still work it we simplify with:


```py
@functools.lru_cache(maxsize=1024)
def _pydantic_core_serializer(cls: type) -> Any:
    """Core serializer, for models that didn't override the model_dump() wrapper."""
    if issubclass(cls, BaseModel) and cls.model_dump is BaseModel.model_dump:
        # getattr: a deferred-schema model has a placeholder that raises on use.
        return getattr(cls, "__pydantic_serializer__", None)
    return None
```

Then in `_serialize_json`, after the `isinstance(obj, type)` check:
```py
        # model_dump() is a Python wrapper over this; calling it direct is cheaper.
        serializer = _pydantic_core_serializer(type(obj))
        if serializer is not None:
            try:
                return serializer.to_python(
                    obj, mode="json", exclude_none=True, warnings=False
                )
            except Exception:  # fall through to _serialization_methods
                pass
```

QuentinBrosse (approved):

python/tests/unit_tests/test_client.py
QuentinBrosse, line 1: Should we include it in the benchmark? idk why it's not posted in the PR, but I see an x1.15 faster [here](https://github.com/langchain-ai/langsmith-sdk/actions/runs/31593351997/job/94103072224?pr=3367)

python/langsmith/_internal/_serde.py
emil-lc: The code is indeed similar to @lru_cache, but with a twist. The tricky part is that:
 - if we try to do the fast `pydantic_serializer` based on `cls` (what the lru_cache would capture)
   - because the class technically supports that method
      - BUT the instance of the class throws an exception (like it happens with `StructuredTool`-s from Langchain)
        - then we can't easily update the lru_cache from the outside (to change its desired result) and we'd keep on paying the exception cost for the rest of the payload
 
 We can counteract with yet another accounting structure, but the only result would be (slightly) more complexity.
 
 Here are some measurements:
 ```
 ┌──────────────────────────┬──────────────────────┬──────────────────────┬──────────────────────┐
│         payload          │ PR (dict + remember) │ direct lru_cache │ lru_cache + fail set │
├──────────────────────────┼──────────────────────┼──────────────────────┼──────────────────────┤
│ messages (all hits)      │ 0.369–0.376 ms       │ 0.353–0.362 ms       │ 0.355–0.365 ms       │
├──────────────────────────┼──────────────────────┼──────────────────────┼──────────────────────┤
│ tools (json dump raises) │ 0.130–0.135 ms       │ 0.187–0.195 ms       │ 0.130–0.131 ms       │
├──────────────────────────┼──────────────────────┼──────────────────────┼──────────────────────┤
│ raises per call, tools   │ 84                   │ 168                  │ 84                   │
└──────────────────────────┴──────────────────────┴──────────────────────┴──────────────────────┘
```

File: python/tests/unit_tests/test_client.py
Comment (emil-lc) [reply python/tests/unit_tests/test_client.py:1]: TBH, I don't think it's representative due to the noise on those benchmarks. 

For example, the "gain" line is showing up there as:
```
dumps_pydantic_nested_50x100                  | 60.3 ms  | 52.3 ms: 1.15x faster
```

So, it sounds like we dropped from 60> 52ms, or an improvement of 8ms, right? Here is what you see in the body, though:
```
dumps_pydantic_nested_50x100: Mean +- std dev: 52.3 ms +- 10.7 ms
```
One standard-deviation interval is already 10ms - larger than the recorded gain of 8ms. So, from a statistical perspective, it might as well all be noise. The benchmark is too noisy to reliably measure this amount of improvement.

PR #3371, opened 2026-08-11
fix(py): mask MCP server credentials in Anthropic wrapper traces

emil-lc:
## What & why

`wrap_anthropic` copied Anthropic MCP server definitions into traces verbatim, so sensitive credentials could get uploaded to LangSmith in two places:

- `run.inputs`
- `run.extra.metadata.ls_invocation_params`

## Fix

`_strip_not_given` now masks `mcp_servers` entries. Keys outside an allowlist (`name`, `type`, `url`, `tool_configuration`) keep their name but get `SECRET_PLACEHOLDER` as their value, so a trace still shows the field was set without exposing it, and any potentially sensitive field that Anthropic would add later is masked by default -rather than exported.

`_infer_ls_params` calls `_strip_not_given`, so this single place covers both exposure paths.

## Tests

`TestRedactMCPServers` covers: token masked in `run.inputs`, token masked in `ls_invocation_params`, the caller's dict still holds the real token after both calls, an unknown field is masked, and unexpected shapes (non-list, non-dict entries, empty list, tuple) don't raise.

Also verified end to end against a live backend: the same `beta.messages.create` call traced before and after the fix, read back from the API, shows the token in both inputs and metadata before, and `[SECRET_DETECTED]` in both after.

### DIFF
--- a/python/langsmith/wrappers/_anthropic.py
+++ b/python/langsmith/wrappers/_anthropic.py
@@ -18,6 +18,7 @@ from typing_extensions import Self, TypedDict
 from langsmith import client as ls_client
 from langsmith import run_helpers
 from langsmith._internal._orjson import dumps as _dumps
+from langsmith.anonymizer import SECRET_PLACEHOLDER
 from langsmith.schemas import InputTokenDetails, UsageMetadata
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


jacoblee93 (approved):

python/langsmith/wrappers/_anthropic.py
```diff
@@ -40,6 +41,36 @@ def _get_not_given() -> Optional[tuple[type, ...]]:
         return None
 
 
+# Keys of Anthropic's MCP server definition that are known to be safe to trace
+_MCP_SERVER_SAFE_KEYS: frozenset[str] = frozenset(
+    {"name", "type", "url", "tool_configuration"}
```
QuentinBrosse, line 46: FYI I double-checked, and we shouldn't have credentials in URL ([specs](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization)) so it's ok to keep in clear in the allow list 👍

QuentinBrosse (approved):

PR #3378, opened 2026-08-12
fix(js): close the argument-shape bypass in Anthropic MCP redaction

emil-lc:
## What & why

#3372 masks `mcp_servers` credentials by reading `params.mcp_servers`. That lookup only fires when the params object *is* the traced input, which holds only for a single-argument call.

`traceable` records any call with a second argument as `{ args: [params, requestOptions] }` (`runInputsToMap`), and `argsConfigPath` leaves `args[1]` in place after extracting `langsmithExtra`. So the documented form

```js
messages.create(params, { langsmithExtra: { metadata: { ... } } })
```

skips redaction and uploads `authorization_token` in the clear. Verified against the payload the SDK would send:

```
inputs: {"args":[{...,"mcp_servers":[{...,"authorization_token":"<token>"}]},{}]}
```

`wrapped_anthropic_mcp.test.ts` passed throughout, because its `traceCreate` helper only ever passed one argument.

`extra_headers` / `extra_body` and the `headers` on the second-argument `RequestOptions` reach `run.inputs` by the same route, so they are masked here too.

## Fix

New `js/src/utils/redaction.ts` holds the primitives the remaining wrappers reuse:

- `mask` — replace a value, keeping object key names
- `redactOutside` — the #3371 / #3372 allowlist primitive
- `overParams` — apply a redaction function to every object argument

`redactMcpServers` keeps its allowlist and delegates to `redactOutside`. Single-argument behavior is unchanged.

## Tests

Assertions run against the payload the SDK would upload, via the existing `mockClient()` fetch spy — no network, no API keys.

Added: masking holds when called with `langsmithExtra`, an empty options object, and real request options; credentials in the request options are masked; both caller arguments keep their real values; `extra_headers` / `extra_body` / `extra_query` keep their key names but not their values.

`traceCreate` now takes optional request options.

### DIFF
--- a/js/src/utils/redaction.ts
+++ b/js/src/utils/redaction.ts
@@ -1,0 +1,49 @@
+import { SECRET_PLACEHOLDER } from "../anonymizer/index.js";
+import type { KVMap } from "../schemas.js";
+
+function isPlainObject(value: unknown): value is Record<string, unknown> {
+  return typeof value === "object" && value !== null && !Array.isArray(value);
+}
+
+/** Mask a value, keeping object key names so a trace shows what was set. */
+export function mask(value: unknown): unknown {
+  if (isPlainObject(value)) {
+    return Object.fromEntries(
+      Object.keys(value).map((key) => [key, SECRET_PLACEHOLDER]),
+    );
+  }
+  return SECRET_PLACEHOLDER;
+}
+
+/** Copy `entry`, masking values outside `safeKeys`, so unknown keys fail closed. */
+export function redactOutside(
+  entry: unknown,
+  safeKeys: ReadonlySet<string>,
+): unknown {
+  if (!isPlainObject(entry)) return entry;
+  return Object.fromEntries(
+    Object.entries(entry).map(([key, value]) => [
+      key,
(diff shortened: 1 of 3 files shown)


js/src/utils/redaction.ts
```diff
@@ -0,0 +1,49 @@
+import { SECRET_PLACEHOLDER } from "../anonymizer/index.js";
+import type { KVMap } from "../schemas.js";
+
+function isPlainObject(value: unknown): value is Record<string, unknown> {
+  return typeof value === "object" && value !== null && !Array.isArray(value);
+}
+
+/** Mask a value, keeping object key names so a trace shows what was set. */
+export function mask(value: unknown): unknown {
```
jacoblee93, line 9: Does this need to recurse into arrays too?

jacoblee93 (approved):

js/src/utils/redaction.ts
emil-lc: Nah - that would just result in having `"array": ["redacted", "redacted", ...]` vs `"array": "redacted"`. No real improvement, and I'd argue it looks even sillier.

PR #3379, opened 2026-08-12
fix(py,js): mask hosted MCP tool credentials in the OpenAI wrappers

emil-lc:
Stacked on #3378 — review that one first; it adds the JS redaction helpers this PR reuses.

## What & why

The OpenAI Responses API accepts hosted MCP tool entries that carry an OAuth access token and arbitrary auth headers (`openai.types.responses.tool_param.Mcp`):

```json
```

Neither wrapper filtered them:

- **Python** — `_process_inputs` only stripped `NotGiven` sentinels (`_openai.py:57-87`)
- **JS** — no `processInputs` registered on any traceable config

So both uploaded the credential verbatim in `run.inputs` on every traced call. `extra_headers` / `extra_body` / `extra_query` reached `run.inputs` through the same path whenever the caller set them — the follow-up #3372 flagged.

```
 "headers":{"Authorization":"Bearer <token>"}}],
 "extra_headers":{"Authorization":"Bearer <token>"}}
```

After: `[SECRET_DETECTED]` in all three places, with the caller's own objects still holding the real values.

The metadata path was already safe — `_infer_invocation_params` has an allowlist (`_openai.py:105-127`) that excludes `tools` and `extra_*`. Tests pin that so it cannot regress.

## Fix

Only `type == "mcp"` tool entries are redacted, against an allowlist derived from the real `Mcp` TypedDict. **Function tools keep their JSON schema**, which is the useful part of a trace — a blanket redaction of `tools` would have gutted it.

New `python/langsmith/_internal/_redaction.py` mirrors the JS helpers from #3378: `mask`, `redact_outside`, `redact_keys`. `redact_keys` is unused here and lands ready for the Gemini follow-up, where google-genai scatters credentials across `http_options.headers`, `tools[].mcp_servers[].streamable_http_transport.headers`, `tools[].google_maps.auth_config`, and `tools[].retrieval.external_api.api_auth` — a per-site allowlist there would need a new patch every release.

The JS side routes through `overParams` from #3378, so the argument-shape bypass fixed there cannot reappear, and the second-argument `RequestOptions.headers` is masked too.

### Note on `server_url`

Kept visible, matching #3371's treatment of Anthropic's `url`. A token embedded in the URL still reaches the trace. Happy to mask it if you'd rather be strict — it's a one-line change to the allowlist in both SDKs.

## Tests

Python asserts on `_process_inputs` and `_infer_invocation_params`; JS asserts against the uploaded payload via the existing `mockClient()` fetch spy. No network, no API keys.

Both cover: credentials masked, allowed fields survive, an unknown field masked by default, function-tool schemas untouched, caller objects unmutated, transport overrides masked, and unexpected shapes (non-list, null entries, nested arrays, empty, tuple) not raising. JS additionally covers the two-argument call shape.

## Follow-ups (not here)

Still unremediated, from the same audit: both Gemini wrappers (`config.http_options.headers`, MCP tool headers, `google_maps.auth_config`), the OpenAI Realtime integration (`session.tools[].authorization` in span outputs), and `js/src/experimental/anthropic` (`options.env`, whose documented usage is `{ ...process.env }`).

### DIFF
--- a/js/src/tests/wrapped_openai_mcp.test.ts
+++ b/js/src/tests/wrapped_openai_mcp.test.ts
@@ -1,0 +1,192 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { wrapOpenAI } from "../wrappers/openai.js";
+import { SECRET_PLACEHOLDER } from "../anonymizer/index.js";
+import { mockClient } from "./utils/mock_client.js";
+
+const FAKE_TOKEN = "fake-token-for-tests-only";
+
+function parseRequestBody(body: any) {
+  // eslint-disable-next-line no-instanceof/no-instanceof
+  return body instanceof Uint8Array
+    ? JSON.parse(new TextDecoder().decode(body))
+    : JSON.parse(body);
+}
+
+/** Stub provider, so input processing runs without a network call. */
+function stubOpenAI(): any {
+  const response = { id: "resp_stub", output: [], usage: {} };
+  return {
+    responses: {
+      create: async () => response,
+      parse: async () => response,
+    },
+    chat: { completions: { create: async () => ({ id: "cc_stub" }) } },
+    completions: { create: async () => ({ id: "c_stub" }) },
+    beta: { chat: { completions: {} } },
+  };
+}
+
+/** The runs the SDK would have uploaded. */
+function uploadedRuns(callSpy: any): any[] {
+  return (callSpy.mock.calls as any[])
+    .map(([, init]: any) => {
+      try {
+        return parseRequestBody(init?.body);
+      } catch {
+        return undefined;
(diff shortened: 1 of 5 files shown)


emil-lc:
Superseded — split by SDK, per review request:

- **#3380** — `fix(py)`: Python half. Independent of #3378, based on `main`, can merge on its own.
- **#3381** — `fix(js)`: JS half. Stacked on #3378, which it depends on for `overParams`.

Same commit contents as this PR, no code changes. Only the Python half was carrying a false dependency on the JS stack, so it is now based on `main` directly.

Branch `fix/openai-hosted-mcp-credentials` left in place for now; say the word and I'll delete it.

PR #3380, opened 2026-08-12
fix(py): mask hosted MCP tool credentials in the OpenAI wrapper

emil-lc:
## What & why

The OpenAI Responses API accepts hosted MCP tool entries that carry an OAuth access token and arbitrary auth headers (`openai.types.responses.tool_param.Mcp`):

```json
```

`_process_inputs` only strips `NotGiven` sentinels, so the wrapper uploads the credential verbatim in `run.inputs` on every traced call. `extra_headers` / `extra_body` / `extra_query` reach `run.inputs` by the same route whenever the caller sets them — the follow-up #3372 flagged.

Same bug class as #3371, different provider field.

```
 "headers":{"Authorization":"Bearer <token>"}}],
 "extra_headers":{"Authorization":"Bearer <token>"}}
```

After: `[SECRET_DETECTED]` in all three places, with the caller's own objects still holding their real values.

The metadata path is already safe — `_infer_invocation_params` has an allowlist that excludes `tools` and `extra_*`. Tests pin that so it cannot regress.

## Fix

Only `type == "mcp"` tool entries are redacted, against an allowlist derived from the `Mcp` TypedDict. **Function tools keep their JSON schema**, which is the useful part of a trace — a blanket redaction of `tools` would gut it.

New `langsmith/_internal/_redaction.py` holds the two shared primitives: `mask` (replace a value, keeping key names) and `redact_outside` (allowlist a known container, so unknown keys fail shut). #3382 reuses both.

`server_url` stays visible, matching #3371's treatment of Anthropic's `url`, but that should be a very niche option (URL-based authentication)

## Tests

`TestRedactHostedMCPTools` and `TestRedactTransportOverrides`. No network, no API keys. Cover: credentials masked in `run.inputs`, never present in `ls_invocation_params`, allowed fields survive, an unknown field masked by default, function-tool schemas untouched, the caller's dicts unmutated, and unexpected shapes (non-list, null entries, nested arrays, empty, tuple) not raising.

## Stack

Python: this → #3382 (Gemini). Based on `main`; independent of the JS stack.
JS counterpart: #3381.

### DIFF
--- a/python/langsmith/_internal/_redaction.py
+++ b/python/langsmith/_internal/_redaction.py
@@ -1,0 +1,47 @@
+"""Mask credentials in provider request params on the way into a trace.
+
+Helpers return new containers: the caller's objects also go to the provider API
+and must keep their real values.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Collection, Mapping
+from typing import Any, Optional
+
+from langsmith.anonymizer import SECRET_PLACEHOLDER
+
+__all__ = ["as_mapping", "mask", "redact_outside"]
+
+
+def as_mapping(value: Any) -> Optional[Mapping]:
+    """Return ``value`` as a mapping; Pydantic configs need a dump to see into."""
+    if isinstance(value, Mapping):
+        return value
+    dump = getattr(value, "model_dump", None)
+    if callable(dump):
+        try:
+            dumped = dump()
+        except Exception:
+            return None
+        return dumped if isinstance(dumped, Mapping) else None
+    return None
+
+
+def mask(value: Any) -> Any:
+    """Mask ``value``, keeping key names so a trace shows what was set."""
+    mapping = as_mapping(value)
+    if mapping is not None:
+        return {key: SECRET_PLACEHOLDER for key in mapping}
+    return SECRET_PLACEHOLDER
+
+
+def redact_outside(entry: Any, safe_keys: Collection[str]) -> Any:
+    """Copy ``entry``, masking values outside ``safe_keys``; unknown keys fail shut."""
+    mapping = as_mapping(entry)
+    if mapping is None:
+        return entry
+    return {
+        key: value if key in safe_keys else SECRET_PLACEHOLDER
+        for key, value in mapping.items()
+    }
--- a/python/langsmith/_internal/_redaction.py
+++ b/python/langsmith/_internal/_redaction.py
@@ -6,12 +6,15 @@ and must keep their real values.
 
 from __future__ import annotations
 
+import logging
 from collections.abc import Collection, Mapping
 from typing import Any, Optional
 
 from langsmith.anonymizer import SECRET_PLACEHOLDER
 
 __all__ = ["as_mapping", "mask", "redact_outside"]
+
+logger = logging.getLogger(__name__)
 
 
 def as_mapping(value: Any) -> Optional[Mapping]:
@@ -22,7 +25,10 @@ def as_mapping(value: Any) -> Optional[Mapping]:
     if callable(dump):
         try:
             dumped = dump()
-        except Exception:
+        except Exception as e:
+            logger.warning(
+                "Could not read %s to redact it: %s", type(value).__name__, e
+            )
             return None
         return dumped if isinstance(dumped, Mapping) else None
     return None
--- a/python/langsmith/wrappers/_openai.py
+++ b/python/langsmith/wrappers/_openai.py
@@ -24,6 +24,7 @@ from langsmith._internal._redaction import as_mapping, mask, redact_outside
 # integrations can reuse it without importing the ``wrappers`` package (whose
 # ``__init__`` warns at import time). Re-exported here for backwards compat.
 from langsmith._internal._usage import _create_usage_metadata
+from langsmith.anonymizer import SECRET_PLACEHOLDER
 
 if TYPE_CHECKING:
     from openai import AsyncOpenAI, OpenAI
@@ -90,6 +91,11 @@ _MCP_TOOL_SAFE_KEYS: frozenset[str] = frozenset(
 _TRANSPORT_SECRET_KEYS = ("extra_headers", "extra_body", "extra_query")
 
 
+def _is_model(value: Any) -> bool:
+    """Whether ``value`` claims to be a Pydantic model."""
+    return callable(getattr(value, "model_dump", None))
+
+
 def _redact_tools(tools: Any) -> Any:
     """Mask hosted MCP entries; function tools carry the caller's schema."""
     if not isinstance(tools, (list, tuple)):
@@ -98,9 +104,13 @@ def _redact_tools(tools: Any) -> Any:
     redacted: list[Any] = []
     for tool in tools:
         mapping = as_mapping(tool)
-        if mapping is not None and mapping.get("type") == "mcp":
-            tool = redact_outside(mapping, _MCP_TOOL_SAFE_KEYS)
-        redacted.append(tool)
+        if mapping is None:
+            # An unreadable model would reach the trace as a repr of its fields.
+            redacted.append(SECRET_PLACEHOLDER if _is_model(tool) else tool)
+        elif mapping.get("type") == "mcp":
+            redacted.append(redact_outside(mapping, _MCP_TOOL_SAFE_KEYS))
+        else:
+            redacted.append(tool)
     return redacted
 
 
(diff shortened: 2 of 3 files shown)


python/langsmith/_internal/_redaction.py
```diff
@@ -0,0 +1,48 @@
...
+"""
+
+from __future__ import annotations
+
+from collections.abc import Collection, Mapping
+from typing import Any, Optional
+
+from langsmith.anonymizer import SECRET_PLACEHOLDER
+
+__all__ = ["mask", "redact_outside"]
```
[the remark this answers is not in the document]
emil-lc: Correct — fixed, and it turned up a wider problem.

The trailing newline came from a regex-based deletion that left the blank line behind. `ruff format` fixes it, and `ruff format --check .` is now clean across all 240 files on both Python branches.

The root cause was mine: I had verified this file with `ruff check` (lint) and assumed that covered formatting. It doesn't — CI runs `ruff format --check` as a separate job, which is exactly what you flagged.

Checking that assumption everywhere found the same mistake on the JS side, worse. I had been formatting TypeScript with `prettier`, but this repo's CI runs `oxfmt` (`pnpm run format:check`). Three of the four JS PRs would have failed the format job:

- #3381 `src/wrappers/openai.ts`
- #3385 `src/wrappers/gemini.ts`
- #3386 `src/experimental/anthropic/index.ts`

And the two formatters disagree on union-type wrapping, so running `prettier --write` on a whole file had silently reformatted **pre-existing** lines that my changes never touched — union casts in `processChatCompletion`, `processGeminiOutputs`, `_getGeminiInvocationParams`, and the `toolCalls` index signature. That churn was sitting in the diffs as unrelated noise.

Running `oxfmt` reverted all of it to match `main`. Each PR's diff against its merge-base is now confined to the files it should touch, with no formatting churn.

Verified with the exact CI commands: `ruff format --check` and `ruff check` clean on both Python branches; `oxfmt --check`, `oxlint`, and `tsc --noEmit` clean on the JS leaf; full suites green at 1980 Python and 896 JS.

python/langsmith/_internal/_redaction.py
```diff
@@ -0,0 +1,47 @@
...
+def _as_mapping(value: Any) -> Optional[Mapping]:
+    """Return ``value`` as a mapping; Pydantic configs need a dump to see into."""
+    if isinstance(value, Mapping):
+        return value
+    dump = getattr(value, "model_dump", None)
+    if callable(dump):
+        try:
+            dumped = dump()
+        except Exception:
+            return None
```
QuentinBrosse, line 32: nit: do we need a warning here?

python/langsmith/wrappers/_openai.py
```diff
@@ -69,9 +70,54 @@ def _strip_not_given(d: dict) -> dict:
...
+    }
+)
+
+# Per-request transport overrides: credentials by construction.
+_TRANSPORT_SECRET_KEYS = ("extra_headers", "extra_body", "extra_query")
+
+
+def _redact_tools(tools: Any) -> Any:
+    """Mask hosted MCP entries; function tools carry the caller's schema."""
+    if not isinstance(tools, (list, tuple)):
```
QuentinBrosse, line 101: nit: should we use `abc.Sequence`?

python/langsmith/wrappers/_openai.py
```diff
@@ -69,9 +70,54 @@ def _strip_not_given(d: dict) -> dict:
         return d
 
 
+# Keys of `openai.types.responses.tool_param.Mcp` that are safe to trace.
```
QuentinBrosse, line 74: Should we use a deny list instead? I assume OpenAI will add non-secret fields more often than secret ones?

I agree the allow list is best for security, but does it also mean we need to patch our integrations every time a provider adds new fields?

python/langsmith/wrappers/_openai.py
```diff
@@ -69,9 +70,54 @@ def _strip_not_given(d: dict) -> dict:
...
+def _redact_tools(tools: Any) -> Any:
+    """Mask hosted MCP entries; function tools carry the caller's schema."""
+    if not isinstance(tools, (list, tuple)):
+        return tools
+
+    redacted: list[Any] = []
+    for tool in tools:
+        if isinstance(tool, Mapping) and tool.get("type") == "mcp":
+            tool = redact_outside(tool, _MCP_TOOL_SAFE_KEYS)
+        redacted.append(tool)
```
QuentinBrosse: From Devin:

python/langsmith/wrappers/_openai.py:R99-103

**Pydantic tool objects bypass MCP masking even though the helper supports them**

`redact_outside` / `_as_mapping` in `python/langsmith/_internal/_redaction.py:17-47` explicitly handle non-Mapping objects that expose `model_dump()` (Pydantic), but `_redact_tools` gates on `isinstance(tool, Mapping)` before calling it. A caller who round-trips a Pydantic MCP tool object (e.g. `openai.types.responses.tool.Mcp` taken from a previous response) back into `tools` would have it serialized by the OpenAI SDK's transform layer, yet it would reach `run.inputs` unmasked. Consider using `_as_mapping(tool)` and checking `.get("type") == "mcp"` on the dumped mapping so the Pydantic path is covered consistently.

python/langsmith/wrappers/_openai.py
```diff
@@ -69,9 +70,54 @@ def _strip_not_given(d: dict) -> dict:
...
+        "tunnel_id",
+        "require_approval",
+        "allowed_tools",
+        "allowed_callers",
+        "defer_loading",
+    }
+)
+
+# Per-request transport overrides: credentials by construction.
+_TRANSPORT_SECRET_KEYS = ("extra_headers", "extra_body", "extra_query")
```
QuentinBrosse, line 91: I had the same question as Devin:

**extra_body masking loses potentially useful non-secret request params**

`extra_body` is commonly used to pass provider-specific but non-secret parameters (e.g. vLLM/Azure extensions), not just credentials. Masking all of its values (keeping only key names) will remove that information from traces for those users. The PR states this is intentional ("credentials by construction"), but it is worth confirming the tradeoff against other wrappers' behavior, since Anthropic's equivalent fix only redacts `mcp_servers` and leaves transport overrides alone (`python/langsmith/wrappers/_anthropic.py:91-93`).

python/langsmith/wrappers/_openai.py
emil-lc: I think we should favor security in this case. No one's workflow would be broken or compromised if OpenAI adds a brand-new property and we aren't tracking it (and the property name would generally still appear, with a redacted value).

OTOH, if a new secret field is exposed, people using older SDK versions will leak secrets until we publish an update and address it. Granted, that scenario would be far less common, but it's also far more dangerous.

python/langsmith/wrappers/_openai.py
emil-lc: Good catch - fixed now.

python/langsmith/wrappers/_openai.py
emil-lc: Sequence would let in bytes + str, which we don't want to catch/iterate over. `Iterable` is a bit better, but it would include generators — something we'd better not consume outright.

python/langsmith/wrappers/_openai.py
emil-lc: That's a sticking point. In theory, each of those fields can carry parameters with arbitrary names at arbitrary depths - it's all up to the third-party MCP server. There are valid examples for all 3 places for credential passing - with headers being the most frequently used one.

For example, body can be used to pass credentials for internal services:
```
response = client.chat.completions.create(
    model="gpt-4o",
    messages=[{"role": "user", "content": "Fetch internal Jira ticket JIRA-101"}],
    extra_body={
        "mcp_tool_credentials": {
            "jira_mcp_server": {
                "user_email": "dev@company.com",
                "api_token": "JIRA_API_SECRET_TOKEN"
            }
        }
    }
)
```
Query string is also used by Azure gateways:
```
response = client.chat.completions.create(
    model="gpt-4o",
    messages=[{"role": "user", "content": "Run MCP file search"}],
    extra_query={
        # Direct query-based secret key required by APIM proxy route
        "mcp-gateway-key": "GATEWAY_SECRET_QUERY_KEY",
        "routing_version": "v2"
    }
)
```

python/langsmith/_internal/_redaction.py
emil-lc: Yeah, that makes sense. Added warning and switched None to a redacted value.

emil-lc:
Closing for the time being - considering a bespoke secrets container class instead.

PR #3381, opened 2026-08-12
fix(js): mask hosted MCP tool credentials in the OpenAI wrapper

emil-lc:
Stacked on #3378 — review that one first; it adds the `overParams` helper this PR depends on. Python counterpart is #3380, which is independent of both.

## What & why

The OpenAI Responses API accepts hosted MCP tool entries that carry an OAuth access token and arbitrary auth headers:

```json
```

`wrapOpenAI` registers **no `processInputs`** on any traceable config, so the caller's request params are recorded verbatim as `run.inputs` and the credential is uploaded on every traced call. `extra_headers` / `extra_body` / `extra_query`, and the `headers` on the second-argument `RequestOptions`, reach `run.inputs` by the same route.

Same bug class as #3372, different provider field.

```
 "headers":{"Authorization":"Bearer <token>"}}]}
```

After: `[SECRET_DETECTED]`, with the caller's own objects intact.

The metadata path is already safe — `getChatModelInvocationParamsFn` filters against `TRACED_INVOCATION_KEYS`, which excludes `tools`. Tests pin that.

## Fix

Only `type == "mcp"` tool entries are redacted, against an allowlist derived from `OpenAI.Responses.Tool.Mcp`. **Function tools keep their JSON schema**, which is the useful part of a trace.

Routed through `overParams` from #3378, so the argument-shape bypass fixed there cannot reappear and a credential in the `RequestOptions` is masked too.

Covers all nine wrapped methods, via the five config objects they share: `chatCompletionParseMetadata` (`chat.completions.create` / `.parse` / `.stream`, `beta.chat.completions.parse` / `.stream`), `completions.create`, and `responses.create` / `.parse` / `.stream`.

`server_url` stays visible, matching #3372's treatment of Anthropic's `url`. A token embedded in the URL still reaches the trace; one line in the allowlist if you'd rather be strict.

## Tests

`wrapped_openai_mcp.test.ts` asserts against the payload the SDK would upload, via the existing `mockClient()` fetch spy — no network, no API keys.

Covers: credentials masked, allowed fields survive, an unknown field masked by default, function-tool schemas untouched, masking survives a second argument, credentials in the request options masked, the caller's params unmutated, transport overrides masked, `ls_invocation_params` clean, and unexpected shapes (non-array, null entries, nested arrays, empty) not raising.

### DIFF
--- a/js/src/tests/wrapped_openai_mcp.test.ts
+++ b/js/src/tests/wrapped_openai_mcp.test.ts
@@ -1,0 +1,189 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { wrapOpenAI } from "../wrappers/openai.js";
+import { SECRET_PLACEHOLDER } from "../anonymizer/index.js";
+import { mockClient } from "./utils/mock_client.js";
+
+const FAKE_TOKEN = "fake-token-for-tests-only";
+
+function parseRequestBody(body: any) {
+  // eslint-disable-next-line no-instanceof/no-instanceof
+  return body instanceof Uint8Array
+    ? JSON.parse(new TextDecoder().decode(body))
+    : JSON.parse(body);
+}
+
+/** Stub provider, so input processing runs without a network call. */
+function stubOpenAI(): any {
+  const response = { id: "resp_stub", output: [], usage: {} };
+  return {
+    responses: {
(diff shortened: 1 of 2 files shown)


emil-lc:
Closing for the time being.

PR #3384, opened 2026-08-12
fix(py): mask credentials in traced voice event payloads

emil-lc:
Stacked on #3382. Completes the Python stream.

## What & why

`EventSession.event_span` writes the scrubbed provider event straight into a span's I/O when the adapter supplies no curated payload:

```python
run.end(outputs={} if inbound else payload)
```

OpenAI Realtime `session.created` and `session.updated` take that path — `_connection.py` spans every readable event, and `_span_kwargs` curates only transcription and `response.done`. So the entire session object, including `session.tools[].authorization` and `session.tools[].headers` on hosted MCP tools, is uploaded verbatim as `run.outputs`.

`scrub` is the only sanitizer in that path, and it only replaces `bytes` and truncates long strings.

## Fix

`scrub` now masks values under a set of keys that are credentials by construction. It is the shared sink for every voice adapter, so one change covers **OpenAI Realtime, Pipecat, and Google ADK Live**.

A denylist rather than an allowlist, deliberately: adapters hand `scrub` arbitrary provider payloads, so unlike the Anthropic/OpenAI/Gemini MCP containers there is no shape to allowlist against. Matching is on **exact** key names — a substring rule on `token` would eat `max_tokens`, `token_count`, `input_tokens`, `top_logprobs`. A test pins that.

Mapping key names are preserved when masking, so a trace still shows *what* was set:

```python
{"env": {"ANTHROPIC_API_KEY": "[SECRET_DETECTED]"}}
```

## Not covered — filing separately

In `integrations/openai_realtime/_session.py`, `_clean` bails to `_shallow` at `_MAX_DEPTH`, and `_shallow` returns a truncated `repr()`. A credential nested past that depth reaches the trace as text inside that string, which `scrub` cannot see into. That is the medium-severity `handoff` / `tool_approval_required` finding from the audit; it needs its own change at the repr boundary.

## Tests

Realtime `session.updated` carrying a hosted MCP tool: token absent from the scrubbed payload, `authorization` and `headers` masked, `server_label` and `model` preserved.

Plus: each credential key masked, mapping key names kept, lookalike keys untouched, masking inside lists, and the caller's event left unmutated.

## Reproducing locally

A set `LANGSMITH_ENDPOINT` or `OTEL_EXPORTER_OTLP_HEADERS` makes ~20 unrelated tests fail in `test_utils.py` / `test_run_helpers.py` / `test_otel_exporter.py` — they patch the `LANGCHAIN_*` equivalents, which the `LANGSMITH_*` names take precedence over. Unrelated to this PR.

### DIFF
--- a/python/langsmith/_internal/_redaction.py
+++ b/python/langsmith/_internal/_redaction.py
@@ -1,0 +1,71 @@
+"""Mask credentials in provider request params on the way into a trace.
+
+Helpers return new containers: the caller's objects also go to the provider API
+and must keep their real values.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Collection, Mapping
+from typing import Any, Optional
+
+from langsmith.anonymizer import SECRET_PLACEHOLDER
+
+__all__ = ["mask", "redact_outside", "redact_keys"]
+
+# Guard against cyclic payloads; request params are shallow.
+_MAX_DEPTH = 12
+
+
+def _as_mapping(value: Any) -> Optional[Mapping]:
+    """Return ``value`` as a mapping; Pydantic configs need a dump to see into."""
+    if isinstance(value, Mapping):
+        return value
+    dump = getattr(value, "model_dump", None)
+    if callable(dump):
+        try:
+            dumped = dump()
+        except Exception:
+            return None
+        return dumped if isinstance(dumped, Mapping) else None
+    return None
+
+
+def mask(value: Any) -> Any:
+    """Mask ``value``, keeping key names so a trace shows what was set."""
+    mapping = _as_mapping(value)
+    if mapping is not None:
+        return {key: SECRET_PLACEHOLDER for key in mapping}
+    return SECRET_PLACEHOLDER
+
+
+def redact_outside(entry: Any, safe_keys: Collection[str]) -> Any:
+    """Copy ``entry``, masking values outside ``safe_keys``; unknown keys fail shut."""
+    mapping = _as_mapping(entry)
+    if mapping is None:
+        return entry
+    return {
+        key: value if key in safe_keys else SECRET_PLACEHOLDER
+        for key, value in mapping.items()
+    }
+
+
+def redact_keys(value: Any, secret_keys: Collection[str], _depth: int = 0) -> Any:
+    """Recursively mask values under ``secret_keys``.
+
+    Scope to config-shaped data, never user content, where these key names are
+    legitimate.
+    """
+    if _depth > _MAX_DEPTH:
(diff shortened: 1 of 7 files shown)


emil-lc:
Withdrawing this for now — the approach is unsound, and the underlying leak needs a different shape. Branch `fix/py-voice-event-credentials` is kept if anyone wants to pick it up.

## The leak this was fixing (still open)

`EventSession.event_span` writes the scrubbed provider event straight into a span's I/O when the adapter supplies no curated payload:

```python
run.end(outputs={} if inbound else payload)
```

OpenAI Realtime `session.created` and `session.updated` take that path, so the whole session object — including `session.tools[].authorization` and `session.tools[].headers` on hosted MCP tools — is uploaded verbatim as `run.outputs`. `scrub` only replaces `bytes` and truncates long strings.

## Why this approach was wrong

It masked a set of credential key names recursively inside `scrub`. But Realtime `session.updated` echoes the session's tool declarations, and a function tool's JSON schema puts **arbitrary user-chosen names** under `properties`. Reproduced:

```json
"parameters": {
  "type": "object",
  "properties": {
    "headers":  { "type": "[SECRET_DETECTED]" },
    "api_key":  { "type": "[SECRET_DETECTED]" }
  }
}
```

A declared parameter named `headers` is not a credential, and the result is not a redacted trace — it is a schema asserting a type that does not exist. Same class of bug @open-swe caught in #3382 (https://github.com/langchain-ai/langsmith-sdk/pull/3382#discussion_r3767159104), which is now fixed by addressing known credential sites explicitly instead of walking.

There is no salvage by narrowing the key set: `headers`, `api_key`, `env`, `secret` are all legitimate function parameter names. And there is no stable subtree to skip — a skip-set is a denylist on a denylist, so it fails open again the moment a provider adds another schema-bearing field.

## What it should look like

Structural, at the one place the shape is known:

1. Revert `scrub` to bytes + truncation only.
2. Give `event_span` an optional `redact` callable, applied once to the payload. It is the shared sink, so the hook is reusable, and it covers `metadata.raw_event` as well as `outputs` — curating `outputs` alone is not enough, because the raw payload is still demoted into metadata and uploaded.
3. Have the Realtime adapter pass a redactor that allowlists `session.tools[]` entries with `type == "mcp"`, mirroring #3380 / #3381.

That narrows the fix to OpenAI Realtime, which is what the audit actually confirmed. The Pipecat and Google ADK Live coverage here was incidental; the ADK finding (`metadata.raw_event` carrying `actions.requested_auth_configs`) is medium severity and separate.

PR #3385, opened 2026-08-12
fix(js): mask credentials in the Gemini wrapper config

emil-lc:
Stacked on #3381. Python counterpart is #3382.

## What & why

`processGeminiInputs` destructures only `contents` and spreads everything else into the traced run inputs, so the caller's `config` is recorded verbatim. `@google/genai` accepts credentials at several sites inside it:

| Path | Credential |
|---|---|
| `config.httpOptions.headers` | `Authorization`, `x-goog-api-key` |
| `config.httpOptions.extraBody` | arbitrary |
| `config.tools[].mcpServers[].streamableHttpTransport.headers` | bearer token |
| `config.tools[].googleMaps.authConfig` | API key / OAuth token |
| `config.tools[].retrieval.externalApi.authConfig` | API key |
| `config.tools[].retrieval.externalApi.apiAuth` | API key (deprecated field) |
| `config.tools[].parallelAiSearch.apiKey` | API key |

The MCP transport headers are the `@google/genai` analogue of the Anthropic `mcp_servers[].authorization_token` bug fixed in #3372.

Two fallback branches — `contents` falsy, and `contents` neither string nor array — also return the caller's own object rather than a copy.

## Fix

Each credential site is addressed explicitly. Nothing else in `config` is walked.

- `httpOptions` → allowlist (`redactOutside`). A flat struct with known fields, so `headers` and `extraBody` are masked, and a credential field added to it later is masked by default.
- `mcpServers[].streamableHttpTransport` → allowlist, same reasoning.
- The remaining paths → masked outright.

`exaAiSearch.apiKey` is listed too. It does not exist in the pinned `@google/genai` 2.10.0 but does in the Python SDK; an absent path is skipped, so listing it early is harmless.

An earlier revision of this PR walked the whole `config` with a recursive denylist of credential key names. That was wrong: `config` also holds user-authored JSON schemas in `tools[].functionDeclarations[].parameters`, `responseSchema` and `responseJsonSchema`, and a declared parameter may legitimately be named `headers` or `apiKey`. The walk rewrote those schema nodes to `{"type": "[SECRET_DETECTED]"}` — not merely redacting the trace but corrupting it into a schema claiming a type that does not exist. Caught by @open-swe on the Python twin, #3382.

Addressing only the known sites makes that class of corruption structurally impossible rather than merely unlikely: a subtree that is never visited cannot be rewritten, whatever `@google/genai` adds later.

Paths are camelCase, matching the SDK's declared types. #3382 carries a test that walks the google-genai type graph and fails if a release adds a credential field the list does not cover; TypeScript types are erased at runtime, so that test guards the shared enumeration for both wrappers.

The normalization body is unchanged: renamed to `normalizeGeminiInputs` and wrapped, which covers both fallback paths and routes through `overParams` from #3378 so no argument shape can defeat masking.

## Tests

`wrapped_gemini_config.test.ts` asserts against the payload the SDK would upload, via the existing `mockClient()` fetch spy — no network, no API keys.

Covers each credential site masked; the allowlists rejecting unknown fields; `functionDeclarations` parameter schemas, `responseSchema`, `responseJsonSchema` and unknown tool variants surviving byte-for-byte; the caller's config unmutated; non-secret config preserved; user content not walked; `ls_invocation_params` clean; and unexpected shapes (null, string, number, non-array `tools`) not raising.

### DIFF
--- a/js/src/tests/wrapped_gemini_config.test.ts
+++ b/js/src/tests/wrapped_gemini_config.test.ts
@@ -1,0 +1,316 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { wrapGemini } from "../wrappers/gemini.js";
+import { SECRET_PLACEHOLDER } from "../anonymizer/index.js";
+import { mockClient } from "./utils/mock_client.js";
+
+const FAKE_TOKEN = "fake-token-for-tests-only";
+
+const authHeaders = () => ({ Authorization: `Bearer ${FAKE_TOKEN}` });
+
+function parseRequestBody(body: any) {
+  // eslint-disable-next-line no-instanceof/no-instanceof
+  return body instanceof Uint8Array
+    ? JSON.parse(new TextDecoder().decode(body))
+    : JSON.parse(body);
+}
+
+/** Stub provider, so input processing runs without a network call. */
+function stubGemini(): any {
+  const response = {
+    candidates: [{ content: { role: "model", parts: [{ text: "hi" }] } }],
+    usageMetadata: { promptTokenCount: 1, candidatesTokenCount: 1 },
+  };
+  return {
+    models: {
+      generateContent: async () => response,
+      generateContentStream: async () => response,
+    },
+  };
+}
+
+/** The runs the SDK would have uploaded. */
+function uploadedRuns(callSpy: any): any[] {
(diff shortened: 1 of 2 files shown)


emil-lc:
Closing for the time being.

PR #3386, opened 2026-08-12
fix(js): mask Claude Agent SDK option credentials in traces

emil-lc:
Stacked on #3385. Final PR in the JS stream.

## What & why

`processInputs` in the experimental Claude Agent SDK wrapper reduces `options.mcpServers` to `{ name, type }`, but passes every other option key through to the traced run inputs verbatim.

`QueryOptions.env` is the problem. The Claude Agent SDK's own documented usage is:

```js
env: { ...process.env, SOME_VAR: "value" }
```

so in normal use it holds every API key on the machine. `extraArgs` passes raw CLI flag values and can carry the same.

This surface was untouched by #3372, which covers only `js/src/wrappers/anthropic.ts`.

## Fix

`env` and `extraArgs` are masked with the shared `mask` helper, so key names survive and a trace still shows what was set:

```json
{ "env": { "ANTHROPIC_API_KEY": "[SECRET_DETECTED]", "PATH": "[SECRET_DETECTED]" } }
```

A denylist rather than an allowlist here, because the useful option keys (`model`, `maxTurns`, `allowedTools`, `systemPrompt`, `permissionMode`, …) are numerous and non-sensitive, while the credential-bearing ones are exactly two.

Safe to mutate `options` in place at that point — the enclosing `toJSON` already made a shallow copy — and a test confirms the caller's object is untouched.

## Tests

`claude_agent_sdk_secrets.test.ts` asserts against the payload the SDK would upload, via the existing `mockClient()` fetch spy — no subprocess, no network, no API keys.

Covers: `env` masked with key names kept, `extraArgs` masked, `mcpServers` still reduced to name and type (including that an `http` server's `headers` never lands), the caller's options unmutated, non-secret options preserved, and absent options not raising.

## Audit status

| Stream | PRs |
|---|---|
| Python | #3380 (OpenAI) → #3382 (Gemini) |
| JS | #3378 (Anthropic argument shape) → #3381 (OpenAI) → #3385 (Gemini) → this |

The remaining high-severity finding is the OpenAI Realtime session leak. #3384 attempted it and was withdrawn: the key-based masking it used corrupted function tool schemas in `session.tools[]`, the same way the first Gemini revision did. That PR records the leak, the reproduction, and the structural shape a fix needs.

Left for separate issues, all medium severity:

- `integrations/openai_realtime/_session.py` — `_clean` bails to `_shallow` past `_MAX_DEPTH`, which returns a `repr()` string that `scrub` cannot mask inside
- OTEL exporters promote `extra_headers` to a `langsmith.request.headers` span attribute
- Vercel middleware spreads `providerOptions` and `headers` into run inputs
- openai-agents copies `model_config` verbatim into invocation params

### DIFF
--- a/js/src/experimental/anthropic/index.ts
+++ b/js/src/experimental/anthropic/index.ts
@@ -6,6 +6,10 @@ import {
 import { StreamManager, type WrapClaudeAgentSDKConfig } from "./context.js";
 import { convertFromAnthropicMessage, mergeMessagesById } from "./messages.js";
 import type { SDKMessage, SDKUserMessage, QueryOptions } from "./types.js";
+import { mask } from "../../utils/redaction.js";
+
+/** Options that are credentials by construction: `env` is documented as `{ ...process.env }`. */
+const OPTION_SECRET_KEYS = ["env", "extraArgs"] as const;
 
 const WRAPPED_TOOL_SYMBOL = Symbol.for(
   "langsmith:claude_agent_sdk:wrapped_tool",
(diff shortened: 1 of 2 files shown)


emil-lc:
Closing for the time being.

PR #3392, opened 2026-08-13
fix(py): preserve integer precision above 2**64 in dumps_json fallback

### LINKED ISSUE #3391: Integers >= 2**64 are silently converted to floats in trace payloads (dumps_json orjson round-trip)
### Summary

Any Python `int` >= 2**64 in a traced run's inputs/outputs/metadata is silently converted to a float before it reaches LangSmith. The type changes, precision is lost, and nearby values collapse to the same payload: `2**64` and `2**64 + 1` serialize to byte-identical JSON, so distinct inputs become indistinguishable on the platform.

Values up to `2**64 - 1` (unsigned 64-bit max) are preserved exactly; the corruption starts at exactly `2**64`.

### Repro (langsmith 0.10.18, Python 3.13)

```python
import json

wei = 19_876_543_210_987_654_321  # ~19.88 ETH in wei

serialized = dumps_json({"inputs": {"balance_wei": wei}})
print(serialized)                              # b'{"inputs":{"balance_wei":1.9876543210987655e+19}}'
back = json.loads(serialized)["inputs"]["balance_wei"]
print(type(back).__name__, int(back) - wei)    # float 847

# collision: distinct values, identical payloads
assert dumps_json({"v": 2**64}) == dumps_json({"v": 2**64 + 1})
```

| input | output | exact? |
|---|---|---|
| `2**63 - 1` | `9223372036854775807` | yes |
| `2**64 - 1` | `18446744073709551615` | yes |
| `2**64` | `1.8446744073709552e+19` | no |
| `2**64 + 1` | `1.8446744073709552e+19` | no (identical to previous row) |
| `1234567890123456789012` | `1.2345678901234568e+21` | no, off by 14,868 |

`dumps_json` is the serializer for run payloads (`client.py` `_dumps_json(run_create)`), so this affects any traced value carrying a large integer: wei/token amounts, hashes or IDs stored as ints, high-precision counters.

### Root cause (`langsmith/_internal/_serde.py::dumps_json`)

1. orjson can't serialize ints outside 64 bits, so the fast path raises `TypeError` and the code correctly falls back to stdlib `json.dumps`, which produces the exact result (`b'{"v": 19876543210987654321}'`).
2. The fallback then round-trips that result through `orjson.loads` / `orjson.dumps`. The round-trip exists to detect lone UTF-16 surrogates (`orjson.loads` rejects them, triggering `_elide_surrogates`), but on success its output *replaces* the stdlib bytes, and `orjson.loads` parses out-of-range ints as floats. The correct result is computed and then thrown away.

For a big-int payload there are no surrogates at all, so the round-trip is pure collateral damage.

### Expected behavior

Integers that stdlib `json` can represent exactly should round-trip exactly. The surrogate check shouldn't rewrite payloads that don't contain surrogates.

### Suggested fix

Run the surrogate check only when escaped surrogates are actually present in the stdlib output, and use `orjson.loads` purely as a detector, keeping the stdlib bytes on success. Lone-surrogate elision behaves exactly as today. Have a patch with regression tests ready, will open a PR.

### System info

langsmith 0.10.18, Python 3.13.13, macOS arm64, SDK/Client tracing, hosted platform. Reproduced at current `main` (85fcdafe).

Vissirexa:
Fixes #3391.

`dumps_json` falls back to stdlib `json.dumps` for objects orjson can't serialize, which includes any int outside the 64-bit range. The fallback then round-tripped the correct stdlib output through `orjson.loads`/`orjson.dumps` to detect lone UTF-16 surrogates, and since `orjson.loads` parses out-of-range ints as floats, the exact result was thrown away. Every int >= 2**64 in a run payload lost precision, and nearby values collapsed to byte-identical JSON (`2**64` and `2**64 + 1` produce the same bytes).

```python
dumps_json({"v": 19_876_543_210_987_654_321})
# before: b'{"v":1.9876543210987655e+19}'  (off by 847)
# after:  b'{"v": 19876543210987654321}'
```

The change is scoped to the one mechanism:

- run the surrogate check only when escaped surrogates are present in the stdlib output (`_SURROGATE_RE` already matches exactly what `ensure_ascii` emits)
- use `orjson.loads` purely as a lone-surrogate detector and keep the stdlib bytes on success, instead of re-dumping through orjson

The success-path change also matters for payloads that contain both a non-BMP character and a big int: `ensure_ascii` escapes the character as a valid surrogate pair, `orjson.loads` succeeds, and the old re-dump would corrupt the int anyway. There's a regression test for that case too.

Lone-surrogate elision behaves exactly as before (existing `test__dumps_json` covers it), and the orjson fast path is untouched — the fallback only runs when orjson has already raised.

`make format`, `make lint`, `make tests` pass (1951 passed, 6 skipped; the mypy `langsmith_pyo3` import-untyped error also occurs on a clean checkout of main).

### DIFF
--- a/python/langsmith/_internal/_serde.py
+++ b/python/langsmith/_internal/_serde.py
@@ -227,10 +227,13 @@ def dumps_json(obj: Any) -> bytes:
             default=_serialize_json_with_normalized_keys,
             ensure_ascii=True,
         ).encode("utf-8")
-        try:
-            result = _orjson.dumps(
+        if _SURROGATE_RE.search(result):
+            # orjson.loads serves only to detect lone surrogates here (it
+            # rejects them, triggering the elision below). On success the
+            # stdlib bytes are kept as-is: re-dumping the parsed value through
+            # orjson would silently convert integers >= 2**64 to floats.
+            try:
                 _orjson.loads(result.decode("utf-8", errors="surrogateescape"))
-            )
-        except _orjson.JSONDecodeError:
-            result = _elide_surrogates(result)
+            except _orjson.JSONDecodeError:
+                result = _elide_surrogates(result)
         return result
--- a/python/tests/unit_tests/test_client.py
+++ b/python/tests/unit_tests/test_client.py
@@ -2395,6 +2395,29 @@ def test__dumps_json():
     assert "\\uDC00" not in serialized_str
 
 
+def test__dumps_json_preserves_ints_larger_than_64_bits():
+    """Integers >= 2**64 must survive the stdlib-json fallback path exactly.
+
+    orjson can't serialize them (TypeError -> stdlib fallback), and orjson.loads
+    parses them as floats, so the fallback's surrogate-detection round-trip must
+    be skipped when no surrogates are present or distinct values collapse to the
+    same float (e.g. 2**64 and 2**64 + 1 both become 1.8446744073709552e+19).
+    """
+    for value in (2**64, 2**64 + 1, 19_876_543_210_987_654_321, -(2**80)):
+        serialized = _dumps_json({"v": value})
+        deserialized = json.loads(serialized)["v"]
+        assert isinstance(deserialized, int)
+        assert deserialized == value
+    assert _dumps_json({"v": 2**64}) != _dumps_json({"v": 2**64 + 1})
+
+    # Non-BMP characters are escaped as valid surrogate pairs by ensure_ascii,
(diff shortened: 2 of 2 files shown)


yumuzi234 (approved):

Vissirexa:
Superseded by #3409, which carries this commit plus follow-ups for the anonymizer pre-pass in `client.py` and the integration test. That merged as b816b5b2, so this branch has nothing left to land. Closing.

Thanks for the review, @yumuzi234.

PR #3399, opened 2026-08-14
fix(serde): collapse plain container cycles instead of raising

### LINKED ISSUE #3393: dumps_json raises RecursionError on plain dict/list cycles
### Summary

`dumps_json` raises `RecursionError` when a run's inputs/outputs contain a plain `dict` or `list` cycle. The serializer is layered with fallbacks and clearly designed not to raise on weird user data (object cycles through `__dict__`-style types are collapsed to a string), but a cycle built out of plain containers escapes all of them. Deeply nested non-cyclic structures (~1000+ levels) fail the same way.

A dataclass cycle serializes fine, which is why the existing test coverage doesn't catch this.

### Repro (langsmith 0.10.18, Python 3.13)

```python

d = {}; d["self"] = d
dumps_json({"cyclic": d})     # RecursionError: maximum recursion depth exceeded

l = []; l.append(l)
dumps_json({"cyclic": l})     # RecursionError

# but the covered case works:
import dataclasses
@dataclasses.dataclass
class C:
    val: object = None
c = C(); c.val = c
dumps_json({"cyclic": c})     # b'{"cyclic":"C(val=...)"}'
```

### Root cause (`langsmith/_internal/_serde.py`)

1. The orjson fast path raises `TypeError: Recursion limit reached` on the cycle, which is caught — falling back is the designed behavior.
2. The fallback then calls `_normalize_json_keys(obj)`, which recursively walks dicts/lists in Python with no cycle or depth guard, and blows the interpreter stack before stdlib `json.dumps` (which would at least raise a catchable `ValueError: Circular reference detected`) is ever reached.
3. The `_normalize_json_keys` docstring says "Cycles are handled downstream by `_serialize_json` (which collapses them to `str`), not here" — but that only holds for objects routed through orjson's `default` hook. orjson serializes `dict` and `list` natively, so a plain container cycle never reaches the hook.

Related: #962 covered the JS side of circular references in trace data.

### Expected behavior

Consistent with how object cycles are already handled: degrade to something serializable (a placeholder string) rather than raising `RecursionError` out of the tracing path.

### Suggested fix

The narrow fix is a seen-set keyed on `id()` in `_normalize_json_keys` that collapses revisited containers to a placeholder, mirroring what `_serialize_json` effectively does for object cycles. That doesn't cover the deep-but-acyclic case, which would need a depth cap (or catching `RecursionError` at the `dumps_json` level and degrading).

Since this is more invasive than a one-line change and touches the normalization pass that all fallback traffic goes through, I'd like a maintainer's take on the preferred direction before opening a PR — happy to write it either way.

### System info

langsmith 0.10.18, Python 3.13.13, macOS arm64, SDK/Client tracing, hosted platform. Reproduced at current `main` (85fcdafe).

eeshsaxena:
Fixes #3393.

`dumps_json` is layered so it never raises on odd user data (an object cycle through `__dict__`-style types is collapsed to a string), but a cycle built out of plain containers slips past all of it:

```python

d = {}; d["self"] = d
dumps_json({"cyclic": d})   # RecursionError

l = []; l.append(l)
dumps_json({"cyclic": l})   # RecursionError
```

orjson rejects the cyclic dict/list (raising `TypeError`), so `dumps_json` drops to its fallback and hands the object to `_normalize_json_keys`, which walks plain containers recursively and therefore recurses forever, letting a `RecursionError` reach the caller. As the issue notes, a dataclass cycle serializes fine, which is why the existing coverage didn't catch this.

This threads a `_seen` set of the container ids on the current path through `_normalize_json_keys` and collapses a container that references itself to `str(obj)`, the same way object cycles are already collapsed. A subtree shared by sibling keys is not a cycle and is still serialized in full, since the ids are tracked per path (added on entry, removed on exit).

Cyclic `dict`/`list`/`deque` now produce valid JSON instead of raising. I added a regression test next to the existing object-cycle one; the serialization tests pass and `ruff` is clean.

One scope note: this addresses the cycle case. The deeply-nested but acyclic case the issue also mentions is a plain Python recursion-depth limit rather than a cycle, so I left it out of this change; happy to follow up if you'd like it handled here too.

### DIFF
--- a/python/langsmith/_internal/_serde.py
+++ b/python/langsmith/_internal/_serde.py
@@ -133,7 +133,7 @@ def _serialize_json(obj: Any) -> Any:
         return str(obj)
 
 
-def _normalize_json_keys(obj: Any) -> Any:
+def _normalize_json_keys(obj: Any, _seen: set[int] | None = None) -> Any:
     """Recursively stringify dict keys that orjson will reject.
 
     Walks ``dict``, ``list``, ``tuple`` and ``deque`` so that unsupported keys
@@ -141,8 +141,10 @@ def _normalize_json_keys(obj: Any) -> Any:
     are covered here even though they're only ever *values*: orjson serializes
     them natively (as arrays) and therefore never routes them through the
     ``default`` hook, so a bad-keyed dict nested inside one would otherwise
-    slip past normalization. Cycles are handled downstream by
-    ``_serialize_json`` (which collapses them to ``str``), not here.
+    slip past normalization. A container reachable from itself is collapsed to
+    ``str`` (mirroring how ``_serialize_json`` collapses object cycles) so a
+    plain ``dict``/``list`` cycle does not recurse forever; ``_seen`` tracks the
+    ids of the containers on the current path for that check.
 
     JSON object keys must ultimately be ``str``/``int``/``float``/``bool``/
     ``None``; other key types are stringified via ``_simple_default`` so they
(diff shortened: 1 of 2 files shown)


eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`fix(serde): collapse plain container cycles instead of raising`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining langsmith-sdk!

eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

PR #3402, opened 2026-08-14
fix(py): count skipped and errored tests in --langsmith-output pass rate

sineha-mani:
## Description

**The problem**

A test can end in more than two states. It can pass, fail, get skipped, crash before it even starts, or never start at all.

The scoreboard only did the math on passed and failed. Everything else fell out of the fraction entirely. So 3 passed and 7 skipped wasn't "3 out of 10" — it was "3 out of 3." A green 100%.

Then it got worse. A test only checked in to the scoreboard once it actually started running. So a test that crashed during setup never checked in — it wasn't just missing from the math, it was missing from the table. No row. No trace it ever existed.

And the plugin had muted pytest's own summary line at the bottom to keep the display clean. So the one thing that would have contradicted the scoreboard was silenced.

Three things stacked: wrong denominator, invisible tests, no second opinion.

**The fix**

Take attendance before the tests run, not after. pytest already knows the full list of tests the moment it collects them — before a single one executes. The fix grabs that roster up front and uses it as the denominator. Now it's 3 out of 10, because the 10 was known from the start.

Give the missing tests a row. Anything on the roster that never reported back gets listed anyway, labelled with why — skipped, error, or queued. Present instead of vanished.

Green means everyone passed. Not "everyone who bothered to report."

Unmute pytest's summary. This is the backstop for the case the table still can't cover: if every test in a file crashes during setup, no table gets drawn at all. Then pytest's own count is the only thing left telling you something went wrong.

Averages say what they cover. correctness: 0.5 (2/5) — so a score from 2 tests doesn't read like a score from 5.

Solves [LSE-2789](https://linear.app/langchain/issue/LSE-2789/langsmith-pytest-langsmith-output-shows-green-100percent-when-only)

### examples 

Screenshot 2026-08-14 at 4 47 56 PM

## Test Plan

Unit tests in `python/tests/unit_tests/test_pytest_plugin.py`.

### DIFF
--- a/python/langsmith/pytest_plugin.py
+++ b/python/langsmith/pytest_plugin.py
@@ -7,11 +7,12 @@ import os
 import time
 from collections import defaultdict
 from threading import Lock
-from typing import Any
+from typing import Any, Optional
 
 import pytest
 
 from langsmith import utils as ls_utils
+from langsmith.testing._internal import _get_test_suite_name
 from langsmith.testing._internal import test as ls_test
 
 logger = logging.getLogger(__name__)
@@ -41,9 +42,11 @@ def pytest_addoption(parser):
 def _handle_output_args(args):
     """Handle output arguments."""
     if any(opt in args for opt in ["--langsmith-output"]):
-        # Only add --quiet if it's not already there
-        if not any(a in args for a in ["-qq"]):
-            args.insert(0, "-qq")
+        # Only add --quiet if it's not already there. Use -q rather than -qq so
+        # pytest still prints its end-of-run counts; the live table only covers
+        # tests that reached a test suite, so it can't stand in for them.
+        if not any(a in args for a in ["-q", "-qq", "--quiet"]):
+            args.insert(0, "-q")
         # Disable built-in output capturing
         if not any(a in args for a in ["-s", "--capture=no"]):
             args.insert(0, "-s")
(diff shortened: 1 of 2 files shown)


IgorVoytyuk:
I reported the original issue by email and was told a fix had shipped, so I checked the released wheel. The pass-rate half is genuinely fixed in `langsmith` 0.12.1 — but 0.12.1 is a narrower change than this PR, and two of the three legs described here are not in the released code.

Released `langsmith/pytest_plugin.py` (0.12.1, sha256 `5354a9df…01025b`, byte-identical to the wheel member):

- lines 46-47 still insert `-qq`, and line 108 is still `return "", "", ""` — so pytest's own end-of-run counts stay suppressed. The `-q` change and the `hookwrapper` + `force_result` in this PR are absent.
- line 231 is `rate = passed_count / len(suite_statuses)` — the denominator fix, in a different implementation than this PR's.
- line 242 is still `f"{k}: {sum(v) / len(v)}"` — without this PR's `({len(v)}/{total_count})`.

None of `_collected_test_suite_name`, `unrun_statuses`, `collected_by_suite`, `total_count` or `force_result` appear in the released file.

The consequence of the missing feedback annotation, on 0.12.1, **with exit code 0**: ten `@pytest.mark.langsmith` tests that all pass; three log `log_feedback(key="correctness", score=1.0)`, four log a categorical `value=` instead (documented as a supported feedback value), three log nothing because the judge did not answer.

```
| Averages |  |  |  | 100% | correctness: 1.0 | 0.00s |
```

Green, and the `1.0` covers three tests of ten. The number three is printed nowhere in the output.

The mixed case puts both denominators in one row — three pass, four skip, three whose fixture raises:

```
| Averages |  |  |  | 30% | correctness: 1.0 | 0.00s |
```

`30%` is over 10; the `1.0` is over 3.

Control, so this is not reading a constant: same shape, three tests log `1.0`, three log `0.0`, four log a categorical value — the footer prints `correctness: 0.5`, over a denominator of 6.

Setup: real pytest 9.1.1 and released langsmith 0.12.1; only the HTTP client and the two dataset/experiment bootstrap calls are stubbed, since those need an API key. Every status string, count and average above is computed by the plugin; I read the numbers back out of `process_status` after the run rather than parsing the rendered table.

The fix is the one line already in this PR. It looks like it simply did not land, and something narrower shipped in its place.

PR #3341, opened 2026-08-05
feat(python): support httpx2 with httpx fallback [closes LSDK-448]

yumuzi234:
## Description
Prefer maintained `httpx2` across the Python SDK while retaining an `httpx` fallback for existing dependency-suppressed environments. A shared compatibility module keeps client, response, transport, timeout, and exception types on one backend. Tracks [LSDK-448](https://linear.app/langchain/issue/LSDK-448/feature-request-support-httpx2-maintained-fork-alongside-httpx).

`httpx2>=2,<3` is maintained by Pydantic, BSD-3-Clause licensed, actively released by multiple contributors, and has no known PyPI advisories; retaining required `httpx` or making `httpx2` optional would not let downstream installs remove the old package.

## Test Plan
- [x] Verify sync and async generated-client requests use `httpx2`
- [x] Verify built wheels import with default `httpx2` and isolated `httpx`-only dependencies
- [x] Run all 1,913 Python unit tests; Ruff passes (Mypy retains the pre-existing Python 3.14 `Executor.map` override error)

Made by [Open SWE](https://openswe.vercel.app/agents/f655c09f-d295-b94f-8f66-c36e957a28cd)

## References
- Plan: https://openswe.vercel.app/agents/f655c09f-d295-b94f-8f66-c36e957a28cd/plan

PR #3406, opened 2026-08-17
chore(py)!: swtich to httpx2 dependency while keeping httpx as a fallback

emil-lc:
Moves the Python SDK's runtime HTTP dependency from `httpx` to `httpx2`, Pydantic's maintained fork.

The generated client now emits its own backend-selection shim at `_openapi_client/_httpx.py`. It prefers `httpx2` and falls back to `httpx` if `httpx2` is not installed. Hand-written code imports `httpx` from that same shim instead of
repeating the selection, so the generated and hand-written halves of the SDK are aligned.

`httpx` stays in the dev group only. The shim's `TYPE_CHECKING` branch resolves against it, and the
test suite exercises the fallback path.

Closes LSDK-448. Replaces the prototype in #3341.

### Dependency change

```diff
-    "httpx>=0.23.0,<1",
+    "httpx2>=2,<3",
```

Net effect on an install is one extra package. `httpx` → `httpx2` and `httpcore` → `httpcore2` are
one-for-one swaps. `truststore` is new — it is a hard dependency of `httpcore2`.

### User agent

So we can see which backend clients actually resolve to during the rollout, the `User-Agent` header
now names it: `langsmith-py/<version>-httpx2`, or `langsmith-py/<version>-httpx` when the fallback is
taken. The suffix is the shim's module name, and it carries no backend version, so the header keeps
the same cardinality as before for metrics faceted on it. `langsmith.Client` already sent
`langsmith-py/<version>` on both its `requests` session and the generated clients, so the suffix
covers every request it makes. `langsmith.AsyncClient` set no `User-Agent` at all and leaked httpx's
own default (`python-httpx2/2.10.0`); it now sends the same header as the sync client.

---

## Caveats

All four caveats have the same cause: **`httpx2` is a fork of `httpx`, not a subclass.** The two
packages have classes and exceptions with the same names and the same behaviour, but no shared base
classes. So `isinstance` checks and `except` clauses that cross the boundary do not match. They fail
silently, because the names still resolve.

Examples below can be executed from `python/` after:

```bash
uvx uv@latest sync --all-extras --dev
```

### 1. Test mocks stop intercepting SDK traffic

`pytest-httpx`, `respx`, and `httpx.MockTransport` all patch old httpx's transport layer. The SDK now
speaks httpx2, so these mocks no longer see its requests.

The requests are not blocked and no error is raised. They are sent for real. In a test suite without
socket blocking, that means live calls to the LangSmith API.

**Example**

```python
# /tmp/repro_mock/test_mock.py
from pytest_httpx import HTTPXMock
from langsmith._openapi_client import Langsmith

def test_info_call_is_mocked(httpx_mock: HTTPXMock):
    httpx_mock.add_response(
        url="https://api.smith.invalid.example/api/v1/info",
        json={"version": "1.2.3"},
    )
    client = Langsmith(api_key="fake-key", base_url="https://api.smith.invalid.example")
    assert client.info.list().version == "1.2.3"
    client.close()
```

```bash
.venv/bin/python -m pytest /tmp/repro_mock/test_mock.py -q -p pytest_httpx
```

```
langsmith._openapi_client.APIConnectionError: Connection error.
1 failed, 1 error
```

The mock was skipped and httpcore2 tried to open a real socket to a domain that does not exist.

**Workaround** `httpx2.alias_httpx()` makes `import httpx` return `httpx2` for the whole process.
Add it to a `sitecustomize.py` on the test run's `PYTHONPATH`:

```python
# tests/httpx2_alias/sitecustomize.py
import httpx2

httpx2.alias_httpx()
```

```bash
PYTHONPATH=tests/httpx2_alias .venv/bin/python -m pytest /tmp/repro_mock/test_mock.py -q -p pytest_httpx
```

Existing `pytest-httpx` tests then work unchanged. This PR uses exactly this approach for the SDK's
own test suite — see `python/tests/httpx2_alias/sitecustomize.py` and the `PYTHONPATH` in the
`tests` target of `python/Makefile`. 

### 2. `http_client=` rejects an old-httpx client

Passing an `httpx.Client` to the generated client is rejected. 

This is only reachable by constructing the generated client directly. `langsmith.Client` takes
`session: Optional[requests.Session]`, and `langsmith.AsyncClient` takes no client argument at all.
The message comes from generated code and cannot be changed in this repo.

**Reproduce.**

```bash
.venv/bin/python -c "
import httpx
from langsmith._openapi_client import Langsmith
try:
    Langsmith(api_key='k', base_url='https://example.com', http_client=httpx.Client())
except TypeError as e:
    print(e)"
```

```
Invalid `http_client` argument; Expected an instance of `httpx.Client` but got <class 'httpx.Client'>
```

*Workaround* Import httpx2 directly or from the Langsmith shim (`from langsmith._openapi_client._httpx import httpx`)

### 3. Sandbox request methods now return an `httpx2.Response`

`langsmith/sandbox/README.md` documents the sandbox service `request()` methods as returning
`httpx.Response`. They now return `httpx2.Response`.

This affects user code that:

- checks the result with `isinstance(resp, httpx.Response)`, which is now `False`,
- passes the result to a helper from old httpx,
- calls `.raise_for_status()` on the result. That raises `httpx2.HTTPStatusError`, which
  `except httpx.HTTPStatusError` does not catch.

**Reproduce.**

```bash
.venv/bin/python -c "
import httpx
from langsmith._openapi_client._httpx import httpx as ls_httpx
r = ls_httpx.Response(200, request=ls_httpx.Request('GET', 'https://example.com'))
print('backend:', ls_httpx.__name__)
print('isinstance(response, httpx.Response):', isinstance(r, httpx.Response))"
```

```
backend: httpx2
isinstance(response, httpx.Response): False
```

This PR updates `langsmith/sandbox/README.md` to state the real return type and note the consequences. The type annotations in `langsmith/sandbox/_models.py` needed no change — they annotate `httpx.Response` using the shim's `httpx` name, which already resolves to whichever backend is active.

### 4. TLS trust now comes from the OS, not certifi

`truststore` is not just an extra package. It replaces certifi's role. `httpx2` builds its default
SSL context as `truststore.SSLContext(ssl.PROTOCOL_TLS_CLIENT)`, which reads the operating system
trust store. Old httpx used certifi's bundled CA list.

`verify` still defaults to `True`, so certificate verification is not weakened. The source of trust
changes, and that cuts both ways:

- Users behind a corporate TLS-inspecting proxy should start working without an `SSL_CERT_FILE`
  workaround.
- Users on minimal containers with no system CA store, or who depended on certifi's bundle, may see
  new verification failures.

---

## Lower-impact notes

- Mixing a transport from one package with a client from the other is accepted at construction and
  fails later at request time, rather than being rejected up front.
- Importing `langsmith.utils` now loads the generated client eagerly, so first-import time goes from
  roughly 180-200 ms to roughly 300-350 ms. Timings vary by 15-20% between runs.
- `python/langsmith/wrappers/_anthropic.py` keeps its plain `import httpx`. That refers to the
  Anthropic SDK's own response types, not this SDK's backend, so it must not be repointed.

## Testing

- `make tests` — full unit suite with the alias active.
- `make test-wheel-imports` — builds the wheel, installs it into a clean venv with only the declared
  runtime dependencies, and asserts the backend resolves to `httpx2`. The assertion matters because
  the shim falls back silently: without it, the job would pass even if `httpx2` were missing from the
  declared deps and `httpx` arrived transitively. 
- `python/tests/unit_tests/test_httpx_compat.py` — asserts the sync and async clients pick httpx2.

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -27,7 +27,7 @@ dependencies = [
     "pydantic>=2,<3",
     "requests>=2.0.0",
     "orjson>=3.9.14; platform_python_implementation != 'PyPy'",
-    "httpx>=0.23.0,<1",
+    "httpx2>=2,<3",
     "requests-toolbelt>=1.0.0",
     "zstandard>=0.23.0",
     "packaging>=23.2",
--- a/python/Makefile
+++ b/python/Makefile
@@ -44,6 +44,7 @@ tests:
 	 -u LANGSMITH_TRACING \
 	 PYTHONDEVMODE=1 \
 	 PYTHONASYNCIODEBUG=1 \
+	 PYTHONPATH=tests/httpx2_alias \
 	 uv run python -m pytest --disable-socket --allow-unix-socket -n auto --durations=10 ${TEST}
 
 tests_watch:
@@ -90,8 +91,21 @@ test-wheel-imports:
 	uv pip install -r /tmp/langsmith-declared-deps.txt --python /tmp/langsmith-strict-test
 	/tmp/langsmith-strict-test/bin/python -c "\
 from langsmith._openapi_client import Langsmith; \
-from langsmith.client import Client"
-	rm -rf /tmp/langsmith-strict-test /tmp/langsmith-declared-deps.txt
+from langsmith.client import Client; \
+assert __import__('langsmith._openapi_client._httpx', fromlist=['httpx']).httpx.__name__ == 'httpx2'; \
+Langsmith(base_url='https://example.com').close(); \
+Client(api_url='https://example.com', api_key='test', auto_batch_tracing=False, info={}).close()"
+	uv run python3 -c "p='/tmp/langsmith-declared-deps.txt'; lines=open(p).read().splitlines(); open('/tmp/langsmith-httpx-deps.txt','w').write('\n'.join(x for x in lines if not x.startswith('httpx2')) + '\nhttpx>=0.23.0,<1\n')"
+	uv venv /tmp/langsmith-httpx-test
+	uv pip install dist/langsmith-*.whl --no-deps --python /tmp/langsmith-httpx-test
+	uv pip install -r /tmp/langsmith-httpx-deps.txt --python /tmp/langsmith-httpx-test
+	/tmp/langsmith-httpx-test/bin/python -c "\
+from langsmith._openapi_client import Langsmith; \
+from langsmith.client import Client; \
+assert __import__('langsmith._openapi_client._httpx', fromlist=['httpx']).httpx.__name__ == 'httpx'; \
+Langsmith(base_url='https://example.com').close(); \
+Client(api_url='https://example.com', api_key='test', auto_batch_tracing=False, info={}).close()"
+	rm -rf /tmp/langsmith-strict-test /tmp/langsmith-httpx-test /tmp/langsmith-declared-deps.txt /tmp/langsmith-httpx-deps.txt
 
 api_docs_build:
 	uv run python docs/create_api_rst.py
--- a/python/Makefile
+++ b/python/Makefile
@@ -95,17 +95,7 @@ from langsmith.client import Client; \
 assert __import__('langsmith._openapi_client._httpx', fromlist=['httpx']).httpx.__name__ == 'httpx2'; \
 Langsmith(base_url='https://example.com').close(); \
 Client(api_url='https://example.com', api_key='test', auto_batch_tracing=False, info={}).close()"
-	uv run python3 -c "p='/tmp/langsmith-declared-deps.txt'; lines=open(p).read().splitlines(); open('/tmp/langsmith-httpx-deps.txt','w').write('\n'.join(x for x in lines if not x.startswith('httpx2')) + '\nhttpx>=0.23.0,<1\n')"
-	uv venv /tmp/langsmith-httpx-test
-	uv pip install dist/langsmith-*.whl --no-deps --python /tmp/langsmith-httpx-test
-	uv pip install -r /tmp/langsmith-httpx-deps.txt --python /tmp/langsmith-httpx-test
-	/tmp/langsmith-httpx-test/bin/python -c "\
-from langsmith._openapi_client import Langsmith; \
-from langsmith.client import Client; \
-assert __import__('langsmith._openapi_client._httpx', fromlist=['httpx']).httpx.__name__ == 'httpx'; \
-Langsmith(base_url='https://example.com').close(); \
-Client(api_url='https://example.com', api_key='test', auto_batch_tracing=False, info={}).close()"
-	rm -rf /tmp/langsmith-strict-test /tmp/langsmith-httpx-test /tmp/langsmith-declared-deps.txt /tmp/langsmith-httpx-deps.txt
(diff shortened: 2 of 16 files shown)


python/pyproject.toml
```diff
@@ -27,7 +27,7 @@ dependencies = [
     "pydantic>=2,<3",
     "requests>=2.0.0",
     "orjson>=3.9.14; platform_python_implementation != 'PyPy'",
-    "httpx>=0.23.0,<1",
+    "httpx2>=2,<3",
```
KiewanVillatel, line 30: Just double checking, are we sure we should define httpx2 as a dependency here?

As I understood it, the goal of the shim in stlc was to use httpx2 if present in the dependencies, otherwise httpx.

By adding httpx2 in the dependencies here, we are forcing it and the shim becomes useless (httpx2 will always be used).

I have a preference for enforcing httpx2 (what this PR is doing), it's a bigger breaking change but makes things simpler (with the shim, if any dependency starts pulling httpx2, it would "silently" start using httpx2 instead of httpx in our SDK and might break things on the client side)

If we go this way, I think we should:
- import httpx2 directly in the different .py files instead of _httpx
- open a follow-up PR on stlc to remove the shim and use directly httpx2 

What do you think?

KiewanVillatel:
nit: if considered a breaking change, should we make it explicit using conventional commit by adding a `!`? chore(py)!

python/pyproject.toml
emil-lc: The goal is to switch to `httpx2` without breaking environments that only have the old `httpx` and aren't necessarily updating all dependencies - as the prototype #3341 calls them "dependency-suppressed environments".

The question that I do not know the answer to is: Are "dependency-suppressed envs" a real population, or just an overly cautious idea? It would require some surgery to get to such an environment. We don't really have data on them atm, though.

What we can do is add a marker to the User Agent on whether the client is using httpx vs httpx2. This can serve as a signal for whether/when we can get rid of the `httpx` fallback and use `httpx2` directly in both autogenerated and custom code.

KiewanVillatel: Ok I see!

No strong opinion between the options, I'd just not spend too much time trying to not break clients messing up with deps. So I'd be ok with:

1. keep shim with couple of weeks / months as a transition and optionally measure if cheap to do
2. hard switch already to httpx2 without shim

For option 1 maybe we could log a warning in case we can't import httpx2 saying that support of httpx will be dropped soon?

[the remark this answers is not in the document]
emil-lc: Documented in the PR description. Merging this would definitely be a Minor release, with documented changes. That being said, there are very few instances where an httpx class/object crosses between "userland" and LangSmith code.

emil-lc: Yep, going with 1. Let's see if the new UA addition points to any significant pools of httpx1 users. If there is a warning, it might be needed — otherwise, we can simply drop the httpx1 support altogether.

emil-lc:
> nit: if considered a breaking change, should we make it explicit using conventional commit by adding a `!`? chore(py)!

I was using a label for this, but perhaps a bang is a better/more standard approach. Updated title.

QuentinBrosse (approved):

PR #3419, opened 2026-08-19
feat(sandbox): add generate_download_url to the Python and JS clients

ramon-langchain:
## Overview

Exposes sandbox download links in the hand-written Python and JS sandbox clients. A download link serves one sandbox file with no LangSmith credential attached, for a browser tab, an `<a href>`, or a webhook consumer — anything that cannot send an API key the way `read()` requires.

Wraps `POST /api/v2/sandboxes/boxes/{name}/download-url`.

**Python** — `generate_download_url()` on `SandboxClient`, `AsyncSandboxClient`, `Sandbox`, and `AsyncSandbox`, returning a new `DownloadURL` model:

```python
link = sb.generate_download_url("/app/report.csv")
print(link.download_url)

link = sb.generate_download_url(
    "/app/page.html",
    expires_in_seconds=3600,
    content_type="text/html",
    content_disposition="inline",
)
print(link.expires_at)
```

**TypeScript** — `generateDownloadURL()` on `SandboxClient` and `Sandbox`, with `DownloadURL` and `GenerateDownloadURLOptions` types:

```typescript
const link = await sandbox.generateDownloadURL("/app/report.csv", {
  expiresInSeconds: 3600,
});
console.log(link.download_url);
```

Omitting the expiry mints a link that never expires, so `expires_at` is modeled as optional-and-omitted (`None` / `null`) rather than defaulting to a time — defaulting it would silently take the never-expires behavior away from callers. A non-positive expiry is rejected client-side.

Method naming matches `GenerateDownloadURL` / `generateDownloadUrl` in the generated Go and Java SDKs.

## Test plan

- [x] `uv run pytest tests/unit_tests/sandbox` (548 passed), `ruff check`, `ruff format --check`, `mypy langsmith/sandbox`
- [x] `tsc --noEmit`, `jest src/tests/sandbox.test.ts` (145 passed)
- [x] Unit (both languages): request body for the full option set, omission of unset options, `expires_at: null` preserved, 404 mapped to the not-found error, non-positive expiry rejected before any request

### DIFF
--- a/js/src/sandbox/client.ts
+++ b/js/src/sandbox/client.ts
@@ -481,6 +481,11 @@ export class SandboxClient {
    * Callers can add specific status checks (e.g. 404) before calling this.
    * @internal
    */
+  private _boxUrl(name: string, ...segments: string[]): string {
+    const suffix = segments.length ? `/${segments.join("/")}` : "";
+    return `${this._baseUrl}/boxes/${encodeURIComponent(name)}${suffix}`;
+  }
+
   private async _postJson(
     url: string,
     body: Record<string, unknown>,
@@ -639,7 +644,7 @@ export class SandboxClient {
     name: string,
     options?: { signal?: AbortSignal },
   ): Promise<Sandbox> {
-    const url = `${this._baseUrl}/boxes/${encodeURIComponent(name)}`;
+    const url = this._boxUrl(name);
 
     const response = await this._fetch(url, { signal: options?.signal });
 
@@ -725,7 +730,7 @@ export class SandboxClient {
       return this.getSandbox(name);
     }
 
-    const url = `${this._baseUrl}/boxes/${encodeURIComponent(name)}`;
+    const url = this._boxUrl(name);
     const payload: Record<string, unknown> = {};
     if (newName !== undefined) {
       payload.name = newName;
@@ -772,7 +777,7 @@ export class SandboxClient {
    * @throws LangSmithResourceNotFoundError if sandbox not found.
    */
   async deleteSandbox(name: string): Promise<void> {
-    const url = `${this._baseUrl}/boxes/${encodeURIComponent(name)}`;
+    const url = this._boxUrl(name);
 
     const response = await this._fetch(url, { method: "DELETE" });
 
@@ -801,7 +806,7 @@ export class SandboxClient {
     name: string,
     options?: { signal?: AbortSignal },
   ): Promise<ResourceStatus> {
-    const url = `${this._baseUrl}/boxes/${encodeURIComponent(name)}/status`;
+    const url = this._boxUrl(name, "status");
 
     const response = await this._fetch(url, { signal: options?.signal });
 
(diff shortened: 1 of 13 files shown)


js/src/sandbox/client.ts
```diff
@@ -895,6 +897,57 @@ export class SandboxClient {
...
+    const { expiresInSeconds, contentType, contentDisposition, signal } =
+      options;
+    if (expiresInSeconds !== undefined && expiresInSeconds < 1) {
+      throw new LangSmithValidationError(
+        `expiresInSeconds must be greater than 0 (got ${expiresInSeconds}); ` +
+          `omit it for a link that never expires`,
+      );
+    }
+
+    const url = `${this._baseUrl}/boxes/${encodeURIComponent(
```
ramon-langchain: surely this type of URL constructions should not be repeated in every method?

js/src/sandbox/types.ts
```diff
@@ -608,6 +608,37 @@ export interface CaptureSnapshotOptions {
...
+ */
+export interface GenerateDownloadURLOptions {
+  /**
+   * Link TTL in seconds. Omit for a link that never expires.
+   */
+  expiresInSeconds?: number;
+  /** Content-Type to serve the file as. */
+  contentType?: string;
+  /** Content-Disposition to serve the file with: "attachment" or "inline". */
+  contentDisposition?: string;
```
ramon-langchain: should we type this more strictly?

langchain-infra (approved):

js/src/sandbox/client.ts
ramon-langchain: Fixed in 75c152a2.

TypeScript now has one private helper on `SandboxClient`:

```ts
private _boxUrl(name: string, ...segments: string[]): string {
  const suffix = segments.length ? `/${segments.join("/")}` : "";
  return `${this._baseUrl}/boxes/${encodeURIComponent(name)}${suffix}`;
}
```

All 8 inline constructions in `client.ts` go through it (`get`/`update`/`delete`/`status`/`start`/`stop`/`snapshot`/`download-url`), so `encodeURIComponent` is applied in exactly one place.

Python had the same repetition, so it got the same treatment: a module-level `_box_url(base_url, name, *segments)` in `_client.py`, imported by `_async_client.py` the way `_quote_path_segment` already was, replacing 9 f-strings in each client.

Left the two collection URLs (`/boxes`, `/snapshots`) alone since they take no name and nothing to encode.

— Claude

js/src/sandbox/types.ts
ramon-langchain: Yes — narrowed in 75c152a2.

`contentDisposition` is no longer a free string. TypeScript:

```ts
export type DownloadContentDisposition = "attachment" | "inline";
```

and `GenerateDownloadURLOptions.contentDisposition?: DownloadContentDisposition`, exported from `langsmith/sandbox`. Python mirrors it with `DownloadContentDisposition = Literal["attachment", "inline"]`, used in all four `generate_download_url` signatures. The CLI now rejects any other value locally too.

That matches the server, which narrows the header to exactly those two and 400s on anything else, so a typo now fails at the type level instead of as a round trip.

Left `contentType` as a plain string on purpose: the server accepts any parseable media type, so an enum there would be wrong. `expiresInSeconds` stays `number` with a runtime check for `>= 1`, since the meaningful constraint is a range, not a set.

— Claude

PR #3430, opened 2026-08-20
feat(sandbox): support proxy_config in update_sandbox (Python + JS)

ramon-langchain:
`proxy_config` is settable only at create time. But proxy-injected credentials are short-lived — a GitHub App installation token lasts an hour — while sandboxes are not: idle TTL is measured in hours and stopped retention in days. So any sandbox reused past its first hour needs its proxy config rewritten, and there is no supported client path to do it. Create-time-only configuration makes the feature usable once per sandbox. Same for `aws_auth` and `gcp_auth` rules.

The server has always accepted it — `UpdateSandboxPayload.ProxyConfig` on `PATCH /boxes/{name}` — so callers work around the gap by hand-building a raw PATCH, which gives up the client's URL construction, auth headers, retry transport (429/502/503/504), and all of the typed-error mapping.

Both clients now take it, mirroring the create parameter and forwarding to the server as-is:

- **Python** — `proxy_config` on `SandboxClient.update_sandbox` / `AsyncSandboxClient.update_sandbox`.
- **JS** — `proxyConfig` on `UpdateSandboxOptions`, serialized as `proxy_config`, and added to the guard that short-circuits an all-empty options object to `getSandbox`.

`SandboxResponse` already returns `proxy_config`, so the response models round-trip it with no change.

### Typed not-ready error

The API rejects a proxy-config update on any sandbox that is not `ready` — which is the *common* case for a reused sandbox, because idle boxes are stopped, not deleted. That rejection is a 400 that `handle_client_http_error` / `handleClientHttpError` funnelled into a bare `SandboxClientError` / `LangSmithSandboxError`, leaving callers to match on status codes.

`raise_if_not_ready` (Python) and `throwIfNotReady` (JS) now raise `SandboxNotReadyError` / `LangSmithSandboxNotReadyError` for it. Both exceptions already exist and are exported, and the sandbox-operation handlers already apply the same 400-means-not-ready mapping; this brings the client path in line. Recovery becomes:

```python
try:
    await client.update_sandbox(name, proxy_config=cfg)
except SandboxNotReadyError:
    await client.start_sandbox(name)
    await client.update_sandbox(name, proxy_config=cfg)
```

Gated on the proxy config being present and on `error_type == "InvalidRequest"`, so it can't swallow a future 400 from another field. The JS version reads a clone of the response, so `handleClientHttpError` can still consume the body when the rejection turns out to be something else.

### Deliberately not included

An auto-start flag. Starting a sandbox consumes quota and takes time, and `POST /start` is not idempotent — it 400s on anything not `stopped`/`failed`. Hiding that inside an update call makes the failure modes worse; the typed error gives callers what they need to decide.

`vcpus`, `cpu_millicores`, `mem_bytes`, `fs_capacity_bytes` and `tag_value_ids` are also server-supported and client-absent in both SDKs. Out of scope here.

### Docs

All three docstrings state two behaviors that are not discoverable from the signature and that callers get wrong: rules **replace** rather than merge, and opaque header values carry over from the existing config, so rotating one credential does not require re-supplying secrets you can no longer read.

## Test Plan

- [x] Python: `uv run pytest tests/unit_tests/sandbox/` — 552 pass
- [x] JS: `pnpm test src/tests/sandbox.test.ts` — 147 pass
- [x] New tests on all three clients: payload carries the proxy config; a 400 `InvalidRequest` raises the not-ready error
- [x] Python `make lint` clean (ruff + mypy, 118 files); JS `pnpm lint` clean and `tsc --noEmit` reports 0 errors

### DIFF
--- a/js/src/sandbox/client.ts
+++ b/js/src/sandbox/client.ts
@@ -36,6 +36,7 @@ import {
 import {
   handleClientHttpError,
   handleSandboxCreationError,
+  throwIfNotReady,
   validateTtl,
 } from "./helpers.js";
 import { validateMountConfigProxyConfig } from "./mounts.js";
@@ -695,8 +696,8 @@ export class SandboxClient {
    */
   async updateSandbox(name: string, newName: string): Promise<Sandbox>;
   /**
-   * Update a sandbox's name and/or retention settings (idle stop and
-   * delete-after-stop).
+   * Update a sandbox's name, retention settings (idle stop and
+   * delete-after-stop), and/or proxy config.
    *
    * @param name - Current sandbox name.
    * @param options - Fields to update. Omit a field to leave it unchanged.
@@ -704,6 +705,8 @@ export class SandboxClient {
    * @throws LangSmithResourceNotFoundError if sandbox not found.
    * @throws LangSmithResourceNameConflictError if newName is already in use.
    * @throws LangSmithValidationError if retention values are invalid.
+   * @throws LangSmithSandboxNotReadyError if proxyConfig was given and the
+   * sandbox is not `ready`.
    */
   async updateSandbox(
     name: string,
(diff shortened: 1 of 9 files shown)


langchain-infra (approved):

socksninja:
Using the newly merged sandbox API as the basis for a one-run interoperability test: would `ramon-langchain` or the LangSmith sandbox team be the best contact for a disposable evaluation workspace/API key? We only need one harmless LangGraph execution and an independently readable trace; no customer data or persistent environment required. SABLE has a fail-closed workflow ready to run.

socksninja:
@ramon-langchain — very small interoperability ask: would the LangSmith sandbox team be willing to provide a disposable workspace/API key for exactly one non-production SABLE run? We already have the fail-closed workflow; the run creates one harmless GitHub test comment, verifies it independently, and emits a compact receipt. No customer data, production credentials, or private traces are needed. A temporary key/workspace that can be revoked afterward is sufficient.

PR #3439, opened 2026-08-23
test(evaluation): add regression test for feedback_config dict handling

nanjeshramesh:
This closes the loop on langchain-ai/langchain#31802, which reported that passing
a plain dict to EvaluationResult.feedback_config silently drops keys it doesn't
recognize.

feedback_config is typed as Optional[Union[FeedbackConfig, dict]], and
FeedbackConfig is a permissive TypedDict where every field is optional. That means
a dict with keys that don't match FeedbackConfig at all can still validate against
it (nothing is required), so there was a real risk of the union match silently
coercing into FeedbackConfig and dropping unrecognized keys instead of falling
through to the plain dict.

I could not reproduce the reported behavior against current dependencies. I even
pinned pydantic to the exact version from the report (2.11.1), and it still
worked correctly, because pydantic-core resolves independently and had already
moved past whatever version had this issue. So there's nothing to fix in this
repo's code today.

There was no test covering this case though, so a future pydantic-core regression
(or someone pinning an older version) could bring the bug back with nothing to
catch it. This adds a test that locks in the correct behavior for a fully unknown
dict and for a dict that mixes a real FeedbackConfig key with an unrelated one.

Ran the full evaluation test suite and ruff locally, both clean.

### DIFF
--- a/python/tests/unit_tests/evaluation/test_evaluator.py
+++ b/python/tests/unit_tests/evaluation/test_evaluator.py
@@ -431,3 +431,29 @@ def test_check_value_non_numeric(caplog):
         "Numeric values should be provided in the 'score' field, not 'value'."
         not in caplog.text
     )
+
+
+def test_feedback_config_arbitrary_dict_is_preserved():
+    """A dict passed as `feedback_config` must keep keys unknown to `FeedbackConfig`.
+
+    Regression test for https://github.com/langchain-ai/langchain/issues/31802:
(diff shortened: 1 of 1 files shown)


nanjeshramesh:
Gentle nudge on this regression test for the `feedback_config` dict handling from langchain-ai/langchain#31802. It is test-only with no behavior change, so it should be a quick read. Happy to adjust if needed.

PR #3444, opened 2026-08-25
fix(py): suppress import-untyped on the optional langsmith_pyo3 import

### LINKED ISSUE #3443: make lint fails on main when the langsmith_pyo3 extra is installed
### Summary

`make lint` fails on a clean checkout of `main` for any contributor whose virtualenv has the optional `langsmith_pyo3` extra installed. mypy reports an unsuppressed `import-untyped` error on the guarded `langsmith_pyo3` import in `client.py`, so the lint target exits non-zero before a contributor has changed anything. CI does not catch it because the lint job never installs that extra, so the error code mypy emits there is a different one that the existing `type: ignore` already covers.

### Repro (versions)

From a checkout of `main` at `2fd60641`, in `python/`:

```bash
uv sync --extra langsmith_pyo3
make lint
```

```
uv run mypy langsmith
langsmith/client.py:1491: error: Skipping analyzing "langsmith_pyo3": module is installed, but missing library stubs or py.typed marker  [import-untyped]
langsmith/client.py:1491: note: Error code "import-untyped" not covered by "type: ignore[import-not-found, no-redef]" comment
Found 1 error in 1 file (checked 118 source files)
make: *** [lint] Error 1
```

| `langsmith_pyo3` in the venv | mypy error code | Covered by the existing ignore | `make lint` |
|---|---|---|---|
| not installed (what CI does) | `import-not-found` | yes | passes |
| installed | `import-untyped` | no | fails |

### Root cause (`langsmith/client.py`::`Client.__init__`)

1. `client.py:1491` guards the optional import with `import langsmith_pyo3  # type: ignore[import-not-found, no-redef]`.
2. `langsmith-pyo3` (0.1.0rc5) ships no `py.typed` marker. When the package *is* installed, mypy resolves it but has no types for it, so it emits `import-untyped` — not `import-not-found`.
3. `import-untyped` is absent from the ignore list, so the error is unsuppressed and `make lint` fails.
4. `.github/workflows/ci.yml` installs the lint environment with `uv sync --group dev --group lint`. `langsmith_pyo3` is a `[project.optional-dependencies]` extra, not a dependency group, so it is never installed there and mypy takes the `import-not-found` path. The Python Lint job stays green while local `make lint` fails.

### Expected behavior

`make lint` passes on an unmodified `main` regardless of whether the optional `langsmith_pyo3` extra is present in the environment.

### Suggested fix

Add `import-untyped` to the existing ignore list on that import. It suppresses the error when the package is installed, and it is inert when it is not: `warn_unused_ignores` is not enabled in `[tool.mypy]`, so an error code in the list that does not fire is not itself an error. I confirmed both directions — mypy clean with the extra installed, and clean with the module absent. Will open a PR.

### System info

- langsmith 0.11.1, repo at `2fd60641`
- langsmith-pyo3 0.1.0rc5
- mypy 2.3.0
- Python 3.13.13
- macOS 26.4.1 (arm64)

Vissirexa:
Fixes #3443.

`make lint` fails on an unmodified `main` for anyone whose venv has the optional `langsmith_pyo3` extra installed:

```
langsmith/client.py:1491: error: Skipping analyzing "langsmith_pyo3": module is installed, but missing library stubs or py.typed marker  [import-untyped]
langsmith/client.py:1491: note: Error code "import-untyped" not covered by "type: ignore[import-not-found, no-redef]" comment
```

The ignore on that guarded import lists `import-not-found`, which is what mypy emits when the module cannot be resolved at all. `langsmith-pyo3` ships no `py.typed`, so when the extra is actually installed mypy resolves the module, finds no types for it, and emits `import-untyped` instead — a code the list does not cover.

CI does not see this. The lint job installs with `uv sync --group dev --group lint`, and `langsmith_pyo3` is a `[project.optional-dependencies]` extra rather than a dependency group, so it is never present there and mypy takes the `import-not-found` path.

```diff
-import langsmith_pyo3  # type: ignore[import-not-found, no-redef]
+import langsmith_pyo3  # type: ignore[import-not-found, import-untyped, no-redef]
```

Adding the code covers the installed case and is inert in the other — `warn_unused_ignores` is not enabled in `[tool.mypy]`, so an error code in the list that does not fire is not itself an error. I checked both directions: mypy is clean with the extra installed, and clean against a module that does not exist.

No test accompanies this since it is a type-checker suppression with no runtime behavior; the reproduction is `uv sync --extra langsmith_pyo3 && make lint`, which fails before this change and passes after.

Scope kept to the one import. `make format`, `make lint` and `make tests` (2019 passed, 6 skipped, 1 xfailed) are green locally on Python 3.13 with the extra installed.

### DIFF
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -1488,7 +1488,7 @@ class Client:
         if ls_utils.get_env_var("USE_PYO3_CLIENT") is not None:
             langsmith_pyo3 = None
             try:
-                import langsmith_pyo3  # type: ignore[import-not-found, no-redef]
+                import langsmith_pyo3  # type: ignore[import-not-found, import-untyped, no-redef]
             except ImportError as e:
                 logger.warning(
                     "Failed to import `langsmith_pyo3` when PyO3 client was requested, "


QuentinBrosse:
Thanks for the PR! I cherry-picked your commit into https://github.com/langchain-ai/langsmith-sdk/pull/3462 so the CI can run.

PR #3445, opened 2026-08-25
fix(py,js): consistent (non-v7) UUID rewriting for replicas

emil-lc:
## Description

`uuid7_deterministic` (py) / `nonCryptographicUuid7Deterministic` (js) builds a UUIDv7 as 6 timestamp bytes + 10 hash-derived bytes. The timestamp is copied from the original ID only when that ID is itself a UUIDv7 - otherwise, the wall clock is substituted, making the function impure.

It is called repeatedly for the same ID: once per `post`, again per `patch`, and again from every descendant remapping its `trace_id`, `parent_run_id` and `dotted_order` segments (a two-run tree derives the root's ID five times). Each call read a different clock, so a replica received a run that never completes plus a completion for a run that does not exist, and descendants pointed at ancestors that were never created.

The fix fills the timestamp field from the six hash bytes the derivation already leaves unused, making it a pure function of `(originalId, key)`. The UUIDv7 branch is untouched, so the default path - every langchain-core run and every `RunTree()` without an explicit ID - is byte-identical. No call sites change.

Both SDKs carried the defect identically and are fixed identically; they already agree bit-for-bit on the hash portion, so landing only one would leave a Python↔JS distributed trace stably divergent rather than merely unstable.

Trade-off: for non-UUIDv7 originals the derived ID is no longer time-ordered. There is nothing to order it by - the previous value recorded when the SDK happened to remap, not when the run started - and `start_time` and `dotted_order` still carry real timestamps.

## Release Note

Replica run IDs derived from non-UUIDv7 run IDs are now stable, so replica runs complete correctly and their traces stay intact.

## Test Plan

- [x] New: post/patch pairing, and an ancestor's ID deriving identically from parent and child - both red without the fix, in both SDKs
- [x] Non-UUIDv7 case added to the existing cross-SDK vector tests, pinning Python/JS agreement
- [x] Existing `timestamp handling` tests updated in both SDKs - they asserted the clock behaviour
- [x] Full suites: Python 2037 passed; JS 879 passed (pre-existing failures unchanged, bisected)
- [x] Live dev deployment: a `uuid4`-ID run's secondary replica completes with `status=success`


python/tests/unit_tests/test_replica_compression.py
KiewanVillatel, line 1: This file most likely belongs to another PR that got mixed with this one.

KiewanVillatel (approved):

python/tests/unit_tests/test_replica_compression.py
emil-lc: Indeed - removed now.

PR #3448, opened 2026-08-26
fix(py): compress replica writes that carry their own credentials

emil-lc:
## Description

`_create_run` and `_update_run` took the compressed (zstd) path only when **all six**
per-call auth arguments were `None`. Any replica carrying e.g. `api_url`/`auth` fell through to
the uncompressed `tracing_queue`, silently. Using `LANGSMITH_RUNS_ENDPOINTS`
always produces that shape, so the whole documented multi-endpoint configuration paid the
uncompressed cost.

This was no accident - a zstd frame had nowhere to record where it should be POSTed.

With this PR the frame now carries the missing field. `CompressedTraces.destinations` records where a
frame goes, `accepts()` admits only ops bound for the same place, and the sender routes by
it (`None` keeps the old `_write_api_urls` behaviour). The check, the commit and the write
happen in **one hold** of the existing lock.

There is _still_ one frame per client, so when a client has several destination sets only one is
compressed: the first to write owns it, and `reset()` deliberately does not clear
`destinations`, so ownership is decided once instead of re-raced after every flush. The
rest use the plaintext queue, exactly as today - nothing is worse than before. A one-time warning
replaces the silence.

`create_feedback` is the third writer into that frame and gets the same check.

## Release Note

Replicas hitting the **same destination** URL now use compressed trace ingestion instead of silently falling back to plaintext. Payloads sent to different projects are bundled in the same zstd Multipart frame, greatly reducing the memory footprint (e.g 8 replicas go from ~90MB -> ~2MB)

## Test Plan

- [x] `python/tests/unit_tests/test_replica_compression.py` (new, 25 tests): no cross-destination bleed with decompressed frame bodies, stable ownership across flushes, feedback kept out of a replica-owned frame, concurrent admission vs. drain
- [x] Credential resolution produces byte-identical headers to the previous paths on all 7 auth shapes, including the `None`-key and `service_key`-only cases
- [x] Non-regression: `api_urls={A,B}` still replays one frame to both; project-only replicas unchanged; existing `test_multiple_endpoints.py` / `test_replica_endpoints.py` untouched and passing
- [x] Full unit suite: 2047 passed (one pre-existing failure from a locally-set `LANGSMITH_ENDPOINT`; passes when cleared)

### DIFF
--- a/python/langsmith/_internal/_compressed_traces.py
+++ b/python/langsmith/_internal/_compressed_traces.py
@@ -49,12 +49,23 @@ class CompressedTraces:
         self.lock = threading.Lock()
         self.uncompressed_size: int = 0
         self._context: list[str] = []
+        # Frame's destinations. Sent verbatim to each, so all ops in one frame
+        # must match. Set once, kept across resets.
+        self.destinations: Optional[frozenset] = None
 
         self.compressor_writer = ZstdCompressor(
             level=compression_level, threads=compression_threads
         ).stream_writer(self.buffer, closefd=False)
 
+    def accepts(self, destinations: frozenset) -> bool:
+        """Report whether an op for `destinations` may join this frame.
+
+        Hold self.lock; commit and write in the same hold.
+        """
+        return self.destinations in (None, destinations)
+
     def reset(self) -> None:
+        # destinations survives: a new frame inherits it, so ownership is decided once.
         self.buffer = io.BytesIO()
         self.trace_count = 0
         self.uncompressed_size = 0
--- a/python/langsmith/_internal/_compressed_traces.py
+++ b/python/langsmith/_internal/_compressed_traces.py
@@ -49,8 +49,9 @@ class CompressedTraces:
         self.lock = threading.Lock()
         self.uncompressed_size: int = 0
         self._context: list[str] = []
-        # Frame's destinations. Sent verbatim to each, so all ops in one frame
-        # must match. Set once, kept across resets.
+        # Where this frame goes. Sent verbatim to each, so all ops in one frame
+        # must share it. Claimed by the first op written, released when the frame
+        # is sent, so a client whose replicas change is not stuck on the first set.
         self.destinations: Optional[frozenset] = None
 
         self.compressor_writer = ZstdCompressor(
@@ -65,7 +66,8 @@ class CompressedTraces:
         return self.destinations is None or self.destinations == destinations
 
     def reset(self) -> None:
-        # destinations survives: a new frame inherits it, so ownership is decided once.
+        # The next op claims the new frame, whatever its destinations.
+        self.destinations = None
         self.buffer = io.BytesIO()
         self.trace_count = 0
         self.uncompressed_size = 0
--- a/python/langsmith/_internal/_compressed_traces.py
+++ b/python/langsmith/_internal/_compressed_traces.py
@@ -62,7 +62,7 @@ class CompressedTraces:
 
         Hold self.lock; commit and write in the same hold.
         """
-        return self.destinations in (None, destinations)
+        return self.destinations is None or self.destinations == destinations
 
     def reset(self) -> None:
         # destinations survives: a new frame inherits it, so ownership is decided once.
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -46,6 +46,7 @@ from typing import (
     Any,
     Callable,
     Literal,
+    NamedTuple,
     Optional,
     TypedDict,
     Union,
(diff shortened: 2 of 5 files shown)


python/langsmith/_internal/_compressed_traces.py
```diff
@@ -49,12 +49,23 @@ def __init__(self, max_uncompressed_size_bytes: Optional[int] = None) -> None:
...
         self.compressor_writer = ZstdCompressor(
             level=compression_level, threads=compression_threads
         ).stream_writer(self.buffer, closefd=False)
 
+    def accepts(self, destinations: frozenset) -> bool:
+        """Report whether an op for `destinations` may join this frame.
+
+        Hold self.lock; commit and write in the same hold.
+        """
+        return self.destinations in (None, destinations)
```
jacoblee93: This is pretty strange looking to me, why not just do an equality check?

python/langsmith/client.py
```diff
@@ -2667,6 +2668,90 @@ def create_run(
...
+        as _apply_auth_overrides does: no fallback beside a service key.
+        """
+        if not any(auth):
+            return frozenset(
+                ReplicaAuth(url, key, None, None, None, None)
+                for url, key in self._write_api_urls.items()
+            )
+        api_key = auth.api_key
+        if api_key is None and not any(
+            [auth.service_key, auth.authorization, auth.cookie]
```
jacoblee93, line 2746: Why cookie?

python/langsmith/_internal/_compressed_traces.py
```diff
@@ -49,12 +49,23 @@ def __init__(self, max_uncompressed_size_bytes: Optional[int] = None) -> None:
         self.lock = threading.Lock()
         self.uncompressed_size: int = 0
         self._context: list[str] = []
+        # Frame's destinations. Sent verbatim to each, so all ops in one frame
+        # must match. Set once, kept across resets.
+        self.destinations: Optional[frozenset] = None
```
jacoblee93, line 55: Can't replicas for a given client change? Since replicas are set at the run tree level

I don't think this is a static list

python/langsmith/client.py
```diff
@@ -2667,6 +2668,90 @@ def create_run(
...
+    def _resolve_destinations(self, auth: ReplicaAuth) -> frozenset[ReplicaAuth]:
+        """Where a frame carrying this op must be POSTed.
+
+        No per-call auth: every write endpoint, verbatim -- a None key must stay
+        None, it strips X-API-KEY. Otherwise the named endpoint, api_key resolved
+        as _apply_auth_overrides does: no fallback beside a service key.
+        """
+        if not any(auth):
+            return frozenset(
+                ReplicaAuth(url, key, None, None, None, None)
```
jacoblee93: nit: use kwargs this is completely unreadable

python/langsmith/client.py
emil-lc: Cookie is one of the supported authorization components that needs to travel with the frame. This PR basically fixes the case where we would just give up compression if any authorization is specified, and makes it so that we still keep compression as long as all authorizations are equal (as in, replica targets are just different projects on the same instance).

Next PR (#3450) is the one that adds auth support per replica for the cases where frame contents don't change.

python/langsmith/_internal/_compressed_traces.py
emil-lc: Switched to the more readable `return self.destinations is None or self.destinations == destinations`

python/langsmith/client.py
emil-lc: Switched to a named version - it does indeed look better!

python/langsmith/_internal/_compressed_traces.py
emil-lc: That's a nice catch. I didn't think replicas would change at runtime. Updated code now to reset destinations on flush, to accommodate for that possibility.

jacoblee93: They don't usually but they technically can yeah

python/langsmith/client.py
```diff
@@ -2710,54 +2799,20 @@ def _create_run(
...
-                        _BOUNDARY,
-                    )
-                    if enqueued:
-                        self.compressed_traces.trace_count += 1
-                        self._data_available_event.set()
-
+                admitted = self._compress_admitted(multipart_form, destinations)
                 _close_files(list(opened_files.values()))
+                if not admitted and self.tracing_queue is not None:
+                    self._enqueue_op(run_create["dotted_order"], serialized_op, auth)
```
jacoblee93, line 2805: Any situations where this gets starved? Probably not right?

jacoblee93 (approved):

python/langsmith/client.py
emil-lc: No - as long as we have a plaintext  (non-compressed) ingestion queue running at all (`self.tracing_queue`), which should always be the case.

PR #3450, opened 2026-08-26
perf(py): serialize identical replicas once, into one frame

emil-lc:
## Description

Stacked on #3448. That change let a compressed frame record where it must be sent, but
nothing ever produced an operation bound for more than one place — so every replica with its
own credentials still got its own serialized copy, and only one could use the frame.

Replicas that would produce the same bytes are now serialized once, and the single operation
is marked as belonging to all of their destinations. One frame carries one copy and is sent
to each destination with that destination's own credentials. This is what
`LANGSMITH_RUNS_ENDPOINTS` with several endpoints needs.

Credentials do not affect the bytes, which is what lets a group span destinations. Project
does not affect the destination, so multi-project replicas that share credentials keep
sharing one frame.

Replicas whose projects **and** destinations both differ still need one Zstd frame each, so one
is compressed and the rest use the non-compressed queue. That would be the next optimization.

## Release Note

Replicas that write the same run to several endpoints, as `LANGSMITH_RUNS_ENDPOINTS` does,
are now serialized once and compressed into a single frame instead of once per endpoint.

## Test Plan

- [x] 9 new tests in `test_replica_compression.py`: one serialization for identical replicas, one frame reaching every destination with its own key, and one test per field of the grouping key asserting that replicas which differ are not grouped
- [x] Three existing fixtures updated — same-project credentialed replicas are now one group
- [x] 23 assertions retargeted in `test_replica_endpoints.py` / `test_multiple_endpoints.py`
- [x] Full unit suite: 2056 passed

### DIFF
--- a/python/langsmith/client.py
+++ b/python/langsmith/client.py
@@ -2781,21 +2781,21 @@ class Client:
         cookie: Optional[str] = None,
         _replica_auths: Optional[Sequence[ReplicaAuth]] = None,
     ) -> None:
+        auths = _replica_auths or [
+            ReplicaAuth(
+                api_url=api_url,
+                api_key=api_key,
+                service_key=service_key,
+                tenant_id=tenant_id,
+                authorization=authorization,
+                cookie=cookie,
+            )
+        ]
         if (
             # batch ingest requires trace_id and dotted_order to be set
             run_create.get("trace_id") is not None
             and run_create.get("dotted_order") is not None
         ):
-            auths = _replica_auths or [
-                ReplicaAuth(
-                    api_url=api_url,
-                    api_key=api_key,
-                    service_key=service_key,
-                    tenant_id=tenant_id,
-                    authorization=authorization,
-                    cookie=cookie,
-                )
-            ]
             destinations = (
                 self._resolve_destinations(*auths)
                 if self.compressed_traces is not None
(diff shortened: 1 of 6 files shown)


python/langsmith/client.py
```diff
@@ -2611,6 +2612,10 @@ def create_run(
             ```
         """
         service_key: str | None = kwargs.pop("service_key", None)
+        # Popped before run_create is built, or it would become a run field.
+        replica_auths: Optional[Sequence[ReplicaAuth]] = kwargs.pop(
```
jacoblee93, line 2616: Why not just make it a named optional kwarg?

python/langsmith/run_trees.py
```diff
@@ -806,32 +806,54 @@ def _remap_for_project(
...
+            key = (
+                replica.get("client") or self.client,
+                replica.get("project_name") or self.session_name,
+                replica.get("updates"),
+                replica.get("primary"),
+            )
+            # ponytail: O(n^2) over a handful of replicas, and `updates` compares by
+            # value without needing a hashable freeze. Use a key if that changes.
+            for group in groups:
+                if group[:4] == key:
```
jacoblee93: Can we just not use tuples this is so unreadable

jacoblee93 (approved):

python/langsmith/client.py
emil-lc: That's where I started :) It's internal, though, and I didn't like the idea of having it exposed as an official function interface. Since we already have blanket kwargs that are following this pattern, I switched to riding on it.

python/langsmith/run_trees.py
emil-lc: I agree this got quite dense too. Added a couple of NamedTuple abstractions and extra comments; I think it's more understandable now. Please take a look at the last commit.

PR #3483, opened 2026-09-02
fix(py): make the multipart body replayable across both retry layers

thibautlehmann:
Fixes [LSDK-501](https://linear.app/langchain/issue/LSDK-501/path-based-attachments-silently-dropped-when-a-multipart-batch-is-re).
Fixes [LSDK-460](https://linear.app/langchain/issue/LSDK-460/make-20mb-multipart-bodies-re-sendable-on-retry).

## Bug
`Path`-based attachments are opened once; the file handle lives in `parts`. A `MultipartEncoder` is single-use — no `tell`/`seek`. Two layers resend the *same* body object on retry (our own app loop, and urllib3 beneath it via `_default_retry_config`'s `status_forcelist`): after the first send the handle's at EOF, so every resend ships an attachment-less run under a stale `Content-Length` — HTTP 200, silent data loss.

Same missing `tell`/`seek` also caused LSDK-460: for bodies over the 20MB streaming threshold, a retryable failure resent the exhausted encoder under a stale `Content-Length` and the server just hung waiting for bytes that would never arrive.

## Fix
`RewindableMultipartBody` wraps the encoder with `tell`/`seek(0)`, used unconditionally for both the buffered (`<=20MB`) and streamed (`>20MB`) branches. `seek(0)` rewinds every file-backed part and rebuilds the encoder, so urllib3's rewind-before-resend actually works. `_MULTIPART_INLINE_MAX_BYTES` replaces the old `20_000_000` literal.

## Tests
`test_multipart_attachment_reencode.py`: real HTTP sinks assert the on-the-wire bytes are complete and intact across every resend (previously empty after attempt 1).

`test_multipart_retry_matrix.py`: ported the full `status_forcelist` matrix (429/500/502/503/504/408/413/425, both retry layers) onto the streamed branch — forced via `mock.patch("langsmith.client._MULTIPART_INLINE_MAX_BYTES", 0)` instead of a real oversized payload, so it stays fast. This is the LSDK-460 regression coverage.


python/langsmith/_internal/_multipart.py
```diff
@@ -29,3 +33,81 @@ def join_multipart_parts_and_context(
         acc_parts.extend(parts_and_context.parts)
         acc_context.append(parts_and_context.context)
     return MultipartPartsAndContext(acc_parts, "; ".join(acc_context))
+
+
+def _rewind_parts(parts: Iterable[MultipartPart]) -> None:
```
QuentinBrosse: nit: should we move it in the class?

python/langsmith/_internal/_multipart.py
```diff
@@ -29,3 +33,81 @@ def join_multipart_parts_and_context(
...
+class RewindableMultipartBody:
+    """A replayable request body for ``MultipartEncoder``-encoded parts.
+
+    ``MultipartEncoder`` is single-use: it reads every file-backed part to EOF
+    and exposes neither ``seek`` nor ``tell``, so no retry layer can replay it.
+    urllib3 retries the statuses in ``_default_retry_config``'s
+    ``status_forcelist`` beneath our own retry loop, and it resends the *same*
+    body object -- an exhausted encoder, i.e. an empty body under the
+    already-declared ``Content-Length``. The server accepts that as a complete
+    run, so the attachment is dropped with an HTTP 200.
```
QuentinBrosse: I think it's the entire payload, not just the attachment, right?

thibautlehmann:
@QuentinBrosse, the unit tests in this PR cover the logic, but I also wanted to see `RewindableMultipartBody` survive a real retry against real infra, so:

**Locally, with a fault-injecting proxy.** The proxy returns a 500 for the first two POSTs to `/runs/multipart`, then lets everything through. I sent one buffered attachment (≤20MB) and one streamed (>20MB), pulled each back down via its presigned URL, and checked SHA-256 against the original. Both matched, no truncation and no empty attachment on the resent attempt.

**Against prod, happy path.** Same two attachments, no way to inject faults there. Both came back byte-identical.

In both runs I confirmed `_send_multipart_req` was the code path that actually fired, not the compressed fallback, and not `_batch_ingest_run_ops` or `_create_run_non_batch`. So the passing checksums are evidence about this code, not just evidence that something worked.

QuentinBrosse (approved):

python/langsmith/_internal/_multipart.py
thibautlehmann: yes, good point, updated

python/langsmith/_internal/_multipart.py
thibautlehmann: good point, done

PR #3484, opened 2026-09-02
feat: LangSmithSecret class for redacting sensitive function parameters

emil-lc:
# Description

Introduces `LangSmithSecret` class: mark a value once, mask it everywhere in a trace

Example:
```
from langsmith import LangSmithSecret

API_KEY = LangSmithSecret(os.environ["VENDOR_API_KEY"])


@traceable
def call_vendor(api_key: str, prompt: str): ...


call_vendor(api_key=API_KEY, prompt="hi")
# traced inputs: {"api_key": "[LANGSMITH SECRET]", "prompt": "hi"}
```
A str subclass that LangSmith serializes as [LANGSMITH SECRET]. It's a real string everywhere else — your vendor client, your logs, and your own json.dumps all see the true value. Derived strings stay marked, so .strip() and slicing don't lose it, and __repr__ is masked so an object that interpolates a secret doesn't leak it.

# How it works

The `OPT_PASSTHROUGH_SUBCLASS` serialization option of `orjson` routes `str` subclasses to orjson's default hook. A side effect is that option also diverts every other builtin subclass, so `_serialize_json` coerces those back to their exact builtin, so output is byte-for-byte unchanged for payloads with no secrets.

The following serialization boundaries were updated to provide the same guarantee:

- stdlib-json fallback (PyPy, no orjson) - masks up front, and again on whatever default returns, since that output is re-encoded natively. Covers sets, frozensets, deques, dataclasses, and .dict()/model_dump() objects.
- OTel span attributes - masked before a value becomes an attribute.
- Dict _keys_ - orjson rejects a str-subclass key outright, so a secret key always lands in the recovery tier of `dumps_json`, which now masks before retrying with `OPT_NON_STR_KEYS`. All of this sits inside except `TypeError`, so the fast path is untouched (locally measured: 2.1 µs/call vs 2.0 µs baseline).
- `create_dataset` - was the one API write in `client.py` still using raw `_orjson.dumps`.

Known limits (documented on the class)
 - F-string and `.format()` interpolation on the masked string undo the masking. Same is true for `'<str>'.join(...)`. The class of the argument has no say in the end result. If/when possible, the results themselves should be wrapped in a `LangSmithSecret` 
 - Two secret _keys_ in one dict collapse to the same (redacted) key. Having a secret as a _key_ in a dict should be rare, having two - even more so. 
 - A `LangSmithSecret` passed to a Pydantic model isn't masked due to Pydantic immediately converting it to a string and stripping the subclass.

### DIFF
--- a/python/langsmith/_internal/otel/_attribute_utils.py
+++ b/python/langsmith/_internal/otel/_attribute_utils.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Any, Optional
 
 from langsmith._internal import _orjson
+from langsmith.secret import _redact_secrets
 
 if TYPE_CHECKING:
     from opentelemetry.trace import Span  # type: ignore[import]
@@ -14,6 +15,11 @@ LANGSMITH_METADATA_PREFIX = "langsmith.metadata"
 
 def otel_safe_attribute_value(value: Any) -> Optional[Any]:
     """Convert a LangSmith metadata value for safe use in application OTel spans."""
+    # OTel has no `default` hook, so mask secrets before they become attributes.
+    try:
+        value = _redact_secrets(value)
+    except RecursionError:
+        return str(value)
     if value is None:
         return None
     if isinstance(value, (bool, bytes, int, float, str)):
--- a/python/langsmith/secret.py
+++ b/python/langsmith/secret.py
@@ -1,0 +1,189 @@
+"""Mark a value as a secret, so LangSmith masks it instead of tracing it.
+
+::
+
+    from langsmith import LangSmithSecret
+
+    API_KEY = LangSmithSecret(os.environ["VENDOR_API_KEY"])
+
+See :class:`LangSmithSecret` for the guarantees and the known limits.
+"""
+
+from __future__ import annotations
+
+import collections
+from typing import Any
+
+__all__ = ["LANGSMITH_SECRET_MASK", "LangSmithSecret"]
+
+LANGSMITH_SECRET_MASK = "[LANGSMITH SECRET]"
+"""Written in place of a secret. Distinct from the regex anonymizer's
+``[SECRET_DETECTED]`` so the two mechanisms are distinguishable in a trace."""
+
+_NOT_HANDLED: Any = object()
+"""Returned by :func:`_coerce_builtin_subclass` when it does not apply."""
+
+# `str` methods returning a new `str`; re-wrapped so a derived value stays
+# secret. `partition`/`split` return sequences of `str` and are handled below.
+_STR_RETURNING = (
+    "__add__",
+    "__format__",
+    "__getitem__",
+    "__mod__",
+    "__mul__",
+    "__rmul__",
+    "capitalize",
+    "casefold",
+    "center",
+    "expandtabs",
+    "format",
+    "format_map",
+    "join",
+    "ljust",
+    "lower",
+    "lstrip",
+    "removeprefix",
+    "removesuffix",
+    "replace",
+    "rjust",
+    "rstrip",
+    "strip",
+    "swapcase",
+    "title",
+    "translate",
+    "upper",
+    "zfill",
+)
+_SEQUENCE_RETURNING = ("partition", "rpartition", "rsplit", "split", "splitlines")
+
+
+class LangSmithSecret(str):
+    """A string that LangSmith serializes as ``[LANGSMITH SECRET]``.
+
+    Wrap a credential once, where it is read, and LangSmith masks it wherever
+    it appears in a trace, at any nesting depth::
+
+        @traceable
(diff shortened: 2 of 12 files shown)


python/langsmith/secret.py
```diff
@@ -0,0 +1,189 @@
...
+    "title",
+    "translate",
+    "upper",
+    "zfill",
+)
+_SEQUENCE_RETURNING = ("partition", "rpartition", "rsplit", "split", "splitlines")
+
+
+class LangSmithSecret(str):
+    """A string that LangSmith serializes as ``[LANGSMITH SECRET]``.
```
QuentinBrosse, line 61: Should we mark this class as experimental and/or move it to `py/src/experimental` until it matures?

python/langsmith/secret.py
```diff
@@ -0,0 +1,189 @@
...
+    - A secret used as a **dict key** is masked, but two of them in one dict
+      collapse to a single entry. A key held inside a dataclass is not masked:
+      orjson serializes those itself.
+    - An object whose own ``__str__`` interpolates a secret leaks it wherever
+      LangSmith stringifies that object. ``__repr__`` is masked; ``__str__``
+      belongs to the object.
+    - Nothing inside a **pydantic model** is masked. Pydantic serializes its
+      own fields and LangSmith deliberately does not override that: a hook
+      there would also replace the credential with the mask when a vendor
+      client dumps the model to build a request.
```
QuentinBrosse, line 95: Is there a way to issue a warning when a secret is stored in a Pydantic/dataclass? I'm worried that customers will miss this limitation.
Or maybe the agent will catch it. 🤔

python/langsmith/_internal/otel/_attribute_utils.py
```diff
@@ -14,6 +15,11 @@
 
 def otel_safe_attribute_value(value: Any) -> Optional[Any]:
     """Convert a LangSmith metadata value for safe use in application OTel spans."""
+    # OTel has no `default` hook, so mask secrets before they become attributes.
+    try:
+        value = _redact_secrets(value)
+    except RecursionError:
+        return str(value)
```
QuentinBrosse, line 22: Should we fail close here? Do we have a test for that?

QuentinBrosse (approved):

python/langsmith/secret.py
QuentinBrosse, line 1: Should we add some benches to measure performance?

python/langsmith/secret.py
emil-lc: Not for the general case in pydantic, where we have a model that defines the secret as a string - as in:
```
class MySettings(BaseModel):
   api_key: str
```
In the above case, Pydantic converts to "str" and strips the wrapper class when you feed the parameter. No chance there. If the attribute is a `dict` or `Any`, then we can mask the value at the cost of extra complexity and perf. It's still half-assed, though, and especially in version 1, it's much more clear to declare Pydantinc as being unsupported outright, than starting to talk about different attribute types (which end-users likely have no control over/knowledge of anyways).

Dataclass _values_ are supported; only dataclass _keys_ are not. I updated the description, as it was confusing on this point. It should be an extremely rare case, though, where you pass secrets as dataclass keys. We _could_ support that, as we already do dict keys, but it feels like an extremely rare edge case to be worth the extra lines.

python/langsmith/_internal/otel/_attribute_utils.py
emil-lc: If we fail closed/return a masked value, we'd be masking something that's non-secret, pretty much all of the time. A secret is a string, and it doesn't have nested children that could lead to recursion. 

What might be better here is to add an explicit guard clause for recursion detection. I'll see where htis leads.

python/langsmith/secret.py
emil-lc: Hmm, we don't have an experimental module for Python. Not sure.

python/langsmith/secret.py
emil-lc: That's a good idea!

emil-lc: Added benchmarks - comparing plaintext/default-anonymizers/LangSmithSecret cases

python/langsmith/_internal/otel/_attribute_utils.py
emil-lc: OK, I added cycle detection in https://github.com/langchain-ai/langsmith-sdk/pull/3502, but it gives me a consistent performance decrease of 13-15% in the `_redact_secrets` function, even for small objects. It's a trade-off that doesn't buy us anything significant - so it's not worth it.

Raising an exception in the exceptional case and returning the plain value is already good.

python/langsmith/secret.py
emil-lc: Discussed with @QuentinBrosse - decided against it due to unnecessary complexity.

QuentinBrosse (approved):

PR #3485, opened 2026-09-02
fix(py): make >20MB multipart bodies re-sendable on retry

thibautlehmann:
Fixes [LSDK-460](https://linear.app/langchain/issue/LSDK-460/make-20mb-multipart-bodies-re-sendable-on-retry).

## Bug
`_send_multipart_req` streams a `MultipartEncoder` for bodies >20MB instead of buffering to `bytes`. That object has no `tell`/`seek`, so urllib3 can't rewind it before a retry (dropped connection, or 502/503/504). The retry resends a stale `Content-Length` against an already-exhausted stream — the server hangs waiting for a body that never arrives.

Same bug existed in `_prepare_multipart_data` (backs `upload_examples_multipart`/`update_examples_multipart`), with no app-level retry loop to fall back on.

## Fix
`SeekableMultipartEncoder` wraps the streaming encoder with `tell`/`seek(0)`: rewinds every underlying file-like part and rebuilds the encoder. Below 20MB, nothing changes — bytes never needed a seek. Used at both call sites.

## Tests
Extended `test_multipart_retry_matrix.py` with >20MB cases for both call sites, driven through a real local HTTP server (not mocks) so urllib3's actual retry layer is exercised. Verified by reverting the fix and confirming the test hangs on the exact bug (`recv_into` blocked waiting for a response that never comes); restored, all pass.

### DIFF
--- a/python/langsmith/_internal/_multipart.py
+++ b/python/langsmith/_internal/_multipart.py
@@ -4,6 +4,10 @@ from collections.abc import Iterable
 from io import BufferedReader
 from typing import Union
 
+from requests_toolbelt import (  # type: ignore[import-untyped]
+    multipart as rqtb_multipart,
+)
+
 MultipartPart = tuple[
     str, tuple[None, Union[bytes, BufferedReader], str, dict[str, str]]
 ]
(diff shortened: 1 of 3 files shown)


thibautlehmann:
Superseded by #3483 (LSDK-501), which fixes a strict superset of this bug — same tell/seek rewind mechanism, applied unconditionally rather than only above the 20MB streaming threshold. Closing in favor of that PR; the retry-matrix test coverage from this branch will be ported over separately.

PR #3492, opened 2026-09-04
feat(py,js): report the configured tracing sample rate on runs

thibautlehmann:
Report the client's configured tracing sample rate on every run, so a customer's total trace potential is recoverable from the sampled subset.

## Known gaps
- if client use raw OTEL we can’t track the sampling


js/src/client.ts
```diff
@@ -3138,7 +3144,7 @@ export class Client implements LangSmithTracingClientInterface {
       headers["x-tenant-id"] = options.workspaceId;
     }
     const body = serializePayloadForTracing(
-      run,
+      await this._mergeRuntimeEnvAndMaskMetadata(run),
```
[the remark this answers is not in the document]
thibautlehmann: @open-swe can you recheck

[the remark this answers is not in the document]
thibautlehmann: @open-swe can you recheck, i removed this part

python/langsmith/client.py
```diff
@@ -2482,6 +2483,8 @@ def _insert_runtime_env(self, runs: Sequence[dict]) -> None:
             metadata: dict = run_extra.setdefault("metadata", {})
             langchain_metadata = ls_env.get_langchain_env_var_metadata()
             added = {k: v for k, v in langchain_metadata.items() if k not in metadata}
+            if sample_rate is not None and "ls_tracing_sample_rate" not in metadata:
```
QuentinBrosse: Small questions:
- Why do we add metadata at runtime? Is it to conform to OTel, where we don't send runtime data?
- Why do we allow users to override the value?

python/langsmith/client.py
```diff
@@ -3881,10 +3884,9 @@ def update_run(
             data["outputs"] = self._hide_run_outputs(outputs)
         if events is not None:
             data["events"] = self._filter_new_token_events(events)
-        if data["extra"]:
-            if metadata := data["extra"].get("metadata"):
-                data["extra"]["metadata"] = self._hide_run_metadata(metadata)
-            self._insert_runtime_env([data])
+        if data["extra"] and (metadata := data["extra"].get("metadata")):
```
QuentinBrosse: Could you elaborate on this change?

If extra doesn't change between POST and PATCH, maybe we don't need to resend it in the PATCH?
I think this change on line 3856 will drop all metadata sent in a POST if PATCH metadata is empty.

QuentinBrosse (changes requested):

python/langsmith/client.py
thibautlehmann: Sorry, I got it wrong, I actually thought not sending an extra on PATCH would erase the POST's one. I reverted this change and the extra or {} line 3856

python/langsmith/client.py
thibautlehmann: 1. yes exactly, it's to conform to OTel
2. good point, let me fix it

js/src/client.ts
```diff
@@ -3138,7 +3144,14 @@ export class Client implements LangSmithTracingClientInterface {
...
     const body = serializePayloadForTracing(
-      run,
+      run.extra
+        ? mergeRuntimeEnvIntoRun(
+            run,
+            this.cachedLSEnvVarsForMetadata,
+            this.omitTracedRuntimeInfo,
+            this.tracingSampleRate,
+          )
+        : run,
```
emil-lc, line 3154: Similar to the Python side. Do we need to send this information during a PATCH/UpdateRun ? Looks like we were not serializing anything beseisdes the `run` before this PR.

thibautlehmann: Yes, the reason we need it is that POSTt's (create) extra is **replaced** on a PATCH, not merged (on the LangSmith server); unless the PATCH carries no extra, in which case the POST's one is kept.

So if the POST includes the runtime env and the PATCH sends a subset of it, that subset is what ends up stored.

Python already does the same on main btw (`if data["extra"]: ... self._insert_runtime_env([data]`) in [update_run, client.py:3887](https://github.com/langchain-ai/langsmith-sdk/blob/aecd6d1fbfaf4e01a60ccc6220c15d380fb126be/python/langsmith/client.py#L3887-L3888)), so this mostly aligns JS with it.

emil-lc (approved):

QuentinBrosse (approved):

PR #3506, opened 2026-09-09
fix(py): carry the ADK tool call ids into traced messages [LSDK-509]

zduric-langchain:
Every tool call in a traced ADK request now gets a stable id, and the
result that answers it carries that id as its `tool_call_id`, so the trace
view pairs the two instead of rendering the result's role as Unknown.
Before, tool results had no `tool_call_id` key at all, and assistant-side
ids were numbered inside the per-content loop, so two single-call turns
both emitted `call_0`.

A shared `ToolIdState` does the numbering, under an `ls-adk-` prefix that
cannot collide with a real `call_*` id. It spans the whole request rather
than one turn, and `_client.py` and the request conversion share an
instance, so an id survives from one run's outputs into the next run's
inputs. Calls are queued per tool name, so parallel and out-of-order
results still pair.

Where ADK does supply its own `adk-` ids, `_serialize_part` keeps them,
and a result consuming one drops the matching call from the queue. That is
the narrower path: ADK preserves its ids only for models that pair tool
calls by id (`AnthropicLlm`, `LiteLlm`, `OpenAIResponsesLlm`, `Gemini`
with `use_interactions_api`); everywhere else `preserve_function_call_ids`
is False and every id is stripped from the request (adk-python
contents.py:47-80, :677-683), and the response path yields the raw
`LlmResponse` before the ids are populated.

### DIFF
--- a/python/langsmith/integrations/google_adk/_messages.py
+++ b/python/langsmith/integrations/google_adk/_messages.py
@@ -3,7 +3,9 @@
 from __future__ import annotations
 
 import base64
+import itertools
 import json
+from collections import defaultdict, deque
 from typing import Any
 
 
@@ -47,6 +49,7 @@ def _serialize_part(part: Any) -> dict[str, Any]:
         fc = part.function_call
         return {
             "type": "tool_use",
+            "id": getattr(fc, "id", None),
             "name": getattr(fc, "name", "unknown"),
             "input": dict(getattr(fc, "args", None) or {}),
         }
@@ -55,6 +58,7 @@ def _serialize_part(part: Any) -> dict[str, Any]:
         fr = part.function_response
         return {
             "type": "tool_result",
+            "id": getattr(fr, "id", None),
             "name": getattr(fr, "name", "unknown"),
             "content": _safe_serialize(getattr(fr, "response", None)),
         }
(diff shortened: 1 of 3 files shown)


python/langsmith/integrations/google_adk/_messages.py
```diff
@@ -107,6 +111,33 @@ def _safe_serialize(obj: Any) -> Any:
...
+def _tool_call_id(
+    tool_call: dict[str, Any],
+    fallback_ids: itertools.count,
+    unanswered: dict[str, deque[str]],
+) -> str:
+    """Return the id of an assistant tool call, preferring the one ADK set."""
+    call_id = tool_call.get("id")
+    if call_id:
+        return str(call_id)
+    call_id = f"call_{next(fallback_ids)}"
```
jacoblee93: Do we actually need fallback ids? When would an id not be present?

jacoblee93 (approved):

thibautlehmann (approved):

python/langsmith/integrations/google_adk/_messages.py
thibautlehmann: The id of FunctionCall type from ADK LlmRequest is set as Optional:
Screenshot 2026-09-11 at 11 23 12

And in some cases ADK strips the id completely, see: https://github.com/google/adk-python/blob/a5791dab0b1aeae88caab31ca6951653dcdbba8b/src/google/adk/flows/llm_flows/contents.py#L677

python/tests/unit_tests/wrappers/test_google_adk.py
```diff
@@ -425,3 +425,134 @@ async def _run():
         f"Expected active context inside tool to be the tool span "
         f"({tool_run['id']}), got {captured_run_id}"
     )
+
+
+def _tool_call_request(*calls: tuple[str | None, str | None], name: str = "weather"):
```
emil-lc: Can you add a test using the `wrap_flow_call_llm_async` wrapper too?

PR #3508, opened 2026-09-09
evaluate: allow upload_results to be resolved per run

sineha-mani:
## Summary
- Lets someone running an evaluation decide, run by run, which ones get saved instead of saving everything — for example, only save 1 out of every 10 runs to cut down on data volume. The overall score still counts every run that happened, even the ones that weren't saved.

## Test plan
- [x] Unit tests


samecrowder:
is the name `upload_results` confusing? my understanding is when `False` you would get scores still in the experiment view, but not the traces. is `upload_runs` or something like that more fitting?

PR #3518, opened 2026-09-10
fix(js): preserve 403 JSON error body in raiseForStatus

### LINKED ISSUE #3517: raiseForStatus discards the 403 JSON error body for all non-org-scoped error codes (empty Message)
## Summary

In `raiseForStatus` (`js/src/utils/error.ts`), the `403` branch reads the response body with `await response.json()` to inspect `errorData.error`, but only assigns `errorBody` when the code is exactly `org_scoped_key_requires_workspace`. For **every other 403 error code**, the parsed `errorData` is discarded and control falls through to the `if (errorBody === undefined)` block, which calls `await response.text()` on the **already-consumed** body. That second read of a spent stream yields `""`, so the thrown error's `Message:` is empty and the real rejection reason the server sent is lost.

## Where

```ts
if (response.status === 403) {
  try {
    const errorData = await response.json();          // consumes the body
    const errorCode = errorData?.error;
    if (errorCode === "org_scoped_key_requires_workspace") {
      errorBody = "This API key is org-scoped and requires workspace specification. ...";
    }
    // no else: errorData is discarded for any other error code
  } catch (_e: any) {
    // non-JSON body path -> throws bare "<status> <statusText>"
    ...
  }
}
if (errorBody === undefined) {
  try {
    errorBody = await response.text();                // body already consumed -> ""
  } catch (_e: any) {
    errorBody = "";
  }
}
```

Introduced in #1920 (org-scoped keys), where the special-case `if` was added without an `else`, orphaning all other 403 error codes.

## Impact

Any 403 whose JSON body carries a descriptive `error` code other than `org_scoped_key_requires_workspace` is logged/thrown with an empty message. We hit this on the multipart trace-upload path, where a constant fraction of `POST /runs/multipart` batches was being rejected with 403 and dropped, and every log line looked like:

```
Failed to send multipart request. Received status [403]: Forbidden. Message:
```

The empty `Message:` made the failures undiagnosable — the server was naming the reason on every rejection, and the client was throwing it away. (This still reproduces on `main` and in the current `latest`, `0.10.2`.)

## Suggested fix

Preserve the parsed `errorData` instead of re-reading the consumed stream:

```ts
if (errorCode === "org_scoped_key_requires_workspace") {
  errorBody =
    "This API key is org-scoped and requires workspace specification. ...";
} else {
  errorBody = JSON.stringify(errorData);
}
```

PR to follow.

## Environment

- `langsmith` `0.7.15` (reproduces), also present on `main` and `0.10.2`
- Node.js (undici `fetch`) — the body is a single-use stream, so the second read returns `""`

xdumaine:
## Description

Fixes #3517.

In `raiseForStatus` (`js/src/utils/error.ts`), the `403` branch reads the body with `await response.json()` to inspect `errorData.error`, but only assigns `errorBody` when the code is `org_scoped_key_requires_workspace`. For every other 403 error code, the parsed `errorData` was discarded and control fell through to `if (errorBody === undefined)`, which calls `await response.text()` on the **already-consumed** body. That second read of a spent stream returns `""`, so the thrown error's `Message:` is empty and the reason the server sent is lost.

This surfaces prominently on the multipart trace-upload path — a 403 rejection logs `Failed to send multipart request. Received status [403]: Forbidden. Message:` with nothing after `Message:`, making the failure undiagnosable even though the server named the cause in the JSON body.

Introduced in #1920, where the special-case `if` was added without an `else`.

## Change

Assign `errorBody = JSON.stringify(errorData)` for any other 403 code, preserving the server's reason instead of re-reading the consumed stream:

```ts
if (errorCode === "org_scoped_key_requires_workspace") {
  errorBody =
    "This API key is org-scoped and requires workspace specification. ...";
} else {
  errorBody = JSON.stringify(errorData);
}
```

The non-JSON body path (`catch`) is unchanged and still throws the bare `<status> <statusText>`.

## Notes

- One-line behavioral change scoped to the 403 branch; the `ok`, 404, and 409 paths are untouched.
- Happy to add a unit test covering a 403 with a JSON body whose `error` code is not `org_scoped_key_requires_workspace` if you'd like — point me at the preferred spec location.

### DIFF
--- a/js/src/tests/error.test.ts
+++ b/js/src/tests/error.test.ts
@@ -1,0 +1,63 @@
+import { raiseForStatus } from "../utils/error.js";
+
+// A real `Response` body is a single-use stream: once `raiseForStatus` reads it
+// with `response.json()`, the fallback `response.text()` can no longer read it.
+// These tests use real `Response` objects so the consumed-stream behaviour the
+// 403 branch has to cope with is reproduced faithfully.
+describe("raiseForStatus 403 handling", () => {
+  it("includes the JSON error body for a non-org-scoped 403", async () => {
+    const body = {
+      error: "trace_project_forbidden",
+      detail: "API key cannot write to the target project",
+    };
+    const response = new Response(JSON.stringify(body), {
+      status: 403,
+      statusText: "Forbidden",
+      headers: { "content-type": "application/json" },
+    });
+
+    await expect(
+      raiseForStatus(response, "send multipart request"),
+    ).rejects.toThrow(/trace_project_forbidden/);
+  });
+
+  it("does not leave an empty Message when the body was valid JSON", async () => {
+    // Regression guard for the bug where response.json() consumed the body and
+    // the response.text() fallback then read a spent stream and yielded "".
+    const body = { error: "some_other_error" };
+    const response = new Response(JSON.stringify(body), {
+      status: 403,
+      statusText: "Forbidden",
+      headers: { "content-type": "application/json" },
+    });
+
+    await expect(
+      raiseForStatus(response, "send multipart request"),
+    ).rejects.toThrow(/Message: (?!\s*$).+/);
+  });
+
+  it("keeps the descriptive message for org_scoped_key_requires_workspace", async () => {
+    const body = { error: "org_scoped_key_requires_workspace" };
+    const response = new Response(JSON.stringify(body), {
+      status: 403,
+      statusText: "Forbidden",
+      headers: { "content-type": "application/json" },
+    });
+
+    await expect(
+      raiseForStatus(response, "send multipart request"),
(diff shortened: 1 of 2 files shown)


xdumaine:
Added a unit test (`js/src/tests/error.test.ts`) covering the 403 branch. It uses real `Response` objects so the single-use-body behaviour is exercised faithfully:

- non-org-scoped 403 → thrown message now contains the server's JSON reason (fails on `main`, passes here)
- valid-JSON 403 → `Message:` is no longer empty (direct regression guard)
- `org_scoped_key_requires_workspace` → descriptive message unchanged
- non-JSON 403 body → still throws the bare `403 Forbidden` via the catch path

Verified the two fix-specific cases fail without the `else` branch and pass with it; `oxlint`/`oxfmt` clean.

PR #3520, opened 2026-09-10
feat(py): address runs by agent instead of by project

KiewanVillatel:
## What

Adds `LANGSMITH_AGENT_ID` / `LANGSMITH_AGENT_ENVIRONMENT` and lets a run be addressed by **agent** instead of by project, per [POST /api/v1/runs/multipart](https://app.notion.com/p/POST-api-v1-runs-multipart-3d7808527b178194bcd6d441bac8e2ba). The two map onto two flat payload fields:

```json
{"id": "...", "agent_id": "checkout", "agent_environment": "staging"}
```

Both are required and neither is defaulted, so a run can't reach `production` by forgetting to name an environment. Python only; the TypeScript mirror and feedback parts (#3533) are separate.

## How

**One addressing mode per part.** `session_name` was defaulted in (to `LANGSMITH_PROJECT`, else `"default"`) on *every* run, so simply adding the agent fields would have made every payload carry both modes — which the endpoint rejects. That default is now suppressed when a run is agent-addressed, and the agent fields go on `patch` parts as well as `post`, so both halves of a run address the same target.

**Precedence: explicit beats env; within a tier, project beats agent.**

| configuration | resolves to |
|---|---|
| agent env only | agent |
| agent env + `LANGSMITH_PROJECT` | both, so the endpoint refuses the pair — the client warns at construction |
| agent env + explicit project | project |
| explicit agent + `LANGSMITH_PROJECT` | agent |
| both named explicitly in one call | **raises** |

That last row is the SDK's **only** rejection. Everything else is left to the endpoint — but resolution drops the agent before the payload is built, so the endpoint never sees this conflict and nothing would otherwise tell the caller their agent was discarded. An agent from the *environment* beside an explicit project is not a conflict: the project wins. That's what lets an evaluation set `project_name=experiment_name` while `LANGSMITH_AGENT_ID` is set process-wide.

**Also:** `tracing_context(agent_id=…, agent_environment=…)` with two context vars that `@traceable` reads and nested runs inherit — without this the main API could never be agent-addressed, since `selected_project` always fell back to `"default"`. `WriteReplica` gains the pair with the same precedence per replica. Two new `baggage` keys so agent addressing survives a service hop.

## Experimental

Marked experimental across the surface. Agent addressing is beta and enabled per workspace; a workspace without it **rejects these runs with a 403**, and a `LangSmithBetaWarning` (via `warnings.warn`, so it is suppressed by `PYTHONWARNINGS=ignore` and by most pytest filters) warns the user once per process.

### DIFF
--- a/python/langsmith/async_client.py
+++ b/python/langsmith/async_client.py
@@ -622,6 +622,11 @@ class AsyncClient:
                 run_create["agent_environment"] = agent_environment
             else:
                 run_create.pop("agent_environment", None)
+        if run_create.get("agent_key") is None:
+            if agent_key := ls_utils.get_tracer_agent_key():
+                run_create["agent_key"] = agent_key
+            else:
+                run_create.pop("agent_key", None)
         await self._arequest_with_retries(
             "POST", "/runs", content=ls_client._dumps_json(run_create)
         )
--- a/python/langsmith/async_client.py
+++ b/python/langsmith/async_client.py
@@ -618,15 +618,15 @@ class AsyncClient:
             **kwargs,
         }
         if run_create.get("agent_environment") is None:
-            if agent_environment := ls_utils.get_tracer_environment():
+            if agent_environment := ls_utils.get_tracer_agent_environment():
                 run_create["agent_environment"] = agent_environment
             else:
                 run_create.pop("agent_environment", None)
-        if run_create.get("agent_key") is None:
-            if agent_key := ls_utils.get_tracer_agent_key():
-                run_create["agent_key"] = agent_key
+        if run_create.get("agent_id") is None:
+            if agent_id := ls_utils.get_tracer_agent_id():
+                run_create["agent_id"] = agent_id
             else:
-                run_create.pop("agent_key", None)
+                run_create.pop("agent_id", None)
         await self._arequest_with_retries(
             "POST", "/runs", content=ls_client._dumps_json(run_create)
         )
--- a/python/langsmith/async_client.py
+++ b/python/langsmith/async_client.py
@@ -617,6 +617,11 @@ class AsyncClient:
             "revision_id": revision_id,
             **kwargs,
         }
+        if run_create.get("agent_environment") is None:
+            if agent_environment := ls_utils.get_tracer_environment():
+                run_create["agent_environment"] = agent_environment
+            else:
+                run_create.pop("agent_environment", None)
         await self._arequest_with_retries(
             "POST", "/runs", content=ls_client._dumps_json(run_create)
         )
--- a/python/langsmith/async_client.py
+++ b/python/langsmith/async_client.py
@@ -212,6 +212,7 @@ class AsyncClient:
                 - `False`: Disable caching (equivalent to `disable_prompt_cache=True`)
                 - `AsyncCache(...)`/`AsyncPromptCache(...)`: Use a custom cache instance
         """
+        ls_utils.validate_agent_addressing_env()
         self._retry_config = retry_config or {"max_retries": 3}
         self._custom_headers = headers or {}
         env_api_url = ls_client._get_langsmith_env_var_uncached("ENDPOINT")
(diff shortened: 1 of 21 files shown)


python/langsmith/utils.py
```diff
@@ -462,6 +462,61 @@ def get_tracer_project(return_default_value=True) -> Optional[str]:
...
+            f"(agent {agent_id!r}, project {project!r}), but a run is addressed "
+            "either by agent or by project, not both. Unset one of them, or "
+            "pass an explicit project per run to override the agent."
+        )
+    if get_tracer_agent_environment() and not agent_id:
+        raise LangSmithUserError(
+            "LANGSMITH_AGENT_ENVIRONMENT is set without LANGSMITH_AGENT_ID. An "
+            "environment only narrows an agent, so on its own it doesn't "
+            "identify where runs should go. Set LANGSMITH_AGENT_ID too, or "
+            "unset LANGSMITH_AGENT_ENVIRONMENT."
```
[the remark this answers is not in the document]
KiewanVillatel: Removed validation from the SDK. Will rely on backend validation and 400 instead.

python/langsmith/utils.py
```diff
@@ -462,6 +464,165 @@ def get_tracer_project(return_default_value=True) -> Optional[str]:
...
+    agent addressing rejects the runs, so tracing is lost rather than falling
+    back to a project.
+
+    Must be one of ``local``, ``development``, ``staging`` or ``production``
+    -- matched case-insensitively, surrounding space ignored. The endpoint
+    rejects anything else rather than defaulting it, so a near miss like
+    ``prod`` fails the whole batch.
+
+    Read from ``LANGSMITH_AGENT_ENVIRONMENT`` only. Unlike most LangSmith
+    variables there is no legacy ``LANGCHAIN_`` alias: the legacy namespace is
```
QuentinBrosse: I am not sure it's legacy. Probably good enough for the experimental release.

python/langsmith/utils.py
```diff
@@ -462,6 +464,165 @@ def get_tracer_project(return_default_value=True) -> Optional[str]:
...
+    """
+    if project is not None:
+        return project, None, None
+    if is_agent_addressed(agent_id, agent_environment):
+        return None, *resolve_agent_addressing(agent_id, agent_environment)
+    env_agent_id = get_tracer_agent_id()
+    env_agent_environment = get_tracer_agent_environment()
+    if is_agent_addressed(env_agent_id, env_agent_environment):
+        return (
+            get_tracer_project(return_default_value=False),
```
QuentinBrosse: why not None?

python/langsmith/utils.py
```diff
@@ -462,6 +464,165 @@ def get_tracer_project(return_default_value=True) -> Optional[str]:
...
+            env_agent_id,
+            env_agent_environment,
+        )
+    return get_tracer_project(), None, None
+
+
+def warn_on_agent_and_project_env() -> None:
+    """Warn when the environment configures both an agent and a project.
```
QuentinBrosse: nice!

python/langsmith/client.py
```diff
@@ -770,6 +770,43 @@ def _format_feedback_score(score: Union[float, int, bool, None]):
     return score
 
 
+def _reject_conflicting_addressing(
```
QuentinBrosse: Should we move it in a dedicated `agent_addressing.py`? This file is already huge.
Same for `_apply_agent_addressing` method.

Maybe we should also move the utils.py funcs here?

python/langsmith/client.py
```diff
@@ -3876,7 +4011,11 @@ def update_run(
             "extra": extra,
             "session_id": kwargs.pop("session_id", None),
             "session_name": kwargs.pop("session_name", None),
+            "agent_id": kwargs.pop("agent_id", None),
```
QuentinBrosse, line 3927: Should we add an experiemental warn?

python/langsmith/async_client.py
```diff
@@ -632,6 +666,7 @@ async def update_run(
     ) -> None:
         """Update a run."""
```
QuentinBrosse: Should we add an experimental warn?

python/langsmith/client.py
KiewanVillatel: Done, except for the `get_tracer_agent_id` / `get_tracer_agent_environment` to keep it close to `get_tracer_project`.

python/langsmith/client.py
KiewanVillatel: Already covered: warn_is_beta() fires inside _agent_addressing.apply_to_payload, which both update_run and AsyncClient call. Verified an update payload carrying the pair warns, and a project-only one doesn't.

python/langsmith/run_helpers.py
```diff
@@ -1544,20 +1669,33 @@ def _get_parent_run(
...
     if isinstance(parent, Mapping):
         return run_trees.RunTree.from_headers(
             parent,
             client=langsmith_extra.get("client"),
             # Precedence: headers -> cvar -> explicit -> env var
-            project_name=_get_project_name(langsmith_extra.get("project_name")),
+            project_name=project_name,
+            agent_id=agent_id,
+            agent_environment=agent_environment,
             replicas=langsmith_extra.get("replicas"),
```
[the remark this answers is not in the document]
KiewanVillatel: Fixed

QuentinBrosse (approved):

PR #3533, opened 2026-09-14
feat(py): address feedback parts by agent

KiewanVillatel:
> **Stacked on #3520** — review that first. Base branch is `kiewan/langsmith-environment-env-var`.

Addresses a `feedback.<run_id>` part by **agent** instead of by project, per the feedback section of [POST /api/v1/runs/multipart](https://app.notion.com/p/POST-api-v1-runs-multipart-3d7808527b178194bcd6d441bac8e2ba). `create_feedback` (sync and async) and `FeedbackBase` take flat `agent_id` / `agent_environment`; conflicts and half pairs are left to the endpoint's 400, as with run parts.

Feedback **never reads the agent env vars** — it copies the addressing of the run it describes, so a stray agent id can't file feedback against a project the run never reached. An agent pair satisfies the `_check_feedback_session_id` gate, since it locates the project directly.

Python only; the TypeScript mirror follows with #3520's.


python/langsmith/async_client.py
```diff
@@ -1204,6 +1206,14 @@ async def create_feedback(
...
             extend_trace_retention: If false, create the feedback without
                 extending the trace's retention tier.
+            agent_id: The agent to attach this feedback to, instead of a
+                project. Pass whatever the run being described was traced to.
+                Cannot be combined with `session_id` / `project_id`, and is
+                never read from `LANGSMITH_AGENT_ID`: feedback follows its run,
+                not the ambient environment. The agent must already exist;
+                unlike run ingestion, a feedback part never creates one.
+            agent_environment: Narrows `agent_id`, and requires it. Defaults
+                server-side to `production` when omitted.
```
QuentinBrosse, line 1236: Should we also flag as experimental?

python/langsmith/client.py
```diff
@@ -8451,6 +8453,17 @@ def create_feedback(
...
+                The agent to attach this feedback to, instead of a project. Pass
+                whatever the run being described was traced to -- for a run
+                created in this process, `run_tree.agent_id`. Cannot be combined
+                with `session_id` / `project_id`, and is never read from
+                `LANGSMITH_AGENT_ID`: feedback follows its run, not the ambient
+                environment. The agent must already exist; unlike run ingestion,
+                a feedback part never creates one.
+            agent_environment (Optional[str]):
+                Narrows `agent_id`, and requires it. Defaults server-side to
+                `production` when omitted.
```
QuentinBrosse, line 8391: Should we also flag as experimental?

python/langsmith/schemas.py
```diff
@@ -649,6 +649,16 @@ class FeedbackBase(BaseModel):
     """The source of the feedback."""
     session_id: Optional[UUID] = None
     """The associated project ID (Session = Project) this feedback is logged for."""
+    agent_id: Optional[str] = None
+    """The agent this feedback is logged for, instead of a project.
```
QuentinBrosse, line 661: Should we also flag as experimental?

python/langsmith/async_client.py
KiewanVillatel: Done

python/langsmith/client.py
KiewanVillatel: Done

python/langsmith/schemas.py
KiewanVillatel: Done

QuentinBrosse (approved):

PR #3547, opened 2026-09-16
feat(evaluate): add disable_evaluator_tracing toggle to evaluate() and aevaluate()

sineha-mani:
Lets you turn off tracing for evaluators when running an evaluation, so only the thing being tested shows up in LangSmith.
Scores are still recorded on each run as before; nothing changes unless you opt in.


python/langsmith/evaluation/_arunner.py
```diff
@@ -92,6 +94,7 @@ async def aevaluate(
     blocking: bool = True,
     experiment: Optional[Union[schemas.TracerSession, str, uuid.UUID]] = None,
     upload_results: bool = True,
+    trace_evaluators: bool = True,
```
jacoblee93: nit: prefer falsey defaults

`disable_evaluator_tracing: bool = False`

jacoblee93 (approved):

python/langsmith/evaluation/_arunner.py
sineha-mani: updated!!

PR #3548, opened 2026-09-16
fix(js): preserve the 403 JSON error body in raiseForStatus

### LINKED ISSUE #3517: raiseForStatus discards the 403 JSON error body for all non-org-scoped error codes (empty Message)

QuentinBrosse:
Fixes #3517 (LSDK-512)

The 403 branch read the body with `response.json()`, then the code below read it again with `response.text()`. A `Response` body is single-use, so the second read returned `""` — and since only `org_scoped_key_requires_workspace` assigned `errorBody`, every other 403 hit that spent read and threw with an empty message:

```
Failed to send multipart request. Received status [403]: Forbidden. Message:
```

**Fix:** read the body once as text, `JSON.parse` it for the org-scoped case.

| Case | Before | After |
|---|---|---|
| 403, org-scoped key | workspace hint | unchanged |
| **403, any other JSON code** | **`Message:` empty** | **`Message: {"error":"..."}`** |
| 403, non-JSON body | bare `403 Forbidden` | unchanged |
| 404 / 409 / other | body text | unchanged |

## What to check

- **`JSON.parse` replaces `response.json()`** — same throw-on-invalid behavior, including empty body.
- **403 bodies now reach logs.** Server-generated, and every other status already does this.
- **New tests use real `Response` objects.** The existing 403 tests in `client.test.ts` mock `text()` to re-serve on every call, so a double read is invisible to them — that's why this is a separate file. Reverting the fix makes test 1 fail with the empty `Message:` above.


js/src/utils/error.ts
```diff
@@ -144,16 +144,17 @@ export async function raiseForStatus(
...
-      const errorCode = errorData?.error;
-      if (errorCode === "org_scoped_key_requires_workspace") {
-        errorBody =
-          "This API key is org-scoped and requires workspace specification. " +
-          "Please provide 'workspaceId' parameter, " +
-          "or set LANGSMITH_WORKSPACE_ID environment variable.";
-      }
+      errorData = JSON.parse(errorBody);
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
     } catch (_e: any) {
```
thibautlehmann: suggestion:
 if the errorBody is a plain string it would hit the catch here, throwing a bare 403 Foribidden with no context. 

I guess most of self-hosted/BYOC clients have LS behind a proxy/waf, often returning 403 as text
I suggest removing the throw in the catch, and keep it empty.
It's out of scope but I think it's still important

thibautlehmann (changes requested):

js/src/utils/error.ts
thibautlehmann: did it, you can approve & merge if relevant for you: https://github.com/langchain-ai/langsmith-sdk/pull/3551 @QuentinBrosse

thibautlehmann: merged

thibautlehmann (approved):

PR #3559, opened 2026-09-18
fix(py): preserve invocation parameters in OTel exports

TorgemanTarak:
When LangChain runs are exported through otel, tool definitions stored in `extra.invocation_params.tools` are omitted from the outgoing spans. This leave the LangSmith Tools tab absent even though the same run renders correctly with native tracing.

Preserve `extra.invocation_params` as a JSON string in the supported OpenInference `llm.invocation_parameters` attribute. The existing metadata and individual GenAI parameter mappings remain intact. This allows an OTLP forwarding service to retain the tool definitions without changing their provider-specific shape.

## Validation

- `make format`
- `uv run ruff check .`
- `python -m mypy --platform linux langsmith` passed for all 123 source files.
- `make tests TEST="tests/unit_tests/test_otel_exporter.py tests/unit_tests/test_otel_url.py -n 0 -q"` 31 passed, including OTLP protobuf round-trip regression tests for run creation and updates.


emil-lc:
@TorgemanTarak , thank you for the report and the associated PR! 

`llm.invocation_parameters` is not a `gen_ai` convention, though, so we'd steer away from it in favor of the neutral `gen_ai` attributes coming from the OpenTelemetry project. I created https://github.com/langchain-ai/langsmith-sdk/pull/3584 instead, which contains the `invocation_params.tools` fix that you are after.

PR #3576, opened 2026-09-22
fix(py): trace ClaudeSDKClient.receive_messages()

thibautlehmann:
`receive_messages()` is now traced like `receive_response()`. Before, client apps looping on the continuous stream  receive_messages()  got zero traces. No root run, no warning.

## Verified

- 2 new unit tests: root run via `receive_messages()` + break; single root run via `receive_response()`.
- Real SDK 0.2.148, both repro paths from the issue, Path B passing now

## Also in this PR

**CI fix for `test_runs.py::test_trace_file_path`** (`7912a63c`). Also red on `main`: `assert run.attachments` -> `{}`. `list_runs()` sends a fixed `select` without `s3_urls` and the backend now honours it, so attachments come back empty from listings. `list_runs()` is deprecated and restoring `s3_urls` would re-enable eager attachment downloads on every call, so the test now reads the run by id instead (`read_run` sends no `select`). Passes against beta locally.


python/langsmith/integrations/claude_agent_sdk/_client.py
```diff
@@ -698,12 +716,35 @@ async def _traced_receive_response(self: Any) -> AsyncGenerator[Any, None]:
...
+                    # Restore the caller's context while yielding so application
+                    # work doesn't become part of this response's trace.
+                    stream_context = get_tracing_context()
+                    _set_tracing_context(caller_context)
+                    _current_session.set(caller_session)
+                    set_parent_run_tree(caller_parent)
+                    try:
+                        yield msg
+                    finally:
+                        # Resume processing or cleanup in the response's context.
```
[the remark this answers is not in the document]
emil-lc: Yeah, that looks like a problem

python/langsmith/integrations/claude_agent_sdk/_client.py
```diff
@@ -698,12 +717,36 @@ async def _traced_receive_response(self: Any) -> AsyncGenerator[Any, None]:
...
```
emil-lc: This code also smells like it would look better in a context manager - `with` statement. Can you check if LLM can make it prettier that way?

thibautlehmann: done, can you recheck?

python/tests/integration_tests/test_runs.py
```diff
@@ -511,7 +511,9 @@ def my_func(foo: Attachment, bar: Attachment):
     )
     runs = list(langchain_client.list_runs(project_name=project_name, filter=_filter))
     assert len(runs) == 1
-    run = runs[0]
+    # list_runs() selects a fixed field set without s3_urls, so attachments are
+    # only populated when reading a single run.
+    run = langchain_client.read_run(runs[0].id)
```
QuentinBrosse, line 516: We might want to switch to new methods in a follow-up PR

thibautlehmann: agreed, added to my todos

